Issue #29, opened 2023-01-29
Header visibility reactive to scrolling


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

It would be great if [WebsiteHeader.vue](https://github.com/activist-org/activist/blob/main/frontend/components/WebsiteHeader.vue) was reactive to user scrolling. The ideality would be that there would be no shadow at first and that once the user scrolls the background of the header would get an alpha value. Beyond this, ideally the header would disappear when scrolling down and reappear upon scrolling up.

### Contribution

I'll work on this eventually, but would be happy to discuss or get some help with it 🙃

andrewtavis:
@purupurupu, if you write in here I can assign the issue to you 😊 Let me know if you'd like me to look into possible solutions to this! Happy to support 🚀

purupurupu:
@andrewtavis 
I would be happy to be assigned to Issue.

andrewtavis:
Assigned! 😊 Here if you have questions or want to check your ideas :)

purupurupu:
@andrewtavis 
Thank you for assign !!
I just fixed and pushed it.please check it.

andrewtavis:
@purupurupu, hope it's ok that I finished the feature off :) I'm trying to keep the styling in Tailwind, so made sense to me 😊

Here's what changed in 6205822:
- Added `duration-500` to the header class, which thus moves the transition there from the `<style>`
- Tested the opacities a bit and decided to go with 0.8 for when we scroll up
- Used Tailwind `:class` and set the background colors so that they're responsive to light and dark mode as well as what the opacity value is
- Referenced the height of the header and used that as the cutoff point for when the header should be made transparent
    - Thus there's no case where we see the background color behind the header :)

Thanks so much for your work on this! Really enjoyed working with you 😊

purupurupu:
@andrewtavis 
I see, your correction is spot on!
Thanks for being so accommodating! I learned a lot!

andrewtavis:
Happy to have you contribute to more issues in the future, @purupurupu! 😊 Let me know if you have ideas or want to work on something :)

PR #57, opened 2023-02-03
update Header visibility reactive to scrolling.

purupurupu:
## Description
I tried to fix Header visibility reactive to scrolling.
Please let me know if there is anything that needs to be corrected.

## Related issue
#29

### DIFF
--- a/frontend/components/WebsiteHeader.vue
+++ b/frontend/components/WebsiteHeader.vue
@@ -61,7 +61,7 @@ import { Ref } from "vue";
 export default {
   setup() {
     const headerOpacity: Ref<number> = ref(1);
-    const prevYOffset: Ref<number> = ref(0);
+    const prevScrollY: Ref<number> = ref(0);
 
     onMounted(() => {
       window.addEventListener("scroll", handleScroll);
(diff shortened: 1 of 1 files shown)


andrewtavis (changes requested):
Hey @purupurupu! Thanks for this, and please don't stress about the requested changes 😊 This is a really good start to closing this issue 👏

Big thing is that the behavior isn't exactly what we're looking for. I'll define it a bit better :)
- At the top of the page the header has full opacity (as it does in your case)
- The user scrolls down -> the header disappears completely (0.0 opacity)
- The user scrolls up -> the header appears, and if the page isn't at the top then we'd have an opacity of 0.5 or so for just the background of the header
    - The logo and buttons would have full opacity, just the background would be a bit transparent

We need to test what opacity we want the header background to be to get this working effectively later on, but I think 0.5 is good to start 😊

andrewtavis:
I'll go ahead and assign the issue to you as you're working on this 😊

purupurupu:
@andrewtavis 
Oops, my bad.
Now that I know what it is, I'll fix the code.

andrewtavis:
@purupurupu, looking forward! 😊

andrewtavis:
@purupurupu, this is looking really good :) Have been testing it out just now. Only one comment, for `(scrollY > prevScrollY.value)` we want the whole header to disappear including the buttons, rather than just the background (maybe could have been a bit more descriptive). 

- `(scrollY > prevScrollY.value)`: everything disappears including the buttons
- `else headerOpacity.value`: background is 0.5 and the buttons are visible as you have it 👏
- `scrollY == 0`: everything is fully visible as you have it 👏

I'll calibrate the opacity and the transition time a bit once I merge 😊

andrewtavis (approved):
Actually let's leave it here 😊 I'm very happy with what's here now, and I want to use some Tailwind specific logic for the last parts of it 🤓

I'll update you in the issue on what I change!

PR #74, opened 2023-03-08
fix: changes to docker files for usability

wkyoshida:
**Description:**
- Updated the docker files to fix path issues to dependencies
- Updated the docker files with some simplification suggestions

----

Also some comments added below for explanations

Cheers :+1:

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,15 +1,14 @@
-FROM python:3.11.1-alpine3.16
+FROM python:3.11-alpine3.16
 
 # Set work dir for relative paths
-RUN mkdir -p /app
 WORKDIR /app
 
 # Install dependencies
-COPY ./backend/requirements.txt /app
+COPY ./requirements.txt .
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
 # Copy code from context to image
 COPY . .
 
-CMD gunicorn backend.wsgi --bind 0.0.0.0
+CMD python manage.py runserver 0.0.0.0:8000
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,24 +4,14 @@ services:
 
   backend:
     build:
-      context: .
-      dockerfile: ./backend/Dockerfile
+      context: ./backend
     container_name: django_backend
-    volumes:
-      - ./backend/:/app
     ports:
       - "8000:8000"
-    command: python manage.py runserver 0.0.0.0:8000
 
   frontend:
     build:
-      context: .
-      dockerfile: ./frontend/Dockerfile
-    image: nuxt_dev
+      context: ./frontend
     container_name: nuxt_frontend
-    volumes:
-      - ./frontend/:/app
-      - ./frontend/:/node_modules
     ports:
       - "3000:3000"
-    command: sh -c "npm rebuild esbuild && yarn dev"
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,14 +1,15 @@
 FROM node:19-alpine
 
 # Set work dir for relative paths
-RUN mkdir -p /app
 WORKDIR /app
 
 # Install dependencies
-COPY ./frontend/package*.json /app
+COPY ./package*.json .
+COPY ./yarn.lock .
 RUN yarn install && yarn cache clean
 
 # Copy code from context to image
 COPY . .
 
-ENV PATH ./node_modules/.bin/:$PATH
+RUN npm rebuild esbuild
+CMD yarn dev


backend/Dockerfile
```diff
@@ -1,15 +1,14 @@
-FROM python:3.11.1-alpine3.16
+FROM python:3.11-alpine3.16
```
wkyoshida, line 1: I don't believe the `3.11.1-alpine3.16` tag is listed in https://hub.docker.com/_/python for the `python` image anymore (it'll still work though, I believe). However, changed the tag here to be `3.11-alpine3.16` as that would be more general and refer to the same image as the following tags: 

[`3.11.2-alpine3.16, 3.11-alpine3.16, 3-alpine3.16, alpine3.16`](https://github.com/docker-library/python/blob/e664abbb9af1ee6aae5778d70a5df79d86d9ec57/3.11/alpine3.16/Dockerfile)

Unless there's a specific reason to use the minor version of 3.11.1 for python, could use the more general tag as that'll likely be updated with newer minor versions by the image maintainers. Hell, if the minor versions don't matter much, could even use the simple `alpine` tag instead. Just a thought if it makes sense here.

backend/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
 
 # Set work dir for relative paths
-RUN mkdir -p /app
```
wkyoshida, line 4: I believe the `mkdir` here might not be necessary as `WORKDIR` below should create the directory if it does not yet exist.

backend/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
 WORKDIR /app
```
wkyoshida, line 4: The `WORKDIR` here changes the working directory to `/app` so the `COPY` command below is already operating from within `/app`. Changing the destination to `.` then does the trick of copying the file into `/app`.

docker-compose.yml
```diff
@@ -4,24 +4,14 @@ services:
 
   backend:
     build:
-      context: .
-      dockerfile: ./backend/Dockerfile
+      context: ./backend
```
wkyoshida, line 7: The `context` here can be where the `Dockerfile` for the service actually is. Simplifies a bit some of the path-ing. 

For instance, the `dockerfile` value now doesn't need specifying as the `Dockerfile` inside the directory `./backend` is automatically used (though if there are multiple `Dockerfile`s in the same directory, then specifying is needed). 

Also, in the `Dockerfile` itself, the `COPY` command doesn't need the `./backend` path anymore as the image is now being built from the context `./backend`

---

Same applies for the frontend

docker-compose.yml
```diff
@@ -4,24 +4,14 @@ services:
...
     container_name: django_backend
-    volumes:
-      - ./backend/:/app
     ports:
       - "8000:8000"
-    command: python manage.py runserver 0.0.0.0:8000
```
wkyoshida, line 14: I believe `command` here overrides the `CMD` that is in the `Dockerfile`. So not sure if the `CMD` is even running when the container is spun up:

`gunicorn backend.wsgi --bind 0.0.0.0`

as that is being overwritten by 

`python manage.py runserver 0.0.0.0:8000`

So, updated the `CMD` to just be the `python` command.

docker-compose.yml
```diff
@@ -4,24 +4,14 @@ services:
...
 
   frontend:
     build:
-      context: .
-      dockerfile: ./frontend/Dockerfile
-    image: nuxt_dev
```
wkyoshida, line 20: I believe `image` just gives a custom name to the built image. It's not necessary really, so I wasn't sure about keeping it, since the same isn't done for the backend image.

frontend/Dockerfile
```diff
@@ -1,14 +1,15 @@
 FROM node:19-alpine
 
 # Set work dir for relative paths
-RUN mkdir -p /app
 WORKDIR /app
 
 # Install dependencies
-COPY ./frontend/package*.json /app
+COPY ./package*.json .
+COPY ./yarn.lock .
```
wkyoshida, line 8: Added the lock file for yarn so that it's also referenced when the dependencies are getting resolved in the image.

docker-compose.yml
```diff
@@ -4,24 +4,14 @@ services:
...
+      context: ./frontend
     container_name: nuxt_frontend
-    volumes:
-      - ./frontend/:/app
-      - ./frontend/:/node_modules
```
wkyoshida, line 24: `volumes` is for attaching an outside volume in the system to the container. The `COPY` command though can be used for getting files into the image, as it's been used already. 

So here this is attaching two outside volumes and I believe is the reason why others were having issues with nuxt not found!

In the previous version of the `Dockerfile`, the `./frontend` directory was being fully `COPY`d in. However as the `volumes` here are attached, they replace the existing `node_modules` that were generated when the dependencies were resolved. For a new contributor who has an empty `node_modules` (or does not have nuxt), the `volumes` is replacing with the empty directory. My suspicion is that the `docker compose` might have still been working for @andrewtavis as nuxt was likely already installed locally and the replacing `node_modules` would have it. Hopefully this makes sense, but that would be my guess.

wkyoshida (commented):
Let me know if there is anything that is still a little unclear!
I can expand if needed.

backend/Dockerfile
andrewtavis: I was just trying to get a generally up to date Python image, so sounds good :)

backend/Dockerfile
andrewtavis: Didn't know that! Thanks for explaining :)

docker-compose.yml
andrewtavis: Thanks for the detailed explanation, @wkyoshida!

docker-compose.yml
andrewtavis: Let's keep it simple I guess :)

docker-compose.yml
andrewtavis: I was wondering if it was something having to do with me having something locally installed. Makes a lot of sense 😊

andrewtavis (approved):
Thanks for the changes and all the explanations, @wkyoshida!

PR #76, opened 2023-03-11
Added error page

5ud03r5:
Created error page in /frontend dir which handles errors


andrewtavis (changes requested):
@5ud03r5, it all generally looks good and contains all the components we'd like for this page, but I'm just confused where the `package-lock.json` files are being added and why there's an edit to the `frontend/yarn.lock`. As we're using Docker to for yarn there should be no need to have the `package-lock.json` files in there. Could you remove these? 

And I guess what's happening with the `yarn.lock` file is that some changes based on your local machine are being pushed 🤔 I guess those are also not necessary to push. 

Happy to remove these files in the commit myself if you agree and then we can merge this 😊🚀 Thanks so much for your contribution!

5ud03r5:
> @5ud03r5, it all generally looks good and contains all the components we'd like for this page, but I'm just confused where the `package-lock.json` files are being added and why there's an edit to the `frontend/yarn.lock`. As we're using Docker to for yarn there should be no need to have the `package-lock.json` files in there. Could you remove these?
> 
> And I guess what's happening with the `yarn.lock` file is that some changes based on your local machine are being pushed 🤔 I guess those are also not necessary to push.
> 
> Happy to remove these files in the commit myself if you agree and then we can merge this 😊🚀 Thanks so much for your contribution!

Just made requested changes. Sorry for any confusion and I'm looking forward for more contributions 😊

andrewtavis (approved):
Thanks for sending along the changes!

Issue #68, opened 2023-02-11
Add transition to sidebar collapse


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to add a transition to the collapse of the activist sidebar. I'm not sure what the exact timing and bezier should be, but generally this would be to make the shift that happens from `sidebar.collapsed` and `sidebar.collapsedSwitch` combinations not instantaneous and have an appealing effect. We can start with 200-300 ms and test from there.

Note that the sidebar can be tested at https://activist.org/home/ :)

### Contribution

Happy to work on this eventually or support someone with interest in implementing it!

momanyisamuel:
Hi again @andrewtavis, I would like to help with this issue.

andrewtavis:
Great, @momanyisamuel! Thanks for your willingness to help 😊 Will assign you now, and please let me know if you need any assistance!

momanyisamuel:
@andrewtavis I was just about to ask a question on this actually glad you brought this up! I agree it makes sense as the behavior of the expand button is tied to the side bar I will factor this into the transition as well.

andrewtavis:
Thank you, @momanyisamuel! Looking forward to the result, and again let me know if I can help somehow :)

momanyisamuel:
@andrewtavis so far there is no issue adding the transition however some elements are affected, i.e the `SearchBar` and `MenuSelector` they both have text that overlaps the sidebar as the transition occurs - see image below. Before i proceed i just wanted to know if these components are being reused elsewhere and you can also advise accordingly

![image](https://user-images.githubusercontent.com/4663177/226129580-fd93fcff-aaa0-4a2d-8a13-e587d33e8b28.png)

andrewtavis:
I’d say the best thing to do would be to hide the text elements that are there are replace them once the transition has finished. How does that sound? Because as of now we also have the activist logo shrinking, which also doesn’t look great. So remove the the texts above and replace them with the new ones, but then how to do that 🤔 

This would need to be triggered on the components themselves likely. Do you want to send a PR and I can take a look, then we discuss from there?

momanyisamuel:
[...]

PR #80, opened 2023-03-18
 Add transition to sidebar collapse #68 

momanyisamuel:
added `transition-all duration-700` to Sidebar, SidebarFooter,SidebarHeader,SidebarMainSectionSelectors and home
also added flex property to center collapse button

### DIFF
--- a/frontend/components/MenuSelector.vue
+++ b/frontend/components/MenuSelector.vue
@@ -9,14 +9,31 @@
     }"
   >
     <Icon :name="iconURL" size="1.25em" />
-    <p
-      v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false"
-      class="px-1"
-    >
-      {{ $t(btnText) }}
-    </p>
+    <Transition>
+      <p
+        v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false"
+        class="px-1"
+      >
+        {{ $t(btnText) }}
+      </p>
+    </Transition>
   </NuxtLink>
 </template>
+
+<style>
+.v-enter-active,
+.v-leave-active {
+  transition: opacity 0.5s ease;
+}
+
+.v-enter-from,
+.v-leave-to {
+  opacity: 0;
+}
+.v-enter-from .inner {
+  transition-delay: 0.25s;
+}
+</style>
 
 <script setup lang="ts">
 defineProps<{
--- a/frontend/components/SearchBar.vue
+++ b/frontend/components/SearchBar.vue
@@ -13,7 +13,8 @@
       type="text"
       placeholder="Search"
     />
-    <div
+    <Transition>
+      <div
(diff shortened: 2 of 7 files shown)


andrewtavis:
Thank you, @momanyisamuel! If the explanation I gave for the components in #68 is enough to work more, then feel free to continue to send along commits :) Really happy about the progress here 😊

andrewtavis:
Hey @momanyisamuel! Let me know how you want to proceed on this issue 😊 Happy to take a look myself :) We're getting to a point where we want to start using the sidebar in some of the directory pages like for the docs, so my focus is going to switch over to finalizing the sidebar soonish :)

momanyisamuel:
Okay so i think its a CSS only issue that requires a different styling strategy so I have been testing different ways to deal with it lemme finalize today and we can discuss

momanyisamuel:
Hi @andrewtavis so sorry for the delay. I think the solution would be to add a delay using the Vue <Transition> component I am not too familiar with VueJs so kindly assist from here. Also for the logo the view box size is too large and causing an issue during the transition.
had no access to the figma file otherwise i would have fixed the icon and logo sizes - I can assist with this just reach out on Matrix

andrewtavis:
How do the logo sizes need to be fixed, @momanyisamuel? Would be happy to do that myself. As of now we're keeping the public designs locked, but there are private designs I could give access to some point (there mostly just sketches of things we could do before they're moved to the public ones).

momanyisamuel:
So basically if the logo is occupying a space of lets say `140px` by `50px` then the SVG logo exported should have a `viewBox` similar to this i.e - `viewBox = <min-x> <min-y> <width> <height>` thus ` viewBox ="0 0 140 50" width="140" height = "50" `. With this you could actually pan or zoom to just the first letter/logo-mark instead of hiding and showing two different SVG's
see here - https://webdesign.tutsplus.com/tutorials/svg-viewport-and-viewbox-for-beginners--cms-30844
you can easily fix this in Figma and ensure you are exporting the correct SVG check here - https://levimcg.com/blog/more-control-over-figma-svg-exports/

andrewtavis:
Sorry for the late reply. I do have the `viewBox` constrains in as you described, but maybe the issue is that I have the SVGs as components rather than just as the SVG files themselves? So within [SidebarHeader](https://github.com/activist-org/activist/blob/main/frontend/components/Sidebar/SidebarHeader.vue) we're referencing `IconsActivist` or `LogosActivist`, where these might need to just be SVGs within `public`? Should I switch that around?

It  makes sense that we should just pan in on the image so just the `a` shows rather than replacing it though 😊 Good idea :)

andrewtavis (approved):
@momanyisamuel, I'm really happy with the progress on this so far regardless of the minor issues from the component/SVG functionality 😊 I'm going to transfer the above comments into the issue, and then we can make some micro tasks to handle all the work that needs to be done :) I feel like the stuff that needs to be done is now more component/SVG based rather than the transition itself.

PR #89, opened 2023-03-30
Search bar height consistent with sidebar

DxVxDOD:


andrewtavis:
Thank you, @DxVxDOD! I’m out at the moment, but will check this when I get home :) Also, consider joining the [public Matrix chat rooms](https://matrix.to/#/%23activist_community%3Amatrix.org) for activist where we’re discussing development 😊

Issue #49, opened 2023-02-02
Close meatball menu when a dropdown option is selected


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The meatball menu within [frontend/components/MeatballMenu.vue](https://github.com/activist-org/activist/blob/main/frontend/components/MeatballMenu.vue) is composed of two different Headless UI components. The first is a [menu](https://headlessui.com/vue/menu), and within that there are two [disclosures](https://headlessui.com/vue/disclosure) that then reveal buttons for changing the theme or language when clicked. When clicking one of the options the menu itself does not close like a normal menu would. This issue is to add this functionality :)

There is a section for Headless UI menus [closing-menus-manually](https://headlessui.com/vue/menu#closing-menus-manually) that likely would be a good reference for this.

### Contribution

I'd be happy to work on this eventually, but others are welcome to contribute or suggest solutions!

Guilherme013S:
Sorry I'm new to contribute to open source, what steps I have to take to get the code in my machine and do these projects?

andrewtavis:
Hi @Guilherme013S :) Thanks for writing! You'd be cloning the repo here `activist-org/activist`. There are [steps to install and run the environment](https://github.com/activist-org/activist#environment-setup-) in the readme 😊 Let me know if you have any questions 🙃

andrewtavis:
You'd first need to fork this repository here with the button at the top right of your screen :) Happy to send along some learning resources for the technologies we use as well 😊

andrewtavis:
One potential for this issue would be to trigger disclosures being closed with a variable on the `@click` of each of the buttons in the drop downs.

purupurupu:
@andrewtavis 
Hello there! It's been a while. I was wondering if I could contribute to this issue?
Also, is this image right?

https://user-images.githubusercontent.com/63239000/232072171-26500cf0-5663-4036-b375-a12beaac4966.mov

andrewtavis:
[...]

PR #100, opened 2023-04-17
#86 fixed logo and search transitions

connorlunsford:
### Contributor checklist

---

### Description

- fixed padding on search bar changing when sidebar transitions
- Added separate transitions to search input and shortcuts to stop clipping
- Changed logo/icon size and style to transition nicer

### Related issue

- #86


connorlunsford:
@andrewtavis, tell me what you think. Not super happy with the icon -> logo transition. I made it so the icon always stays there and the logo just fades in to make it more cohesive. However the icon and logo don't match up 100% so it gets a little thicker as it transitions. Could fix by changing the image unless you can find another way but any way I tried looks a little jarring.

Let me know what you think and feel free to tag me in any other issue you think I can handle. Thanks!

andrewtavis:
Sorry for taking a bit to get to this, @connorlunsford! Has been a very busy week for me :) Checking the [deploy preview](https://deploy-preview-100--activist-org.netlify.app/) I'm seeing both the icon and the logo as well as box around the search bar icon 🤔 I've learned recently in #49 that the deploy preview is going to be what it is that we're getting, regardless of if the behavior we're seeing within a docker run looks fine.

I also agree that the transition isn't ideal... Do you want to check out the [icon designs](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?node-id=10%3A18&t=OPwgkJGmzFlTeJdl-1) and let me know what you'd need me to make to have this working the way you'd like it to? Would be happy to export whatever you need for this!

connorlunsford:
@andrewtavis no worries at all. I'm fairly busy too so this pace works well for me! I didn't notice the box around the search bar too, I'll fix that in a push later today.

In terms of fixing the logo, the options are not optimal in my opinion and I'm not sure if I know enough to figure out how to fix it in the "best" way. These are the options I'm thinking about.
1. Create an icon that's an exact copy of the a in the logo (important part would be the exact same spacing on the top and left side as the logo) so that even though the logo and icon exist at the same time, they overlap without notice.
2. Create a logo that is a copy of the current logo, but without the a (so just "ctivist") and then the a won't overlap at all. I like this better than 1 but it's still not great.
3. Figure out another transition. I've tried a couple different options but all of them just don't look great. Fading in the logo and fading out the icon at the same time creates this weird overlap effect. Hard cut looks far too jarring. I'm not a huge design guy so I would love some help if you had any ideas for a way we can do this well. Thanks!

andrewtavis:
Hey @connorlunsford! I like the ideas! 😊 Let me toy around with it a bit, but either of the options would at least be better than what we have now. Exporting things so they’re exactly what we need is easy enough :)

andrewtavis (approved):
Hey @connorlunsford! I'm gonna bring this in and play around with it a bit more with some others changes I'm going to be making 😊 Will try to figure out the logo transition myself, but will check in for your feedback. I'll be posting a bunch of issues in the coming days as well :) Let me know if you have interest in any of them!

PR #102, opened 2023-04-24
Use ref and use menu instead of disclosure.

purupurupu:
### Contributor checklist

---

### Description

- use headlessUI menu and use ref.
  - disclosure is not used, I think menu only used is simple.
  - The mix of menus and disclosures was difficult to implement for me...

### Related issue

- #49

### DIFF
--- a/frontend/components/MeatballMenu.vue
+++ b/frontend/components/MeatballMenu.vue
@@ -1,14 +1,16 @@
 <template>
-  <Menu as="div" class="relative inline-block text-left">
-    <div>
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
All makes a lot of sense, @purupurupu! I appreciate you working on this issue 😊 Really is a help to get this done, as it's doubtless a pattern that we'll reuse at some point :) Was a smart solution to make the new states for is sub-menus were open, although I can tell this likely was a bit annoying to work out as you said 🙏

PR #105, opened 2023-05-01
fixed icon-logo transition

connorlunsford:
### Contributor checklist

---

### Description

Set the width of the logo to keep it from going small during the transition. Also removed the if statement on the icon to keep it from disappearing.

### Related issue

- #86

### DIFF
--- a/frontend/components/Sidebar/SidebarHeader.vue
+++ b/frontend/components/Sidebar/SidebarHeader.vue
@@ -42,7 +42,6 @@
         }"
       >
         <IconsActivist
-          v-if="sidebar.collapsed == true && sidebar.collapsedSwitch == true"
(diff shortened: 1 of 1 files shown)


connorlunsford:
@andrewtavis can't assign you but here you go. Makes the logo a bit more consistent. The change you made to the logo is definitely better but it still goes a bit bold after the transition. Maybe you can fix it somehow? I'm out of ideas.

andrewtavis (approved):
Ya I removed the width I think on accident, @connorlunsford 🤔 This seems good. Only issue is that we have a slight transparency that causes the lower icon to also show up. I'll maybe just keep the `v-if` to hide it, or will figure out something else :) Bit thing is that it's sliding like the rest 😊

andrewtavis:
No you're definitely right that the `v-if` should stay, but I figured out just setting the width and I'll set the colors explicitly so there's no alpha overlap :)

Issue #107, opened 2023-05-02
Fix sidebar icon positioning


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

An issue with the sidebar that can be found on the [home page](https://activist.org/home/) is that the icons are not centered properly. Especially when the sidebar is collapsed it's apparent that the two icons that are in [components/Sidebar/SidebarMainSectionSelectors.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Sidebar/SidebarMainSectionSelectors.vue) are slightly to the right and the three that are in [components/Sidebar/SidebarFooter.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Sidebar/SidebarFooter.vue) are slightly to the left within their respective containers. It'd be great if these could be centered 😊

This is a `good first issue` and help is welcome for those who want to work on this. I'm happy to support someone who wants to give this a go :)

momanyisamuel:
Hi @andrewtavis I can take a look at this I think

andrewtavis:
Thanks, @momanyisamuel! Will assign you now 😊 Let me know if there are any questions. The edits to the footer icon positions should go right into [components/Sidebar/SidebarFooter.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Sidebar/SidebarFooter.vue) and for the `MainSectionSelectors` it should be editing the [components/MenuSelector.vue](https://github.com/activist-org/activist/blob/main/frontend/components/MenuSelector.vue) :)

momanyisamuel:
@andrewtavis this seems to affect the transition timing slightly but the is no overlapping as before. Maybe you could assist with the transition timing??

andrewtavis:
@momanyisamuel, 599712b does some final adjustments for this. The icons are note perfectly centered, but for me they're maybe a pixel off in certain places. More to the point they look centered 😅 Also we should note that for the future the activist `a` can't be centered on the other icons and should be as it is now - a few pixels over to the left so that the curvature of the `a` can be displaced a bit.

Thanks again for your support! Lots of more issues coming soon as we'll be making the ones for the hackathon next Wednesday in the coming days :D

momanyisamuel:
[...]

PR #112, opened 2023-05-05
 Fix sidebar icon positioning #107 

momanyisamuel:
### Contributor checklist

---

### Description

- to ensure all icons on `Sidebar` are centered introduced css classes `basis-full justify-center gap-1` to `menuSelector`
- wrapped the `menuSelector` icon in a span and also wrapped both the `menuSelector` icon and text in a div similar to `sidebaFooter` -> `DisclosureButton`
- applied a basis to both the icon and text in `menuSelector`
- applied similar styles to `sidebaFooter` as those in `menuSelector` list 
- this was tested by applying outlines on `menuSelector` and related parent elements to ensure all items were centered correctly.

### Related issue

- #107


andrewtavis:
Thank you, @momanyisamuel! I'll check this out later :) :)

andrewtavis:
Quick comment on this, @momanyisamuel :) The icons in the main section selector seem to be inline with the ones in the footer, but all of them seem to still be a little bit over to the left 🤔 What I'm seeing is that we have a space between the icon and the word that would be displayed if the sidebar was expanded, and this space is still being accounted for when it's compressed. As we need the icon to stay in one place as the word is extended, maybe what makes sense is to just put a space before the icons as well? We can then maybe reduce the x-padding if it's not what we need. What do you think? 😊

momanyisamuel:
@andrewtavis The spacing while expanding was being caused by the basis property actually. the icon was growing to occupy the space being left when the text is removed on collapse. i changed this to width and it seems to maintain position. let me know what you think.

andrewtavis (approved):
Looks wonderful, @momanyisamuel! Thanks for coming through again with some great work! 🙏 Plus with the most recent version the little jump that was happening in the text reveal of the main section selectors is gone. Super happy 😊

Issue #125, opened 2023-05-15
Add transition to search bar icon hide/appear


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

After #118 the hotkey indicators are hid when the search bar is focused. This issue is a very simple one to add a transition duration to this change so that it’s no so immediate. The change for this would go into [frontend/components/SearchBar.vue](https://github.com/activist-org/activist/blob/main/frontend/components/SearchBar.vue). 

### Contribution

This issue is a `good first issue` and perfect for someone who wants to get a bit of experience with the code base :) Would be happy to support someone who’s interested!

TeddyGavi:
Hello! Considering @samcdonovan did #118 I don't want to step on any toes, but I would like to take this one on.

andrewtavis:
All good, @TeddyGavi! I realized that a bit of a transition would be good only after merging #118 :) You’re welcome to work on this 😊

andrewtavis:
Assigned and looking forward to the contribution, @TeddyGavi!

andrewtavis:
Closed by #132 😊 Thanks, @TeddyGavi! 🙏 Again pleasure talking to you :) :)

PR #132, opened 2023-05-16
feat: add transition to hidden hotkey indicators

TeddyGavi:
### Contributor checklist

---

### Description

1. Add `transition-opacity transition-duration-200` to the hotKeyIndicators `div`
2. modify the existing CSS conditional `.hide` class from `display: none;` to `opacity: 0;`
3. Remove `focus:w-5/6` from parent input as it was causing a slight layout shift during the transition

Before:
beforeHotkey

After:
afterHotkeyChanges

Layout Shift:
layoutshiftissue

### Related issue

- #125

### DIFF
--- a/frontend/components/SearchBar.vue
+++ b/frontend/components/SearchBar.vue
@@ -7,18 +7,19 @@
       <input
         v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false"
         ref="input"
-        class="w-16 h-5 bg-transparent outline-none focus:w-5/6"
+        class="w-16 h-5 bg-transparent outline-none"
         type="text"
         placeholder="Search"
         @focus="onFocus"
         @blur="onFocusLost"
+        :class="{'focus:w-5/6': isInputFocused }"
       />
     </Transition>
     <Transition name="shortcuts">
       <div
         v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false"
         ref="hotkeyIndicators"
-        class="flex space-x-1"
+        class="flex space-x-1 transition-opacity transition-duration-200"
       >
         <div
           class="w-5 h-5 text-sm text-center rounded-md has-tooltip bg-light-highlight dark:bg-dark-highlight text-light-special-text dark:text-dark-special-text"
@@ -61,6 +62,7 @@ import { ref } from "vue";
 const sidebar = useSidebar();
 const input = ref();
 const hotkeyIndicators = ref();
+const isInputFocused = ref(false)
 const keys = useMagicKeys();
 
 whenever(keys.slash, () => {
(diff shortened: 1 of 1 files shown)


andrewtavis:
Woooot first code committt 🙌 Thanks @TeddyGavi! Will check it later tonight odds are 😊 Excited to have another issue done to get this sidebar looking better and better :)

TeddyGavi:
@andrewtavis I am not 100% sure about this solution! I am sure there is another way. I am new to Vue. My experience has been primarily in Nextjs and React. I am excited to learn and continue to contribute. :smiley:

andrewtavis:
At this stage it doesn't have to be perfect, @TeddyGavi :) I'm also learning Vue/Nuxt myself. We just need to keep at it and slowly more folks will join to exchange ideas and expertise 😊

andrewtavis:
One thing @TeddyGavi :) We actually do need `focus:w-5/6` or a similar style added in as this allows the user to type almost the full length of the search bar. As it is now the search text only fills half of the search bar before it starts to overflow. Aside from that all's wonderful! 🥳 Super smooth timing and just feels great :)

TeddyGavi:
@andrewtavis Ok! I will look into this layout shift further.

andrewtavis:
Thank you, @TeddyGavi! 😊

samcdonovan:
Hey @TeddyGavi, I just wanted to chime in a bit here because I was the one that originally added the code to hide the hotkeys. I tried adding an animation to it much in the same way you have done, but found that using 'opacity: 0' meant that the input bar would not extend to its full length (because the hotkeys container was still affecting the document flow). That's why I used 'display: none' which allowed the input to extend to its full length, but because of the way the 'display' property works, it meant that the animation no longer worked properly. Just thought I'd give a little bit of my thought process around that :)

andrewtavis:
Thanks for giving these insights, @samcdonovan!

TeddyGavi:
Thank you for sharing @samcdonovan I didn't mean to step on any toes here. I believe I figured out a solution. It involves adding both `display: none;` and the conditional class of `.hide` to control the opacity. 

When search input is focused, immediately apply `.hide` aka `opacity: 0;` (with transition) and after a delay of 200ms apply `focus:w-5/6` to input and also at the same time `hidden` aka `display: none;` to the hotkey indicators (to allow for the search input to take up 83.3% of the width :) ). 

when blurred, immediately remove `hidden` and `focus:w-5/6`, then after a 200ms delay remove `.hide`. This is the same logic but in reverse.

 A new commit is incoming. Let me know what you think. I am sure there is a better way and something I missed :thinking:

andrewtavis (approved):
Changes all look good and the functionality is exactly what we're looking for, @TeddyGavi 😊 Thanks for this! 🌙🥳

Issue #146, opened 2023-05-17
Add accessible labels to image links under supporters


james-jenkinson:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

- [x] Add aria label for each of the three images
- [x] Use translation keys for each aria label (values do not need a translation for german or french yet)

In the supporter section, the image links have no `aria-label` defined, this results in them not be accessible to screen reader

andrewtavis:
Thanks @james-jenkinson :) Specifically what we're talking about here is that we need `aria-label` for Wikimedia Deutschland, Wikimedia Serbia and Impact Hub to be added.

andrewtavis:
If someone would like to help with this this would be great! Happy to assist 😊

trishas95:
Hi @andrewtavis I'd like to work on this issue, I'd be happy if you could assign this to me. It would really help if you could explain in brief what I have to do :) 
Thanks

andrewtavis:
Hey @trishas95 👋 Thanks for your interest in activist :) The images that we're talking about are those that are in the [frontend/components/Grid/GridSupporters.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Grid/GridSupporters.vue) component. Within each `<LandingPageLogoImage>` component we need to add in an `aria-label` attribute that will then allow them to be read by screen readers. 

We should use a translatable key for each of the labels such that we can eventually translate them. Those keys should go within [frontend/i18n/en-US.json](https://github.com/activist-org/activist/blob/main/frontend/i18n/en-US.json). The keys for example could be `"wikimedia-de-logo-label"`, `"wikimedia-rs-logo-label"` and `"impacthub-belgrade-logo-label"`. From there we'd assign a value like the following `"Link to the website of Wikimedia Deutschland."` and assign the value in the area label via the following:

```
:aria-label="$t(wikimedia-de-logo-label)"
```

Will assign you now, but let me know how this all sounds! Happy to explain more :)

trishas95:
Okay, so I'll have to add the aria-label(s) to the images along with the appropriate translation keys, and add the values for those translation keys in frontend/i18n/en-US.json ,if I'm not wrong?

andrewtavis:
Exactly :)

[...]

PR #162, opened 2023-05-21
Second

trishas95:
### Contributor checklist

---

### Related issue

- #146

### DIFF
--- a/frontend/i18n/en-US.json
+++ b/frontend/i18n/en-US.json
@@ -39,7 +39,7 @@
   "impress": "Impress",
   "theme": "Theme",
   "wikimedia-de-logo-label": "Link to the website of Wikimedia Deutschland.",
-  "wikimedia-rs-logo-label": "Link to the website of Wikimedia Serbia",
-  "impacthub-belgrade-logo-label": "Link to the website of Impact Hub Belgrade"
+  "wikimedia-rs-logo-label": "Link to the website of Wikimedia Serbia.",
+  "impacthub-belgrade-logo-label": "Link to the website of Impact Hub Belgrade."
 
 }
--- a/frontend/i18n/en-US.json
+++ b/frontend/i18n/en-US.json
@@ -37,5 +37,9 @@
   "about": "About",
   "supporters": "Supporters",
   "impress": "Impress",
-  "theme": "Theme"
+  "theme": "Theme",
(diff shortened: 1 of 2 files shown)


frontend/i18n/en-US.json
```diff
@@ -37,5 +37,9 @@
   "about": "About",
   "supporters": "Supporters",
   "impress": "Impress",
-  "theme": "Theme"
+  "theme": "Theme",
+  "wikimedia-de-logo-label": "Link to the website of Wikimedia Deutschland.",
+  "wikimedia-rs-logo-label": "Link to the website of Wikimedia Serbia",
```
andrewtavis, line 42: Could you add a period here for consistency?

frontend/i18n/en-US.json
```diff
@@ -37,5 +37,9 @@
...
+  "impacthub-belgrade-logo-label": "Link to the website of Impact Hub Belgrade"
```
andrewtavis, line 43: A period here too would be great :)

andrewtavis (changes requested):
Thanks @trishas95! This is basically done. The last thing that's needed is that you need to add a `:` before the aria labels so that the translated values can be read in rather than a string. I'm realizing I should have explained this better. I just pulled in your branch and the correct way of doing this would be:

```
:aria-label="$t('wikimedia-de-logo-label')"
```

Once the fixes I mentioned about adding some periods in the descriptions and changing the `aria-label` attributes is done we'll be good to merge!

trishas95:
Hello @andrewtavis, I made the changes you've mentioned above. This is one my first few PRs, looking forward to learning more! :)

andrewtavis:
Thank you, @trishas95! Happy to help you learn more :) I’ll do the final review tomorrow 😊

andrewtavis:
Hey @trishas95! Have had a few busy days and didn't get a moment to do the final check of this till now. I'm realizing there's still one very small issue here 🙃 You need the elements within the `t()` to be in single quotes, so `t('wikimedia-de-logo-label')`. Sorry I didn't mention this part more explicitly when I discussed the `:` above. As of now what we have there is `translate(wikimedia - de - logo - label)` like it's subtraction, but what we want is `translate('the-string-in-kebab-case')` that will then feed the key into translation and provide the appropriate value :)

trishas95:
@andrewtavis I made the changes :D

Issue #154, opened 2023-05-18
Add and remove links from connect card


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to allow someone to add or remove links to the connections card as seen on [activist.org/organizations/tech-from-below/](https://activist.org/organizations/tech-from-below/). This issue is blocked by #153 where the new field component that would appear on `Add account` being pressed would be made. The goal would be that the add field component would allow users to input new fields, and further that by pressing the `X` icons next to accounts that they could be removed. We'll think about editing in the future :)

Changes for this issue would go within [frontend/components/Cards/CardsConnect.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Cards/CardsConnect.vue) 😊

### Contribution

Happy to work on this or support someone who has interest!

andrewtavis:
Now that #153 is done, this issue is now unblocked and the component for adding in a field can be found at [frontend/components/TooltipNewField.vue](https://github.com/activist-org/activist/blob/main/frontend/components/TooltipNewField.vue) :)

andrewtavis:
Hey @messiuw! Nice to meet you and thanks for your interest in activist :) Happy to assign you as this should be a good issue for a first timer contributor 😊 Please let me know if you have any questions. Would be happy to discuss :)

messiuw:
What is the intended behaviour of the "New account" button? Upon clicking the button, should there be an empty input field to the left (or right) of the button or should a small modal dialog pop up where the user can enter the details?

andrewtavis:
There should be a pop up that happens that allows the user to enter in a new field. It was originally labeled as a tooltip, but I renamed the file and the component can now be found at [frontend/components/Popup/PopupNewField.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Popup/PopupNewField.vue) :) For this leave `descriptionPrompt` blank as we just want the text input, not the text area. Let me know if you have more questions!

messiuw:
[...]

PR #163, opened 2023-05-21
Add and remove links from connect card

messiuw:
### Contributor checklist

---

### Description

Changes introduced in this pull request adress the issue [#154](https://github.com/activist-org/activist/issues/154). It is now possible to add and remove contacts. Smoke tests have been performed.

### Related issue

- #154


andrewtavis:
Hey @messiuw! This is generally working really well 😊 If I delete an account it won't let me enter it in again, but then this is fine as this part of it will be being handled by the backend anyway. My one question is why Flowbite needed to be added in. Is it possible to use [HeadlessUI](https://headlessui.com/) for similar use cases?

andrewtavis:
And no stress about the merge conflicts! I'd be happy to deal with those :)

messiuw:
Thanks for your Review! I mainly added flowbite because I saw that the UI uses tailwind and flowbite is supposed to be designed on top of tailwind. If you rather would like to use headless UI I can check if this will work as well. I'm always eager and happy to learn something new, so if this is not time critical I can check :)

andrewtavis:
Happy to review, @messiuw! It's not time critical as of now, and if you could check switching it to Headless UI that'd be great :) That components framework is directly from the makers of Tailwind, and I'd prefer to keep our dependencies reliant on as few and as large of dev communities as we can 😊 

Thanks so much!

messiuw:
Hi @andrewtavis! I pushed some changes with respect to your most recent comment. Have a look and let me know what you think :) Sorry for the long waiting time, I was quite busy at work these days.

Issue #152, opened 2023-05-18
OpenStreetMap media component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would focus on creating the map media component for events on activist.org. The preference would be that this would make use of the [OpenStreetMap api](https://wiki.openstreetmap.org/wiki/API) so that activist can continue to align its technology with free, open-source communities. It'll also make it easier to work with Wikidata later 🙃

Deisgns for this component can be found in the [designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=62%3A2&t=ydwzf7PjjXr9DPuW-1) and can further be seen below:

![Page Media](https://github.com/activist-org/activist/assets/24387426/1297f8cf-698e-43d1-8434-ff03913efe64)

Generally I'm thinking this component could be called `MediaMap`, but am open to suggestions.

### Contribution

Would be really great to get some help with this! Happy to support as I can 😊

andrewtavis:
@samcdonovan 🙃

samcdonovan:
I think using the OpenStreetMap API is a great idea! And for now I think `MediaMap.vue` is a good name for the component too. I'll get working on it right away!

andrewtavis:
🔥🔥🔥🔥

Thanks, @samcdonovan! Let me know if you need some help! :)

samcdonovan:
So it seems like `OpenStreetMap` doesn't currently offer a way to embed a map on the frontend. I have seen solutions that use either [Leaflet](https://leafletjs.com/) or [OpenLayer](https://openlayers.org/) to create the map using an OpenStreetMap URL. I think I'll look into that later today when I have more time

andrewtavis:
Thanks for the initial research, @samcdonovan! [Leaflet](https://leafletjs.com/) is the one that I've seen before, but let me check both again right now. The ability to map out routes is something to consider in this for a feature that would be really useful and might not be as intuitive in one as it is in the other. So by this I mean that say we want to put the route of a protest onto a map so that people know where they're walking/biking (driving protests where I am are not really the kind we'd want on the platform 😅), and further would be able to more easily catch up if they come late or fall behind.

[...]

PR #164, opened 2023-05-21
feat: Add media map component (#152)

samcdonovan:
### Contributor checklist

---

### Description

1. Added `leaflet`, `@vue-leaflet/leaflet` and `@types/leaflet` dependencies, and updated the `README` to match this.
2. Added the MediaMap component

Below are screenshots of the map:
- Single location/pin:
 Screenshot 2023-05-20 070439
- Multiple locations/pins:
Screenshot 2023-05-21 090422

So currently the MediaMap component takes `addresses` (see below) as a prop (which is an array of strings), formats those addresses and passes them to the `OSM API`, which then returns the latitude and longitude of each location. These are used to display the pins on the map but are also used to calculate average lats and lons between all the addresses for the center point of the map. I realised that this will probably be done in the backend in the future, where the lats and lons will just be directly passed into the frontend, so in the component I provided a commented-out block of code demonstrating how the component may handle that.

MediaMap with props example:
```
  <MediaMap
      :addresses="[
        'Oxford Circus, Oxford St, W1B 3AG, London',
        'Marble Arch, Oxford St, W2 2UH, London',
        'Green Park, Piccadilly, W1J 9DZ, London',
      ]"
    />
```

### Related issue

- #152

### DIFF
--- a/frontend/components/MediaMap.vue
+++ b/frontend/components/MediaMap.vue
@@ -1,0 +1,132 @@
+<template>
+  <div class="flex w-96 h-96 rounded-lg map border p-0 m-auto">
+    <div
+      id="map-div"
+      class="w-full h-full z-0 rounded-lg"
+      ref="map"
+      alt="Map displaying a pin at the location of this organisation or event."
+    ></div>
+
+    <p
+      class="text-2xl text-center z-10 m-auto align-middle p-5 text-light-cta-orange dark:text-dark-cta-orange"
+      :class="{
+        absolute: !errorOccurred,
+      }"
+      :key="rerenderKey"
+    >
+      {{ errorMessage }}
+      <br />
+      {{ sorryMessage }}
+    </p>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { onMounted } from "@vue/runtime-core";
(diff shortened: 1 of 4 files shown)


frontend/components/MediaMap.vue
```diff
@@ -0,0 +1,132 @@
...
+  );
+  leafletMap.addLayer(layer);
+  markers.map((marker: Marker) => {
+    let pin = L.marker([marker.lat, marker.lon], { title: marker.address }); // add location pin to map
+    pin.addTo(leafletMap);
+  });
+}
+
+/*
+    NOTE: Below is an exmaple of the code to use when the backend code
```
andrewtavis, line 79: Thanks for this, @samcdonovan! Glad to have some reference for when we start making the shift here soon 😊

andrewtavis (approved):
This is really some amazing work, @samcdonovan! :) :) A dream for this platform has always been to one day see a map with all the events scattered all over it, and this is the first glimpse 😊 All the code and the calculations to center the map look good to me. We can check this a bit more as we experiment, and I think it makes sense to also make a template event now to allow us to do just that 😇

Issue #122, opened 2023-05-10
Resources search result component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create parts of the search result component that is being made in #119. We want the features made in this issue to be conditionally switched to based on the whether `resource` is passed to the aforementioned component's props :)

The designs for this component can be found on the [Components tab in Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1606%3A585&t=5CR5mIoy4P0OjwjS-1) and can further be seen below:

![Mode=Light, Mobile=False, Type=Resource, Reduced=False](https://github.com/activist-org/activist/assets/24387426/cd30a87f-ffe5-4588-befb-ea4651bfda40)

Mobile:

![Mode=Light, Mobile=True, Type=Resource, Reduced=False](https://github.com/activist-org/activist/assets/24387426/ddcd12d4-8ab6-4ba7-9631-02fdba10525f)

- This issue is blocked by #119 that makes the baseline component

### Contribution

Would be great if someone had interest in working on this! Happy to help, or will get to this myself :)

andrewtavis:
Assigning myself to hold this issue for the team working on it for Code Night 🌙😊

anthonyshibitov:
I’ll take this :)

andrewtavis:
Thanks @anthonyshibitov! Feel free to do whatever icon you see fit for a placeholder resource icon. I’ll get the real one in there in #161 :)

andrewtavis:
@anthonyshibitov, FYI I added the resource icon component via #161 :) You can access that now via `<Icon name="IconResource" />` and also pass arguments like `size`. For coloration you need to use Tailwind's `fill` class property rather than text coloration as works for default icons, but so far it's been working well 😊

Hope all's good with you! :) :)

anthonyshibitov:
@andrewtavis cool cool. i'll get this finished up this week. thanks! :D all's been good, just keeping busy :)

andrewtavis:
[...]

Issue #121, opened 2023-05-10
Events search result component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create parts of the search result component that is being made in #119. We want the features made in this issue to be conditionally switched to based on the whether `event` is passed to the aforementioned component's props :)

The designs for this component can be found on the [Components tab in Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1606%3A585&t=5CR5mIoy4P0OjwjS-1) and can further be seen below:

![Mode=Light, Mobile=False, Type=Event, Reduced=False](https://github.com/activist-org/activist/assets/24387426/2efcc6e5-c1df-4328-846c-1077407a373d)

Mobile:

![Mode=Light, Mobile=True, Type=Event, Reduced=False](https://github.com/activist-org/activist/assets/24387426/b7ae4fa4-558a-43da-955c-67701e725df2)

- This issue is blocked by #119 that makes the baseline component

### Contribution

Happy to provide to support to someone with interest in working on this, or will also get to it eventually 😊

andrewtavis:
Assigning myself to hold this issue for the team working on it for Code Night 🌙😊

andrewtavis:
I'll also make a component for the event icon that has the type color indicator on the left :)

andrewtavis:
@james-jenkinson, FYI there's a component for the event icon now as seen on the [home page](https://activist.org/home/). Basically fills whatever `div` it's put into vertically and is centered at 85% width with the intent being that we'll put it into square containers :)

andrewtavis:
@anthonyshibitov, can you write in here so I can assign this to you? The changes on my end will be going out later this evening :)

anthonyshibitov:
Gotcha

andrewtavis:
Post pull you're good to start working on this, @anthonyshibitov :) Your template event icon is also found in [frontend/components/Icon/IconEvent.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Icon/IconEvent.vue). Let me know if you have any questions!

anthonyshibitov:
Awesome thanks :D (get some sleep!!!!!)

[...]

PR #165, opened 2023-05-22
Safari click area fix

### LINKED ISSUE #72: [Safari] mobile hit area shifting
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Safari is currently having issues with the landing page on mobile. Specifically some of the page elements’ hit areas are not coinciding with their positioning on the page. `LabeldBtns` at times aren’t clickable within their boundaries, and the elements of the supporters section at times have hit areas that extend beyond their boundaries and interfere with other elements on the page being accessible.

anthonyshibitov:
Fix Safari mobile click area bug by explicitly stating the height and width of the Logo components in LandingPageLogoImage. 

### Contributor checklist

---

### Description

Previously, the clickable area was spilling out of the LandingPageLogoImage component.

Since the SVGs are now resized to this 200/250px width, scale-[x] can be mostly removed except to fix small spacing/aesthetic issues. I left scaling in the WikimediaRS and ImpactHub components to match what it looked like beforehand.

This could be further fixed by using SVGs specifically made for the Supporters section, instead of scaling them via tailwind.

---

I've confirmed that this is now working on my iPhone, and the click area is no longer spilling from the logo area.

Note, that I think this is an issue with tailwind, as this isn't reproducible on Safari emulating an iPhone. 

### Related issue

 - Fixes #72


andrewtavis:
Hey @anthonyshibitov! In checking it out this is firstly working really well, but I realized the big thing that's still causing a problem :) Maybe you can check to see if you can also reproduce on your end with the commits in the PR, and if so we can think about how to fix that 😊 

I'm realizing that some of the buttons are not working still, but that's specifically because when you try to click them we're instead clicking the logo in the header (try long holding them and you see that the preview is the landing page itself). The `Request access` button right at the top works for me if I scroll down a bit such that the header disappears, but then if the header's there then I just click that instead and even see the logo blink a bit.

Could you let me know if you're also experiencing this?

anthonyshibitov:
Ahhh, okay.

The header logo button is extending down to around the center of the page for some reason. Is this what you're talking about? Sometimes clicking a button when its underneath this header click-area it will click the header logo instead and make it grey, and not register the actual button click.

Here's a screenshot of it mid load, showing the actual clickable area. In this example I tried clicking the `Request access` button, but it registered clicking the `activist` logo at the top.

IMG_9063

I didn't realize this was an issue, I was just looking at the supporter logos at first. I'll dig into this!

andrewtavis:
> I didn't realize this was an issue, I was just looking at the supporter logos at first. I'll dig into this!

Ya I didn't think of it until I realized that it wasn't happening when the header was hidden :) Thank you! 🙌

andrewtavis:
I'm going to bed now 😅 I'd try adding `overflow-clip` to the logo's class if you haven't already :) This tends to restrict the bounds appropriately 🤞

andrewtavis:
If not to the logo, than to the header's class :)

anthonyshibitov:
I believe it already has `overflow-clip`, no?

https://github.com/activist-org/activist/blob/main/frontend/components/WebsiteHeader.vue#L16

anthonyshibitov:
Alrighty, check out this commit and let me know what you think.

I tried adding `overflow-clip` to a few different elements and couldn't get the desired behavior.

Let me know if `overflow-clip` works instead somewhere else - I think that's probably a better way to fix this than explicitly stating the dimension the SVG _should_ be. I'll revert this commit and commit the `overflow-clip` change if that's preferred, just let me know! :)

It's weird though. Even though in the Chrome/Mac Safari element inspector it shows that the SVG is expanded beyond it's parents container, it still doesn't register it as a click - only on mobile Safari.

Before
Screenshot 2023-05-23 at 8 53 45 PM

After
Screenshot 2023-05-23 at 8 53 59 PM

andrewtavis:
Let's definitely ignore my `overflow-clip` suggestion :) This is really great! The one thing is that we'd be applying the size to the logo file itself rather than to the instance in the mobile header, so it's thus shrinking it to being too small on desktop. Maybe applying the style [directly on the mobile header](https://github.com/activist-org/activist/blob/71eff59534cd1f088f160bfd705dc07f28ec557a/frontend/components/WebsiteHeader.vue#L15)? We're also getting rid of some of these component SVGs soon, btw. I historically had a hard time getting them to change color with the system color scheme, but there are better ways of doing this than having them be components.

anthonyshibitov:
So unfortunately the only way I found this was able to work was by changing the SVG class directly. 

I’ve tried passing a `isMobile` prop to the logo component and this seems to work, but it might be a bit verbose. Let me know if there’s a better way of doing this. Thanks!!

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
andrewtavis, line 1: Looks like you yarn installed in the root :) I'll go ahead and merge and delete the file 🙃

andrewtavis:
Checked with that friend about whether you can pick up #121 or #122, btw. Will let you know and then you'd just need to write in the one that we decide on so that I can assign you 😊

I have some local changes for that file that I'll send later on tonight with a bunch of updates. Should be good to start tomorrow :)

anthonyshibitov:
Sounds good!

PR #168, opened 2023-05-24
Move MediaMap to /media, add different colour marker for 'act' events

samcdonovan:
### Contributor checklist

---

### Description
- Moved MediaMap.vue to the /media folder
- Added the 'type' prop (act or learn)
- Changed the colour of the marker/pin to red if the type is 'act'
- Changed the OpenStreetMap URL to a https URL (fix)

act:
Screenshot 2023-05-24 165743

learn:
image

I did the change in colour using `filter: hue-rotate(145deg)`, a CSS style. Leaflet does not currently provide a way to change the colour so the only options were getting different icons from elsewhere (e.g. [leaflet-color-markers](https://github.com/pointhi/leaflet-color-markers)) or using the CSS method that I used. I did try using the icons from [leaflet-color-markers](https://github.com/pointhi/leaflet-color-markers) but found that the markers were slightly out of position with these icons.

### Related issue

- #152


samcdonovan:
@andrewtavis I just tested the map in the deploy preview and it seems to be working now. The only issue is that the marker is no longer displaying for some reason. It could have something to do with the method I used for the changing the colour (`filter: hue-rotate(145deg)`) but I'm not sure

andrewtavis:
Thanks for this, @samcdonovan! I’ll be able to review later this evening :) We can read up/experiment a bit and figure it out 😊

andrewtavis (approved):
Thanks for all this, @samcdonovan! I'm going to take it in and play around a little bit with the answers in [this SO question](https://stackoverflow.com/questions/23567203/leaflet-changing-marker-color) :)

We'll ultimately need to do custom colors at one point as well given that dark mode will need to change the map layers. Will let you know if I can figure something out!

Issue #170, opened 2023-05-26
GitHub community shields for landing page


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create the GitHub community shields for the GitHub that we can display on the landing page. The designs for these can be seen on the [designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=517%3A3258&t=MDQbwmpLwutjdUXH-1) and can further be seen below:

![GitHub Stats](https://github.com/activist-org/activist/assets/24387426/1d2bde12-d770-4b58-a812-8dd7e7250397)

Each of these shields can be an option for a general component that would be in the [icons directory of the frontend](https://github.com/activist-org/activist/tree/main/frontend/components/Icon). For now we can call it `IconGitHubShield` 😊 We'd need to access the GitHub API to get the information for the shields, with how we exactly do this being open for discussion.

### Contribution

Happy to work on this myself or support someone who has interest :) This should be a relatively doable `good first issue` with another section being added to the landing page to show them off and talk about open source development :upsi

rohram04:
Hello I would be willing to work on this. In terms of accessing the github api - would a simple fetch request work. If you make a request to   https://api.github.com/repos/activist-org/activist it gives the following json result:

`{
    "id": 332064599,
    "node_id": "MDEwOlJlcG9zaXRvcnkzMzIwNjQ1OTk=",
    "name": "activist",
    "full_name": "activist-org/activist",
    "private": false,
    "owner": {
        "login": "activist-org",
        "id": 79784013,
        "node_id": "MDEyOk9yZ2FuaXphdGlvbjc5Nzg0MDEz",
        "avatar_url": "https://avatars.githubusercontent.com/u/79784013?v=4",
        "gravatar_id": "",
        "url": "https://api.github.com/users/activist-org",
        "html_url": "https://github.com/activist-org",
        "followers_url": "https://api.github.com/users/activist-org/followers",
        "following_url": "https://api.github.com/users/activist-org/following{/other_user}",
        "gists_url": "https://api.github.com/users/activist-org/gists{/gist_id}",
[...]

PR #179, opened 2023-06-02
feat(home page): Metrics component initial commit

rohram04:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
  
  

---

### Description

Created a requested feature "At a glance metric component" in issue #171. The card component is located in the CardsMetrics.vue file under the grid component directory "frontend/components/grid/" along with the GridCardsMetrics.vue file which is a component with a sample grid using the CardsMetrics component.

I created another vue project in codesandbox.io to develop this feature. It is mostly tailwind.css classes applied to a component so no rigorous testing is required. Here is a link for the sandbox https://codesandbox.io/p/sandbox/lingering-meadow-enxeij?file=%2Fsrc%2Fcomponents%2FCardsMetrics.vue%3A13%2C77. Run "npm run dev" in a new terminal window and view the component in the host window codesandbox creates.

### Related issue

- #171

PR #181, opened 2023-06-06
feat(landing page): Initial commit of github icon shields

rohram04:
### Contributor checklist

---

### Description

Created Github Icon shields as requested in issue #170 like so:
image

The shields are currently in a grid component and the leftmost visit us shield is actually an <a> tag linking to [https://github.com/activist-org/activist](https://github.com/activist-org/activist).

Please note that it is probably better to separate each of these icon shields into their own component files. Also I utilized vue.js options api as I was having trouble getting the composition and normal api to work.

### Related issue

- #170


andrewtavis:
Thanks for your strong initial dedication to activist, @rohram04! 😊 Looking forward to getting these on the landing page :) :) Will review later today odds are.

andrewtavis (changes requested):
Hey @rohram04, checking this one out first 😊 I agree with you on the following:

> Please note that it is probably better to separate each of these icon shields into their own component files.

Do you want to make three different icon files within the `Icon` directory and then a grid component within the `Grid` directory that houses them all? And as far as links are concerned, maybe we can do a link to the [forks of activist](https://github.com/activist-org/activist/forks) for the forks link and the [stars for activist](https://github.com/activist-org/activist/stargazers) for the stars one?

A final note would further be could we get `Visit us on` a little bit over to the right? I know that it's inline, but given the curvature of the letters it looks like it's to the left a bit :)

Thanks so much for the efforts here! I'll jump over to #179 right now :)

rohram04:
Hello, I made some changes to the files. Because the components are so similar I decided to make an IconGithubShield component and then use the <slot> tag to pass in child components (svgs and other divs), instead of creating 3 separate components. This reduces repetition. Let me know if this approach works for you.

andrewtavis (approved):
Thanks for the quick edits to the components, @rohram04! Will play around with these a bit and then add the section in to the landing page 😊 Will write in the issue when it's been added :)

Issue #129, opened 2023-05-15
Organization about component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create parts of the about component that is being made in https://github.com/activist-org/activist/issues/127. We want the features made in this issue to be conditionally switched to based on the whether `organization` is passed to the aforementioned component's props :)

The designs for this component can be found on the [Components tab in Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1606%3A585&t=5CR5mIoy4P0OjwjS-1) and can further be seen below:

![Org Information](https://github.com/activist-org/activist/assets/24387426/8d05795e-093b-4ca2-a511-03fc9092444b)

Mobile:

![About](https://github.com/activist-org/activist/assets/24387426/507e2d10-d060-4743-8aa4-3afbfb381556)

- This issue is blocked by https://github.com/activist-org/activist/issues/127 that makes the baseline component.

### Contribution

Will be working on this with people in the coming days :) Happy to support or work on it myself 😊

andrewtavis:
Note that this component and the one in #128 need to have the same height for consistency.

andrewtavis:
Assigning myself to hold this issue for the team working on it for Code Night 🌙😊

momanyisamuel:
Hi @andrewtavis gladly work on this as well

andrewtavis:
762a874 finalized this and #128, @momanyisamuel :) I'll continue the work on this in #198 for the finalized versions at various breakpoints :)

Btw, as of now the `line-clamp-1` CSS isn't working, but by the looks of it we just need to update the version of `nuxtjs/tailwindcss`. We likely don't need `tailwindcss` in the dependencies by itself either :) You'd be welcome to look into these updates and getting the clamp text to work properly with just the Nuxt Tailwind dependency in another issue.

Issue #128, opened 2023-05-15
Event about component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create parts of the about component that is being made in https://github.com/activist-org/activist/issues/127. We want the features made in this issue to be conditionally switched to based on the whether `event` is passed to the aforementioned component's props :)

The designs for this component can be found on the [Components tab in Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1606%3A585&t=5CR5mIoy4P0OjwjS-1) and can further be seen below:

![About Section](https://github.com/activist-org/activist/assets/24387426/4acd9f46-848a-461d-b952-691d815d2641)

Mobile:

![Event Info](https://github.com/activist-org/activist/assets/24387426/da201688-8c4f-40cd-ae37-d7ba12a4df28)

- This issue is blocked by https://github.com/activist-org/activist/issues/127 that makes the baseline component.

### Contribution

Will be working on this issue and other about card related ones in the coming days. Happy to focus on this myself or support others who want to work on it 😊

andrewtavis:
Assigning myself to hold this issue for the team working on it for Code Night 🌙😊

andrewtavis:
Assigned! Thank you, @momanyisamuel!

andrewtavis:
Feel free to write in #129 and I’ll assign you as well 😊

andrewtavis:
Closed via 762a874. Thank you, @momanyisamuel! 🚀😊

PR #200, opened 2023-06-24
Event about component #128

momanyisamuel:
### Contributor checklist

---

### Description

Added missing elements for CardAbout.vue

### Related issue

-  #128 
- #129

### DIFF
--- a/frontend/components/Card/CardAbout.vue
+++ b/frontend/components/Card/CardAbout.vue
@@ -2,31 +2,60 @@
   <div
     class="relative flex flex-col items-center w-full h-min card-style sm:items-start"
   >
-    <div class="bg-slate-400 w-10 h-10 sm:w-16 sm:h-16 absolute right-0 m-3 rounded-md"><img src="" alt=""></div>
+    <div
+      class="bg-slate-400 w-10 h-10 sm:w-16 sm:h-16 absolute right-0 m-3 rounded-md"
+    >
+      <img src="" alt="" />
+    </div>
     <div class="px-4 py-3 sm:px-2 sm:py-2 w-full">
       <div class="flex justify-start items-center gap-3 mb-3">
         <h3 class="text-left responsive-h3 font-display">About</h3>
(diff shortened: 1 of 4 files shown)


frontend/components/Card/CardAbout.vue
```diff
@@ -1,11 +1,38 @@
...
+      <div class="flex justify-start items-center gap-3 mb-3">
+        <h3 class="text-left responsive-h3 font-display">About</h3>
+        <Icon name="bi:pencil-square" size="1.2em" />
+      </div>
+      <button
+        class="px-2 py-1 border-2 border-light-cta-orange bg-[#F1993D4D] text-light-cta-orange rounded-2xl flex items-center gap-2 mb-3"
+      >
+        <Icon name="bi:globe" size="1.2em" />
+        Housing and Community
+      </button>
```
Mradosi, line 16: I suggest using the TopicMarker component instead, as it is already created and styled and used in other places.

momanyisamuel: oh cool thanks, will make the changes, still working on this for #129

andrewtavis: Thank you, @Mradosi!! :)

momanyisamuel: @andrewtavis @Mradosi  Added requested changes above and some more changes for #129

andrewtavis: Totally fine with updating Tailwind, @momanyisamuel :) Do you want to send that change along as well?

momanyisamuel: done 👍

andrewtavis (approved):
Thanks for all this, @momanyisamuel! I think that we may need to hold off on `line-clamp` for now as it might not be available in [Nuxt Tailwind](https://tailwindcss.nuxtjs.org/), but let's keep it in mind for later :) I'll investigate, and we can invest some more time into this later when we're closer to beta 😊

Issue #151, opened 2023-05-17
Improve the alternate text of all buttons


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to go through and add better alternate texts for all buttons on the website. These texts should be added to the [English i18n file](https://github.com/activist-org/activist/blob/main/frontend/i18n/en-US.json) so that we can remember to localize them. As an example: the `Learn more` buttons on the home page all need alternate texts that are more like, "Learn more about getting involved in an activist organization."

### Contribution

This is a great first issue for someone who'd like to get some experience with the project :) Happy to support someone who has interest!

Gauravvpnd:
Hey, i want to work on this issue can you please explain me a little?

andrewtavis:
Hey @Gauravvpnd! Thanks for your interest in activist :) This issue is specifically about accessibility for people who need screen readers. An issue with the website right now is that if someone is using a screen reader they would hear a button like `Learn more`, but then they wouldn't necessarily know what they'd be learning more about without investigating other parts of the page or just choosing to use the link and find out. 

It'd be great if there would be alternate texts for people with screen readers that were more descriptive of the eventual destination, so as in my example we could have the alternate text for the screen reader be "Learn more about getting involved in an activist organization."

I think the easiest way of achieving this would be to add a prop to [frontend/components/LabeledBtn.vue](https://github.com/activist-org/activist/blob/main/frontend/components/LabeledBtn.vue) that would take in the alternate text and assign it to the button. From there we could then add some longer texts to the [English i18n file](https://github.com/activist-org/activist/blob/main/frontend/i18n/en-US.json).

Let me know if this sounds like something you'd be interested in working on! Would of course be happy to explain more 😊 If it does sounds like something you want to work on, then I'd be happy to assign you :)

Gauravvpnd:
Thanks for explaining this sounds interesting I would love to work on this. you can assign me this

andrewtavis:
Assigned! Let me know if you need some help :)

Gauravvpnd:
> 

I have created a pull request doing the changes. You can have and look at and merge it.

andrewtavis:
Thanks @Gauravvpnd! Will check this in the coming days :) :)

[...]

PR #204, opened 2023-06-27
feat: improve alternate text of buttons on index page

luuu-xu:
### Contributor checklist

---

### Description

- added conditional rendered aria-label to BtnLabeled based on prop called alternateText
- LandingPageContent now has btnAlternateText1 and btnAlternateTest2 as optional props for rendering aria-label to buttons
- index.vue receives some according props
- LandingPageSplash's button also has aria-label now but I am not sure whether to keep it
- i18n en-US.json has according button aria-label texts

### Related issue

- #151


andrewtavis (approved):
This is really great, @luuu-xu! I just tried out the buttons with VoiceOver and it worked really well 😊 Thanks so much for your efforts here!

I'll merge and then write to you in the issue :)

PR #206, opened 2023-07-01
Updated Documantion

cicada0007:
## Description

This pull request updates the README.md file and requirements.txt file 

In the `README.md` file, a new section  has been added. This section provides step-by-step instructions on how to install the required Python libraries. It guides users to clone the repository and then install the dependencies using `pip` by running 
`pip install -r requirements.txt`.

The `requirements.txt` file has been updated to include the latest versions of the Python libraries. Each library is listed on a separate line in the format `library_name==version_number`. This ensures that users can easily install the necessary libraries with the correct versions.

These updates aim to enhance the user experience by providing clear instructions on installing the required Python libraries and ensuring that the latest versions are used for compatibility and potential bug fixes.

Please review this pull request and consider merging it into the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,27 @@ choco install docker-for-windows  # for windows
 
 You can then visit <http://localhost:3000> to see the development frontend build once the container is up and running.
 
+### For Backend file
+
+To set up the environment and install the necessary dependencies for the backend, please follow these steps:
+
+
+
+Change directory to the cloned repository:
+
+```
+cd activist
+cd backend
+```
+
+Install the required Python packages by running the following command
+
+```
+pip install -r requirements.txt
+Make sure you have Python and pip installed on your system before running this command.
+```
+
+
 ## Ways to Help [`⇧`](#contents)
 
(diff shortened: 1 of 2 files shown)


andrewtavis (changes requested):
Hey @cicada0007 👋 Appreciate your interest in the project and wanting to improve the documentation. I'm honestly not seeing too much of a need for these additions though. New contributors are already told to install Docker above before they get to the command and are also told to check out Docker desktop if they're new to container usage, so the Windows specific install isn't needed. Beyond this, as we're using Docker, the person doesn't actually need to install the `requirements.txt` file as it should be installed in the container instead.

If you want to send along new commit that just updates `black` and `Django` then I'd be ok with merging. Please remove the `Pillow` install though as we likely will use it, but I'd prefer that we add it when it's used in the codes.

Thanks again!

cicada0007:
hi @andrewtavis 

Thank you for helping me out,  I have done the changes as you have suggested

andrewtavis (approved):
Thanks for the quick fixes and the dependency updates, @cicada0007! 😊 We'll be making more issues in the coming weeks after the backend is setup, so I hope you check back in with us then!

Issue #138, opened 2023-05-16
Event about page layout


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create the layout that will be used to populate event about pages. The layout should include the sidebar, the footer, and the cards to describe the event including about, a page media map (a placeholder is fine for now), connect, join, etc. There also should be a title and an optional tagline (without a tagline the page elements should still be in the same place).

An example of this layout can be seen in the [designs in Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=766%3A214&t=sYF0QkAOBUxfAMJd-1) as well as below:

![Event About](https://github.com/activist-org/activist/assets/24387426/52f3dc00-3034-422b-a5b5-9aa432a999ec)

### Contribution

Would be great to work on this issue with someone!

andrewtavis:
This issue is blocked until some of the components needed are finished :)

andrewtavis:
Assigning myself to hold this issue for the team working on it for Code Night 🌙😊

andrewtavis:
Hey @samcdonovan! The required components aren't finished as of now, but you could start working on the basics of this like including the sidebar, header, tagline and the various buttons at the top?

andrewtavis:
If you're in the mood for something more ambitious, the map view is something that's definitely not finished :) We'd need to find a module or another kind of dependency to get it all up and running, so some research into that would also be super helpful 😊 Whatever sounds most fun 🙃

samcdonovan:
Sure, I'll do the basic layout for now and once all of the required components are done, I can put it all together.

Yeah, I'd also definitely be up for doing the map view! I guess the most obvious choice would be to use the [Google Maps API](https://developers.google.com/maps/documentation) (which I have a bit of experience with) but if you're against using that I can research other options?

andrewtavis:
Wonderful :) :) Let me know if there's anything that I can do to help with the layout. You have most of the base components, and we're just waiting on people to finish the work from Code Night 😊 Will assign you now 🥳

[...]

PR #208, opened 2023-07-03
feat: Add event page layout (#138)

samcdonovan:
### Contributor checklist

---

### Description
Event about page layout; essentially a normal Vue page right now. Open to any suggestions!

Differs slightly from the original design outlined in the #138 but those differences are due to components which were designed differently over the course of the project. Screenshot (in dark mode) below for context:

image

### Related issue

- #138

### DIFF
--- a/frontend/layouts/event.vue
+++ b/frontend/layouts/event.vue
@@ -1,0 +1,121 @@
+<template>
+  <div
+    class="text-light-text dark:text-dark-text bg-light-content dark:bg-dark-content"
+  >
+    <Head>
+      <Title>{{ title }}</Title>
+    </Head>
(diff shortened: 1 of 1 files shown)


andrewtavis:
Really wonderful at first glance, @samcdonovan! Will give it a more thorough review soon :) Thanks so much!

andrewtavis (approved):
Sorry that this took so long, @samcdonovan :) I think that the commit here is great and is generally what we're looking for. We'll have a bit more of an idea how this all works once the backend is fully up and running, but then that should be soon 🙃

Thanks for the commit! I'll do some minor fixes for this and some other issues and then finalize in the issue.

Issue #207, opened 2023-07-01
Baseline frontend-backend connection


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue focuses on doing a basic connection of the activist frontend and backend. The goal is to add in a few of the basic data schemas as defined on [Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377%3A33014&mode=design&t=VeTcTBT4dlBn2XkQ-1) and then reference the information in the frontend as a proof of concept.

### Contribution

@momanyisamuel asked to work on this 🙏🚀 I'll be happy to work on it a bit, review and also will do a call with @momanyisamuel next week to go over the results :)

@wkyoshida maybe you'd be willing to review with me?

wkyoshida:
Yup! Sure can :+1:

andrewtavis:
Once the PR for this is merged I'm going to look a bit into implementing [Django RESR fromework](https://www.django-rest-framework.org/) :)

andrewtavis:
Closed via #209 🚀 Thank you, @momanyisamuel! Glad to have these decisions and the basis behind us. We'll be able to speed up from here :) :)

Issue #174, opened 2023-05-26
activist backend planning


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue will serve as a discussion of how to implement the backend of activist in [Django](https://www.djangoproject.com/). Topics could include:

- Structure of the backend
- Data structures
- How to connect it with the frontend effectively
- Dependencies to add
- Switching the deployment from Netlify
- ...

Once specific tasks are defined by the community we can then make issues and begin on development 😊 Thanks all for any ideas!

andrewtavis:
For this point above:

> - Switching the deployment from Netlify

It would be really good to define the exact set of changes that would happen for the change to need to happen :)

andrewtavis:
Another thing to account for is the many ways that user will be able to sign up:

- User name, email, password and verification
- Username, password, secure verification (in person or via encryption service outside of activist)

5ud03r5:
Hey @andrewtavis! I would love to work on backend for activist. My first question here would be why do we want to proceed with such big framework as django? FastAPI seems to be much suitable for Activists.
Then to migrate it from netlify we can use any cloud provider like GCP to structure it accordingly.
For deployment integration we can use Github Actions or GCP build functionality.

Let me know if you like it and want to hear more regarding implementation/structure details with tools I suggested above!

andrewtavis:
Hey @5ud03r5! Great to hear from you! 😊

Generally with Django we're thinking that a lot of the data control and security that's built in would be something that would lend itself to platform that has to focus on security. With that being said, FastAPI was the other option as Flask doesn't seem to offer enough out of the box functionality. I agree that the templating from Django is a bit overkill though and FastAPI lends itself a bit better for integration with Vue. Not sure. I honestly was thinking more to go with the more popular one and I really shy away from larger frameworks that are not org managed like FastAPI that has a [very shallow dev base](https://github.com/tiangolo/fastapi/graphs/contributors). 

Regarding Netlify shift: the big thing is that we're trying to avoid as many larger firms like Amazon or Google as possible, but then let me say that we will use their services if it comes to it and there isn't a valid open-source alternative. GCP would likely have the edge over AWS, but then the goal is for neither. Glad to have you in this discussion though!

[...]

PR #209, opened 2023-07-04
 Baseline frontend-backend connection #207 

momanyisamuel:
### Contributor checklist

---

### Related issue

- #207


andrewtavis:
@wkyoshida, post call we're still getting some issues with Docker not working for the database. I'm going to look into a few example projects to see if there are some suggestions for this, but would be great if you can see some errors in the `docker-compose.yml` from the get go :)

docker-compose.yml
```diff
@@ -8,7 +8,9 @@ services:
     container_name: django_backend
     ports:
       - "8000:8000"
-
+    env_file:
+      - ./backend/backend/.env  # Path to the .env file containing the environment variables for Django
```
wkyoshida, line 12: I see that there is a `backend/backend/.env.example` file that was created. Should that be what is referenced here? Or should the file be renamed to `backend/backend/.env` instead?

docker-compose.yml
```diff
@@ -17,5 +19,15 @@ services:
...
+
+  db:
+    image: postgres:15
+    container_name: postgres_db
+    environment:
+      - POSTGRES_DB=${DATABASE_NAME}
+      - POSTGRES_USER=${DATABASE_USER}
+      - POSTGRES_PASSWORD=${DATABASE_PASSWORD}
+    ports:
+      - "5432:${DATABASE_PORT}"
```
wkyoshida, line 33: Should `DATABASE_PORT` rather be `5432`? Or was there a reason to choose `3306`?

backend/apps/views/__init__.py
```diff
@@ -0,0 +1 @@
+from .organizations import *
```
wkyoshida, line 1: A more general comment, but I would see to avoid omitting a newline character at the end of files. Just always add one.

This is more due to some tools having issues processing files, since some of them rely on a newline character to determine the end of the file. Does not always happen, but it can, which is why GitHub usually adds this warning.

wkyoshida (commented):
Not a thorough review, but just a couple of items that jumped out at me at first glance
:rocket:

docker-compose.yml
andrewtavis: I've seen @momanyisamuel's convention other places as well – that there's an `.env.example` file that people would need to convert to their own `.env` file. Would you have another suggestion on this? I guess for dev purposes we can share an admin pass and there needs to be another one that's hidden what's live?

andrewtavis:
> Not a thorough review, but just a couple of items that jumped out at me at first glance 🚀

Ya no stress, I can do the full review!

backend/apps/views/__init__.py
andrewtavis: Ya I think at one point we need to do a sit down for the dev community that's focussed on what tools we're using in our environment. Specifically we all should be using the same formatting process, and this process should enforce having a line at the end of the file on save. TBH I have not consciously done this in years, and if I didn't have an extension doing it for me I probably would forget :) What it is that we agree on can then be added to [STYLEGUIDE.md](https://github.com/activist-org/activist/blob/main/STYLEGUIDE.md) as well as [.vscode/extensions.json](https://github.com/activist-org/activist/blob/main/.vscode/extensions.json). This wouldn't be a full dev sync, but maybe 15 min in one where people are told to prepare before. Maybe the one after next as the next one will be quite busy with backend again?

For the above an idea on reinforcing this would be to have another workflow similar to the [maintainer checklist](https://github.com/activist-org/activist/pull/209#issuecomment-1620588231) that would fire on first PRs to the repo. We could shift the check of the commit messages to this as it really just needs to be done once, and beyond that we could have a point that directs people to the style guide and links to the extensions that we as a community think they should be using.

Plus we can all take a look at @momanyisamuel's VS Code setup and marvel at his sidebar on the left 🤪😅😊

docker-compose.yml
momanyisamuel: Yeah the `.env.example` can just be renamed to `.env`

docker-compose.yml
momanyisamuel: Actually this correct will make the changes

docker-compose.yml
andrewtavis: We'd need to remove `.env` in the [.gitignore](https://github.com/activist-org/activist/blob/main/.gitignore) then :)

andrewtavis: And how would we hide the actual values later on when we need them?

backend/backend/.env.example
```diff
@@ -0,0 +1,6 @@
+DEBUG=True
```
andrewtavis, line 1: @momanyisamuel, we still need to delete the one that's here, right?

andrewtavis:
I found the following two repos that might serve as a good reference for some of the settings files, btw:
- https://github.com/melehin/django-celery-vue-template
- https://github.com/MakSzafa/docker-django-vue-postgres-project-template

docker-compose.yml
wkyoshida: > I guess for dev purposes we can share an admin pass and there needs to be another one that's hidden what's live?

Yup - I would think that, with this dev instance, people would really only run it locally. So the credentials would be "shared" in the sense that it would just be the same one for anyone spinning it up and running it locally. 

> And how would we hide the actual values later on when we need them?

That would depend on the production deployment strategy. Deploying within containers, [Docker Compose secrets](https://docs.docker.com/compose/use-secrets/) can be used. Where to store them beforehand and how to fetch them though depends on the deployment strategy, e.g. AWS has a [Secrets Manager](https://aws.amazon.com/secrets-manager/). Seems like Netlify has one [alternative](https://www.netlify.com/blog/a-guide-to-storing-api-keys-securely-with-environment-variables/). GitHub has their own [secrets manager](https://docs.github.com/en/actions/security-guides/encrypted-secrets), so they could be used potentially within GitHub Actions to fetch the secrets, build the production containers, store the secrets inside the containers, and then deploy the containers.

i.e. TBD, but there are various options. Important thing though, as you're likely alluding to, is to not expose them as regular environment variables.

backend/backend/.env.example
momanyisamuel: Yes this will be removed as discussed

momanyisamuel:
Hi @andrewtavis please review and see if this is good to merge

docker-compose.yml
Vince155: I think another option for storing secrets would be to encrypt them using something like [Mozilla SOPS](https://github.com/getsops/sops#encrypting-an-existing-file) that way a file like `.env.example` can be encrypted as `.enc.env.example`.

andrewtavis:
Sorry for this taking so long, @momanyisamuel. There's a lot I still need to do to get settled into the new place + work + etc. I'll review this this evening :)

backend/apps/models.py
```diff
@@ -1,3 +1,308 @@
...
 
 # Create your models here.
+# New user model
+class User(AbstractUser):
+    password = models.CharField(max_length=255)
+    location = models.CharField(max_length=30, null=True)
+    description = models.TextField(max_length=500, null=True)
+    verified = models.BooleanField(default=False)
+    verification_method = models.CharField(max_length=30, null=True)
+    verification_patner = models.ForeignKey(Support, on_delete=models.CASCADE, null=True)
```
andrewtavis, line 13: `verification_partner` :)

backend/apps/models.py
```diff
@@ -1,3 +1,308 @@
...
+    
+class Format(models.Model):
+    name = models.CharField(max_length=255)
+    creation_date = models.DateField
+    description = models.TextField(max_length=500)
+    last_updated = models.DateField
+    deprecation_date = models.DateField
+    class Meta:
+        verbose_name = 'format'
+        verbose_name_plural = '1. Format'
```
andrewtavis, line 105: Could you let me know why we're doing `1./2. CAP_VERBOSE_NAME` for the `verbose_name_plural`, @momanyisamuel? Is this something to format it for the admin panel?

backend/apps/models.py
```diff
@@ -1,3 +1,308 @@
...
+    def __str__(self):
+        return self.creation_date
+
+class OrganizationResource(models.Model):
+    index: models.IntegerField
+    org_id: models.ForeignKey(Organization, on_delete=models.CASCADE)
+    def __str__(self):
+        return self.index
+
+class OrganiZationMember(models.Model):
```
andrewtavis, line 279: `Organization` without the `Z` :)

backend/apps/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from . import views
+
+app_name = 'apps'
```
andrewtavis, line 5: `app_name` doesn't appear to have been used?

backend/apps/views/__init__.py
andrewtavis, line 1: For Django do we have to do explicit imports within `__init__.py`? For packages I read a while ago that this should not be done.

backend/apps/urls.py
andrewtavis: Or is this a variable that's set here but read elsewhere? I'm seeing that we have `apps.User` and `apps.url`.

docker-compose.yml
```diff
@@ -17,5 +26,15 @@ services:
       - ./frontend:/app
       - /app/node_modules
     ports:
-      - "3000:3000"
+      - "4000:4000"
```
andrewtavis, line 29: Just out of curiosity, why does the frontend port need to be switched to `4000`? If this change does go through, we need to make sure to update the readme and the contributing guidelines :)

andrewtavis:
General review's all done, @momanyisamuel :) Mostly is questions for the implementation and some minor minor changes being needed. Happy to merge once these are answered/fixed and then we can continue the conversation in #174 😊

docker-compose.yml
andrewtavis: With the decision to leverage supabase I think checking out their [secrets management](https://supabase.com/docs/guides/functions/secrets) would also be good :) I'll look into this more over the weekend after this is merged 😊

backend/apps/models.py
momanyisamuel: yes this is for the admin panel

backend/apps/views/__init__.py
momanyisamuel: Well i think this will probably change with the use of Django REST framework

docker-compose.yml
momanyisamuel: will change this to 3000

momanyisamuel:
Made changes as reviewed @andrewtavis please let me know if the changes are okay

backend/apps/views/__init__.py
andrewtavis: Totally fine then :) Just checking based on prior experiences and the love of not having import statements in Nuxt ❤️🙃

andrewtavis (approved):
We've got ourselves a backend!! 😊😊🚀 Thanks so much for the hard work and patience on this, @momanyisamuel! I'll have some time on Saturday to work on this a bit more. Would be great if you could sketch some future steps in #174 :)

momanyisamuel:
Great glad this is done @andrewtavis

Issue #216, opened 2023-07-07
Documentation: Typecheck process should include `yarn run postinstall`


TeddyGavi:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Without first generating new types, it is possible that running `yarn nuxi typecheck` will show TS errors that do exist in the current project. 

In PR #203 I omitted an important note regarding the workflow for manually checking types with `yarn nuxi typecheck`

- `yarn nuxi typecheck` needs to be preceded by a command that first generates the types inside `.nuxt`, such as `nuxt prepare`, ( triggered by running `yarn run postinstall` as per `package.json`) or `yarn nuxi dev`

- The reason for this is listed in the [Nuxt docs](https://nuxt.com/docs/guide/concepts/typescript) 
> Some of the references in the (type) file are to files that are only generated within your buildDir (.nuxt) and therefore for full typings, you will need to run nuxi dev or nuxi build.

Updates are required in STYLEGUIDE and CONTRIBUTING. I suggest recommending `yarn run postinstall` as it is already included in the project scripts.

andrewtavis:
Thanks @TeddyGavi! Do you want to add this to the docs yourself, or should I/someone else get to it? :)

TeddyGavi:
I have a PR incoming soon! :smile:

andrewtavis:
Fantastic and much thanks, @TeddyGavi! Will review in the coming days :)

andrewtavis:
Looks to be all done now, @TeddyGavi! Glad to be making so much progress on documentation. Many new joiners have expressed how much the appreciate it all 😊 Thanks for all the efforts here!

PR #217, opened 2023-07-07
Update documentation to include 'yarn run postinstall' 

TeddyGavi:
### Contributor checklist

---

### Description

- Edit documentation to recommend running `yarn run postinstall` before manually checking types with `yarn nuxi typecheck` in STYLEGUIDE and CONTRIBUTING 

### Related issue

- #216


andrewtavis (approved):
Looks great, @TeddyGavi! I'll do a couple minor edits as I'm realizing we have `PR's are always welcome...`, which should be `PRs`. Will also switch the phrasing in the second step so it's all one clause :)

Issue #234, opened 2023-07-18
Mark `imageURL` and `altText` props in `LandingPageContent` as optional


TeddyGavi:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Continuing #233 and #211 

As noted by @luuu-xu. 

In `LandingPageContent.vue` mark both `imageURL` and `altText` props as optional.

luuu-xu:
@TeddyGavi Shall I make an PR regarding changing props for our `BtnLabeled` and `LandingPageContent` now? As we just discussed, although it seems like there are a lot of problems with TS now, but we can at least fix these two components first.

TeddyGavi:
@luuu-xu Sure thing. The `withDefault` is the better approach either way. It appears there are differences in the local configs between our two environments (at least), which needs some investigation. :+1:

andrewtavis:
Closing this issue as it was handled in #235, but will keep #233 open so we can continue the TS discussion :)

Issue #233, opened 2023-07-16
TypeScript Error in `BtnLabeled.vue`


TeddyGavi:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Error as follows:

> components/Btn/BtnLabeled.vue:45:5 - error TS2322: Type 'string | null' is not assignable to type 'string | undefined'.                                                                       
  Type 'null' is not assignable to type 'string | undefined'. 

![Screenshot from 2023-07-16 06-15-01](https://github.com/activist-org/activist/assets/102203363/a118cff6-243a-48d3-ac6f-71f340cf3e51)

A recommendation would be to make the string optional. However, from my understanding alt text is used by screen readers. :smile:  

@luuu-xu do you have thoughts on this?

Continues issue #211.

versacodes:
What is the value of alternateText? Maybe the error shows because of  alternateText: string | null in defineProps. Try string | undefined instead.

luuu-xu:
Yes, maybe it should be declared as `alternateText?: string`. 

The purpose of `BtnLabeled`'s `alternateText` prop is to add optional `aria-label` text string to our buttons where a more detailed explanation can be provided (with i18n) comparing to just having an ambiguous text inside of the button. 

By the way, @TeddyGavi  how did you set up your project that you have TS checkings running correctly in editor? I am having trouble with my project where all of these props yell at me so I couldn't declare my props right. See: https://github.com/activist-org/activist/issues/213#issuecomment-1634877949

Your help will be greatly greatly appreciated because I want them stop yelling at me. 😭

luuu-xu:
Right, so I have also found that in our landing page, the "Our supporters" section, which comes from `LandingPageContent  contentPosition="top"`, since it does not have a image on the side, ts says we need to pass in `imageURL` and `altText` props. But in fact we don't need to, so we should declare these two props as **optional** too.

andrewtavis:
This makes sense, @luuu-xu! As far as the errors - just checking - have you installed the VS Code extensions that are listed in the [.vscode/extensions.json](https://github.com/activist-org/activist/blob/main/.vscode/extensions.json) file? That is if you’re using VS Code :)

luuu-xu:
[...]

PR #235, opened 2023-07-18
Fix: TS prop errors for some components

luuu-xu:
### Contributor checklist

---

### Description

Fixes ts prop errors of `BtnLabeled.vue` and `LandingPageContent.vue`:

1. `BtnLabeled.vue`: 
   1. changed `alternateText` to be optional, its type to be string only.
   2. changed `iconSize` to be correctly typed with `withDefaults()` wrapper function so it is now optional and has a default value.

2. `LandingPageContent.vue`:
   1. changed `imageURL` and `altText` to be optional for the top position variant where there is no image.

### Related issue

- #234 
- #233

### DIFF
--- a/frontend/components/Btn/BtnLabeled.vue
+++ b/frontend/components/Btn/BtnLabeled.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     v-if="linkTo == 'placeholder-link'"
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Thanks for these changes, @luuu-xu, and also for the discussion @TeddyGavi 👏 We'll have these 33, 59 or 66 TS errors fixed in no time! 😊

Issue #228, opened 2023-07-15
Add `yarn nuxi typecheck` to `package.json`


TeddyGavi:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The goal would be to display the output of `yarn nuxi typecheck` in the Netlify deploy logs. 

The suggestion by @wkyoshida was to add the command to the `postinstall` script in `package.json`

This is an extension of issue #211 

### Contribution

I will add this command to line 9 and test it locally.

andrewtavis:
@TeddyGavi, another thing that @wkyoshida and I were discussing is that I could also add this to the Netlify build directly such that the errors would just be logged in there. As of now we have a build command that is `CI= npm run generate`, which could then be changed to `CI= yarn nuxi typecheck && npm run generate`.

We'd just need to make sure that this wouldn't break the build if there are TS errors. I could add this in and send along a PR after some of these PRs are merged 🙃

I think we can also include changing the [pr_maintainer_checklist.yaml](https://github.com/activist-org/activist/blob/main/.github/workflows/pr_maintainer_checklist.yaml) file in this issue as well :)

TeddyGavi:
I am thinking about this a bit more, should we be adding a config file for Netlify builds? If we are moving away from Netlify in the future perhaps not? 

After reading a bit more about `generate` it creates static files for static hosting. So it runs `build` afterwards. Reading the build logs, everything file I see is `*.js` from the `/dist` So I believe we would also need to include `nuxt prepare` [to generate the types first](https://nuxt.com/docs/api/commands/prepare), as I am not sure if they are autogenerated for static hosting or by `build`. I will need to try some options out, to make sure the build doesn't fail :worried:

TeddyGavi:
@andrewtavis circling back to this in regards to #242  
 > Hey @TeddyGavi! Thanks for the efforts here :) It seems like there’s no option for the type check tests to fail quietly in this way. Should I try a PR that edits the Netlify run?
 
 Or should we close this and create a new issue completely?

andrewtavis:
I’ll do this within this issue, @TeddyGavi! Thank you for reminding me! :) I’ll get to this early next week hopefully 😊

andrewtavis:
[...]

PR #242, opened 2023-07-26
feat: Extend `postinstall` script with `nuxi typecheck`, remove duplicate dependencies, ammend `pr_maintainer_checklist` 

TeddyGavi:
### Contributor checklist

---

### Description

Regarding a discussion at the latest dev sync, this PR will test if the `typecheck` log prints on the Netlify build logs. 

Chores:
- a checkbox is added to the `pr_maintainer_checklist` as a reminder to check the type safety of PR's 
- Duplicate dependency is removed

### Related issue

- #228


TeddyGavi:
@andrewtavis @wkyoshida  Interesting! 

I believe this caused by the error code thrown from running `yarn nuxi typecheck` if we change this command to be `npx vue-tsc --noEmit` no exit codes will be thrown, should I try out `npx vue-tsc --noEmit` or did I miss something?

andrewtavis:
I would say that it's worth a try, @TeddyGavi :) [vue-tsc](https://github.com/vuejs/language-tools/tree/master/packages/vue-tsc) definitely qualifies as something we could start using.

TeddyGavi:
@andrewtavis Will do once I'm back from holiday. If you want to go ahead and test it out and do a PR feel free!

TeddyGavi:
@andrewtavis It looks like adding `npx vue-tsc --noEmit` still causes the build the fail...not sure if I missed something here. Any suggestions or thoughts are welcomed! :thinking:

andrewtavis:
Hey @TeddyGavi! Thanks for the efforts here :) It seems like there’s no option for the type check tests to fail quietly in this way. Should I try a PR that edits the Netlify run?

TeddyGavi:
@andrewtavis I think that's a great next step, I'll close this PR now.

Issue #249, opened 2023-08-04
Navbar


to-sta:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The burger menu is not centered in the y-direction.
![image](https://github.com/activist-org/activist/assets/49518857/97b5d4df-89ba-4efd-9a48-d4d8d6534cee)

The active searchbar is colliding with the burger menu.
![image2](https://github.com/activist-org/activist/assets/49518857/e8e3caab-e06e-4854-952e-5e6df2fc9d81)

andrewtavis:
Hey @to-sta! Thanks for the issue! 🙏🚀

Big thing with these is we need to check them on multiple browsers as it does look good to me on chrome via shrinking the breakpoint to below `sm` :)

Do you have interest in working on this?

to-sta:
I would like to work on that.

I have also used chrome and checked on my phone as well. But it's a good point, I should check at least firefox and edge.

andrewtavis:
Sounds good, @to-sta! Assigned to you just now :) I'll make the issue for the footer right now 😊

andrewtavis:
Closed via #256 :) Keep the suggestions coming, @to-sta! 🚀

PR #256, opened 2023-08-11
Navbar

to-sta:
### Contributor checklist

---

### Description

This PR fixes a `Navbar` issue on small screens. On small screens if the `SearchBar` is expanded it could overlap the burger menu on the right side. Now the expanded `Searchbar` takes the remaining space next to burger menu.

This PR fixes the issues that arises on small screens if the searchbar is expanded, it can overlap the burger menu. The searchbar had a fixed width, now the searchbar is responsive and takes the remaining space next to the burger menu. The space between the burger menu and the expanded searchbar can be adjusted with the `gap-`

I have check the landing page and /home with different browsers:

- Mozilla Firefox
- Chrome 
- Edge

### Related issue

#249


frontend/components/SearchBar.vue
```diff
@@ -5,30 +5,18 @@
...
+        :class="{ 'focus:w-5/6': isInputFocused }" />
     </Transition>
     <Transition name="shortcuts">
-      <div
-        v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false"
-        ref="hotkeyIndicators"
-        class="absolute right-0 flex pr-6 space-x-1 transition-opacity transition-duration-200"
-      >
+      <div v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false" ref="hotkeyIndicators"
+        class="absolute right-0 flex pr-2 space-x-1 transition-opacity transition-duration-200">
```
andrewtavis, line 14: I think that this should be maintained at `pr-6`, but it's also mainly because of a change I sent through in the interim. I remember that we were talking about making this relative to the contianer, but then it makes it so that the tooltips get cut off at the edge of where the searchbar's margins intersect with the sidebar header.

If you have another suggestion I'd be happy to consider it!
```suggestion
        class="absolute right-0 flex pr-2 space-x-1 transition-opacity transition-duration-200">
```

andrewtavis, line 14: I think that this should be maintained at pr-6, but it's also mainly because of a change I sent through in the interim. I remember that we were talking about making this relative to the contianer, but then it makes it so that the tooltips get cut off at the edge of where the searchbar's margins intersect with the sidebar header.

If you have another suggestion I'd be happy to consider it!

andrewtavis: This is what the tooltip looks like for me locally with `relative` in the top div that I removed from here and pr-2:

Screenshot_2023-08-12 01 06 54

frontend/components/WebsiteHeader.vue
```diff
@@ -1,21 +1,14 @@
...
+  }">
     <!-- Note: mobile header -->
-    <div class="flex py-3 md:hidden">
-      <div class="z-0 flex justify-center mx-auto">
-        <div class="relative z-0 w-24 h-6">
-          <LogoActivist
-            :is-mobile="true"
-            class="absolute inset-0 flex items-center justify-center z-1 overflow-clip"
-          />
+    <div class="flex md:hidden py-3 px-4">
```
andrewtavis, line 8: Another minor thing I'm wondering about is why for the website header on mobile the focus for it is now the entire container rather than just `activist` 🤔

Screenshot_2023-08-12 01 12 58

Seems to be coming from here. Generally any edit that gets the focus back properly would be great 😊

andrewtavis:
These were the only major things that jumped out to me, @to-sta 😊 Thanks for the work on this! 🙌

to-sta:
@andrewtavis thanks for the review. I will revise my changes to include your findings.

andrewtavis:
Thanks, @to-sta! If you can find a way to maintain `relative` for the search bar then that would likely be best. Maybe playing around with `z-` a bit could bring the tooltip out :)

to-sta:
@andrewtavis fixed the focus on the `LogoActivist`. Figured out that the `NuxtLink` creates the anchar tag and the anchor tag inherits the size of the parent element. Added a div that solves that issues. 

Added the pr-6 back in again for the `SearchBar`. I could not reproduce the issue with the tooltip. Try it in firefox, chrome and edge. Always saw the tooltips. I am not sure how i can reproduce the issue you encountered.

andrewtavis:
@to-sta, sorry, I didn't do a good job of explaining a change I made in 1b03d73 that's preventing you from reproducing the issue. You'd need to add in `relative` to the top `<div>` for the search bar. This `relative` was the Tailwind class that was allowing us to place the `/` in relation to the search bar rather than in relation to the sidebar itself. The `pr-6` is applied as `pr-0` puts us at the sidebar's edge, whereas with `relative` `pr-0` goes to the edge of the search bar. With relative I was also getting this error with the `/` tooltip. I think that this works for now, and my suggestion of adding in a `z-INT` class won't actually work as that's already being applied by the following custom class:

https://github.com/activist-org/activist/blob/6086585fcac56b9a581acc78d8bcac84305c8d7e/frontend/assets/css/tailwind.css#L38

Thanks for the changes! Both things are working great 😊 Let me fix the merge conflicts and we'll be good to go 🚀

andrewtavis (approved):
Thanks for the quick changes here, @to-sta! Again I really appreciate the diagnosis and fixing this. Will be a lot happier looking at activist on my phone now 😊

Issue #259, opened 2023-08-12
Fix TS errors in components/card directory


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue is related to #211 in which we've been documenting the current TS errors that activist is facing. As of now the [components/card](https://github.com/activist-org/activist/tree/main/frontend/components/card) directory has the following error output when running `yarn nuxi typecheck` in the `frontend` directory:

```
Errors  Files
     2  components/card/CardAbout.vue:70
     1  components/card/CardConnect.vue:96
     1  components/card/CardFeedItem.vue:41
     2  components/card/CardGetInvolved.vue:64
     3  components/card/CardSearchResult.vue:183
```

Would be great if we could fix the TS errors in each of these files :)

andrewtavis:
CC @TeddyGavi and @luuu-xu 😊

TeddyGavi:
@andrewtavis @luuu-xu  I would like to take this one on :smile:

I have some thoughts I wanted to run by everyone. 

In  `components/card/CardAbout.vue` the different sections `organization or event` are rendered dependent on the `about-type` whereas in `components/card/CardGetInvolved.vue` the `event or organization` is rendered explicitly with a `v-if`. 

In `CardAbout` this means that without first checking that `organization` or `event` is not empty, you could possibly get undefined values. Whereas in `CardGetInvolved` they are only rendered if they have data. 

Which methodology is preferred? I haven't yet run the project to see if there is a visual need to render `event` and `organization` in `CardAbout` despite having no data. If this is the case it would be a good call to use `withDefault` with useful default descriptions for the user.

luuu-xu:
I took a look and think that there is definitely a need to render `CardAbout` even when `organization` or `event` is empty. So maybe it should render explicitly with `v-if` and some default placeholder text with a `v-else`? @andrewtavis What do you think though?

andrewtavis:
Hey @TeddyGavi! Thanks for picking this up! Great question :) I’d say that we should use the style that’s in `CardGetInvolved` as it’s one less prop to worry about. It’s all explicit enough as it is, I’d say. But getting this defined and documented in the style guide would also be great 😊 Maybe we can include that in this issue 😇

andrewtavis:
[...]

PR #280, opened 2023-08-15
fix: TS errors in multiple files in components/card.*

TeddyGavi:
### Contributor checklist

---

### Description

Points to review:
1. `tsconfig` included an alias that appeared to be overriding the defaults from nuxt, causing imported types to not be recognized as well as leading to types from inside `node_modules` not being auto-imported. https://nuxt.com/docs/api/configuration/nuxt-config#alias

2. Change `v-if` in `CardAbout` to use `event` and `organization` instead of `aboutType` following the same style as `CardGetInvolved`

3. Added typing for template ref in `PopupNewField` according to https://vuejs.org/guide/typescript/composition-api.html#typing-template-refs (emit is typed now)

4. In each search result component, the specific result ie `User, organization, resource, event` should not be optional

5. In `CardConnect` removed the arrow function from the emit of `@on-cta-clicked` (I don't believe this is needed, and it was causing account to register `no explicit any` ts error)

Current TS errors

image

### Related issue

- #259

### DIFF
--- a/frontend/components/card/CardAbout.vue
+++ b/frontend/components/card/CardAbout.vue
@@ -1,4 +1,4 @@
-<template>
+  <template>
   <div class="px-6 py-5 card-style">
     <div class="relative flex-col w-full gap-5">
       <div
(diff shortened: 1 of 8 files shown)


andrewtavis (commented):
Thanks for the fixes and all the extra work here, @TeddyGavi! All of this makes sense :) Thanks further for fixing the optional props for me for search results as I forgot to do that when I split the file 😊

Issue #272, opened 2023-08-13
Create IconOrganizationStatus component


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would create the component that indicates whether an organization is approved or pending approval and would serve as the link to its application page to see what organizations supported it coming on to activist. An example of it can be seen on [this page of the MVP designs](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2408%3A4794&mode=design&t=hJRakNbmbK83mZ9P-1), and its options can be seen below:

![Organization Status](https://github.com/activist-org/activist/assets/24387426/4b7f6039-9c0d-4d7d-8537-39e7be79ca04)

This component would include:

- A required `status` prop
- The styling would be based on the colors that are defined in [tailwind.config.ts](https://github.com/activist-org/activist/blob/main/frontend/tailwind.config.ts)
- The size would react to certain breakpoints

### Contribution

This is a great `good first issue`! I'd be happy to support someone who wants to work on this 😊

to-sta:
I would like give this issue a try. Is there any prefered icon package for the icons?

andrewtavis:
There's no preferred icon set per say, @to-sta, but we generally have been using [Bootstrap Icons](https://icones.js.org/collection/bi) for most things :) Thanks for your interest in working on this!

andrewtavis:
We're good to go now, @to-sta! Minor edits are in f49c3fd:

- As discussed removing the import
- For the Tailwind coloration we need to apply `dark:fill-FILL_COLOR` for the dark mode colors to be used
  - Didn't catch this in the review. My apologies!
- Made the space between the name and icon a bit bigger :)

Thanks again so much!

andrewtavis:
Updated the styleguide with some documentation for the color mode in e2e7950 :) :)

PR #281, opened 2023-08-16
IconOrganizationStatus component added

to-sta:
### Contributor checklist

---

### Description

In this PR the IconOrganizationStatus component was added. The Icon was made with [bootstrap icons](https://icones.js.org/collection/bi). 

The approved icon is a combination of `check-circle` and `check-circle-fill`.
The pending icon is a combination of `exclamation-triangle` and `exclamation-triangle-fill`.

Prop `status` takes a string `'approved' ` or else it's not approved by default.

To test the component i used the url `.../en/organizations/1/about`. I implemented the component in the about section.

### Related issue

- #272

### DIFF
--- a/frontend/pages/organizations/[id]/about.vue
+++ b/frontend/pages/organizations/[id]/about.vue
@@ -13,7 +13,7 @@
           {{ organization.name }}
           
         </h1>
-        <IconOrganizationStatus :status="'approved'"></IconOrganizationStatus>
+        <IconOrganizationStatus :status="''"></IconOrganizationStatus>
       </div>
     <flex class="relative flex items-center w-full py-6"
       ><h2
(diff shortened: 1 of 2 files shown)


frontend/pages/organizations/[id]/about.vue
```diff
@@ -67,6 +71,7 @@ definePageMeta({
 });
 
 import { Organization } from "~~/types/organization";
+import IconOrganizationStatus from "../../../components/icon/IconOrganizationStatus.vue";
```
Comment (andrewtavis) [inline frontend/pages/organizations/[id]/about.vue:74]: @to-sta, I think that this import here is unnecessary :) Nuxt covers all the component imports for us automatically, and we just need to load in the types 😊

Comment (to-sta) [reply frontend/pages/organizations/[id]/about.vue:74]: @andrewtavis thanks for the info. Will keep that in mind 👍

andrewtavis (approved):
Thanks for the changes here, @to-sta! I'll remove the component load as discussed and set it so that all the placeholder org pages are "approved" 😊

Issue #274, opened 2023-08-13
Create CardDangerZone component


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would create the `CardDangerZone` component that would be used by the user to delete certain elements on the page. Designs for it can be seen on the [MVP page of Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2475%3A10600&mode=design&t=hJRakNbmbK83mZ9P-1). This component would include:

- A required `description` prop
- A required `ctaBtnText` prop
- Text entry for user names and passwords
- Styling using the colors found in [tailwind.config.ts](https://github.com/activist-org/activist/blob/main/frontend/tailwind.config.ts)

An example of the component can be seen here:

![Personal Information](https://github.com/activist-org/activist/assets/24387426/151995ee-33fb-4e47-bf66-935d5689e8da)

### Contribution

This is a great `good first issue` that I'd be happy to support someone in working on! Also happy to work on it myself if others don't have interest :)

to-sta:
I would like to work on this issue.

andrewtavis:
Thanks, @to-sta! Really great to get this done so we can do the settings pages 😊

andrewtavis:
4eb4bd6 does some minor changes to finish this up, @to-sta! Thanks again! 🙏😊 Coloration changes were more based on getting a distinct color for light and dark mode, although all this will likely shift a bit given #215 and the coloration shift that will happen when I get to that :)

Issue #270, opened 2023-08-13
Nuxt layout for a page with a header and sidebar


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

A very common layout that is needed is the one that's displaying content on organization and event pages. As can be seen in the [MVP designs](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=766%3A214&mode=design&t=pctNmHd3eWcrj6hc-1), this layout consists of a sidebar that would be inherited from the [sidebar layout](https://github.com/activist-org/activist/blob/main/frontend/layouts/sidebar.vue) and would further include a page header, a page tagline, a slot for CTA buttons on the top of the page and then page content would fall below the CTA buttons and tagline. Having one layout for this would allow us to consolidate breakpoint styling for many pages and then have individual components control how they appear at certain screen widths.

### Contribution

This is a major priority now to make sure that we are not repeating our page logic too much. I'll be working on this, and would appreciate ideas of how best to implement this :)

luuu-xu:
I agree this is the major priority now. It seems like it's easy to implement since the layout is only consisted of header, tagline and main content. I imagine Event/Organization Settings page will also use this layout too? 

Make sure you assign yourself to this issue! 😁

andrewtavis:
Thanks for the heads up, @luuu-xu! :)

andrewtavis:
Thinking about this atm and looking into some documentation, I don't think it's advisable to make a layout in this way. There are no props for layouts, so we wouldn't be able to pass the needed text information to the layout and still have them formatted appropriately. If we want to make sure that all headers are correctly formatted and can just have strings passed, then the best way to do it would be to create a component :)

Will look into this now 😊

andrewtavis:
1a9bcdf adds in [HeaderAppPage.vue](https://github.com/activist-org/activist/blob/main/frontend/components/header/HeaderAppPage.vue) that styles the header and tagline of a page. There's also a slot for right justified content that's inline with the tagline in which buttons and drop downs can be placed 😊 Using this will make the designs for future app pages much easier 🚀

PR #282, opened 2023-08-17
CardDangerZone implementation

to-sta:
### Contributor checklist

---

### Description

This PR contains the CardDangerZone component. I tested the url `/en/organizations/1/about` and used the chrome devtools to check light/dark mode. 

Comments:
I used the styles from `card-style` for the first `<div>` element without the colors to keep the card consistent with the rest. But it will not be updated in case the `card-style` changes. I think a `base-card-style` without colors could help.

### Related issue

- #274


andrewtavis:
Great idea on the `base-card-style`, @to-sta! I'll implement that after the review :)

andrewtavis (approved):
Thanks for this, @to-sta! Really is working well, and looks great 😊 We'll need to get this onto the `[id]/settings` page, but I can do that right now with the change to add in the `card-style-base` Tailwind class :) :)

andrewtavis:
We need to get #270 done before adding it to settings makes much sense anyway. Let's keep it in mind and add it back in there when that issues's done 😊

Issue #263, opened 2023-08-12
Website footer as flex at large breakpoint when there's no sidebar


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

As of now the way that the [footer component group](https://github.com/activist-org/activist/tree/main/frontend/components/footer) works is that `FooterFlex` is only applied after the `xl` breakpoint so that it's not restricted when there's a sidebar. It would be great if we could check to see if the sidebar layout is being used, and if not such as for the landing page we could apply `FooterFlex` after the lg breakpoint. Reasoning for this is that `FooterFlex` is more of a desktop footer and thus fits better if it won't be restricted :)

andrewtavis:
The screenshot below illustrates this a bit better. This is the footer at the large breakpoint on the desktop, which is currently `FooterFlexCol` and has tons of space to the left and right. It'd be great for it to be `FooterFlex` while still being like this when there's a sidebar.

![Screenshot_2023-08-13 01 58 55](https://github.com/activist-org/activist/assets/24387426/40803f79-923b-446d-b81d-82868df2c2e9)

Vidhanvyrs:
hey @andrewtavis can i work on this

andrewtavis:
Fantastic, @Vidhanvyrs! Thanks for picking this up, and please let me know if you have any questions :) :)

Vidhanvyrs:
thanx for assigning me this issue
If I understand correctly, you want to adjust the behavior of the FooterFlex component based on whether a sidebar layout is being used and also the screen size breakpoints. You also want the FooterFlex to be applied at the lg breakpoint for the landing page when there's no sidebar right

andrewtavis:
It's more just when `FooterFlex` is applied, @Vidhanvyrs. Something that could happen here is we rearrange [Footer.vue](https://github.com/activist-org/activist/blob/main/frontend/components/footer/Footer.vue) and its sub-components such that the Tailwind based `hidden` and `flex` classes to hide and show the footers are contained within `Footer`. Then from there we can add a boolean condition in for whether the sidebar is present ot not at any given point and then change the class based on this. What I mean would be something along the lines of:

```html
<div class="..." :class="{ 
  'hidden xl:flex': sidebarPresent, 
[...]

Issue #269, opened 2023-08-13
Cover screen with transparent view within lg/md breakpoint when sidebar is open


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The way we want the activist sidebar to function is that within the large and medium breakpoints opening the sidebar will cause a semi-transparent view to cover the screen's content. This change would be made within the [sidebar layout](https://github.com/activist-org/activist/blob/main/frontend/layouts/sidebar.vue). If the user clicks to keep the sidebar open, then clicking on the semi-transparent view would close the sidebar, and if the user has opened the sidebar by hovering over it, then moving the mouse away would cause the semi-transparent view to disappear.

### Contribution

Happy to work on this or support someone who would like to work on it! This feature can be tested on the [home page](https://activist.org/home/).

andrewtavis:
Closed by 3ef2f7c 😊

PR #287, opened 2023-08-22
condition_classes_applied

Vidhanvyrs:
### Contributor checklist

---

### Related issue

- #263

### DIFF
--- a/frontend/components/footer/Footer.vue
+++ b/frontend/components/footer/Footer.vue
@@ -1,6 +1,11 @@
 <template>
   <footer
-    class="responsive-px-5 responsive-py-5 bg-light-header dark:bg-dark-header text-light-special-text dark:text-dark-special-text"
+    :class="{
+      'responsive-px-5 responsive-py-5 bg-light-header dark:bg-dark-header text-light-special-text dark:text-dark-special-text':
+        !hasSidebar,
+      'responsive-px-5 responsive-py-5 bg-light-header dark:bg-dark-header text-light-special-text dark:text-dark-special-text with-sidebar':
+        hasSidebar && isSidebarOpen,
+    }"
   >
     <!-- Note: Content Sections Top for Mobile -->
     <FooterFlexCol class="flex flex-col xl:hidden" :links="links" />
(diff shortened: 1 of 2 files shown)


Vidhanvyrs:
hey @andrewtavis Please give the feedback for the applied changes

andrewtavis:
Thanks for the PR, @Vidhanvyrs! I'm seeing that for the landing page in the deployment preview we're still getting the `flex-col` version at the large breakpoint. Specifically the conditions that we want to set for `FooterFlex` and `FooterFlexCol` should be along the lines of:

```
<!-- Note: Content Sections Top for Mobile -->
<FooterFlexCol class="flex flex-col" :class="{'xl:hidden': hasSidebar, 'lg:hidden': !hasSidebar}" :links="links" />

<!-- Note: Content Sections Left and Right for Desktop (xl) -->
<FooterFlex class="hidden" :class="{'xl:flex': hasSidebar, 'lg:flex': !hasSidebar}"  :links="links" />
```

The sidebar being open shouldn't change anything because at the `lg` and `md` breakpoints we'll just cover the page content and put an overlay over it after we finish #269. We can look into changing this behavior later though :)

I'd say generally the way to go about this would be to check the layout being applied. If the layout is `sidebar` then we want `hasSidebar === true`. Happy to look into how best to check which layout is being applied in Nuxt! And please let me know if the above needs a bit more explaining :) :)

Vidhanvyrs:
yeah @andrewtavis thankyou for correcting it🙇‍♂️ i have done the asked changes and i think it should work and about this 
" I'd say generally the way to go about this would be to check the layout being applied. If the layout is sidebar then we want hasSidebar === true "
I think i already added the ref to it so this probably will work 
and yes let the  Cover screen with overlay issue get done then I can look into  that 🙇‍♂️😄

andrewtavis (approved):
The footer is definitely working the way it's supposed to @Vidhanvyrs. One thing is that now the sidebar's gone on pages that need it, but I'll play around a bit and get it back post merge :)

Thank you for working on this! I'll ping you in the issue once it's all finished up :) :)

PR #289, opened 2023-08-24
#228 check to see build with yarn nuxi typecheck

andrewtavis:
### Contributor checklist

---

### Description

Note that this PR is not meant to be merged. Contents are the removal of the community banner of the landing page, but the purpose is to see the build output for Netlify when we include `yarn nuxi typecheck` :)

### Related issue

- #228

### DIFF
--- a/frontend/pages/index.vue
+++ b/frontend/pages/index.vue
@@ -4,7 +4,6 @@
       <Title>{{ title }}</Title>
     </Head>
     <LandingSplash />
-    <LandingCommunityBanner />
     <div
       class="flex flex-col mt-4 mb-2 sm:mb-4 sm:mt-10 xl:mt-16 xl:mb-8 2xl:mt-24 2xl:mb-12"
     >


andrewtavis:
@TeddyGavi, @wkyoshida, @momanyisamuel, @luuu-xu:

Finally got around to checking `yarn nuxi typecheck` within Netlify build. We now have `CI= yarn nuxi typecheck && npm run generate`.

As I was kind of expecting, it does cause the build to fail. With that being said, we do have a really nice output for all the TS errors in the deploy log 🤔 We were saying in the dev sync before last that we don't want introduced TS errors to cause the build to fail and then have someone scrambling to figure out why. I feel like with a bit of guidance from the maintainer checklist this could work though? They see that there are failed checks, look at the deploy preview that has a giant red `X` next to it, and then they can see all the TS errors to cause it to fail.

Thoughts on this? If we were to do this, is there a preference between this implementation that fails on the build vs. #242 from @TeddyGavi that fails on initialization.

andrewtavis:
Also see #290 that has the same :) Seems like the `--noEmit` isn't able to prevent a return of `1` in this case as well... Any other ideas of how we can get the errors without failing?

TeddyGavi:
@andrewtavis I was hoping for this not to be the case! Well, the thing that comes to my mind is returning to my original idea of creating a GitHub action that will run and log the output from type-checking. Bypassing the whole "build fail due to TS error" issue. A maintainer could also check the TS log as part of the checklist.

My general idea here is to keep the type-checking automated and separate from the build to prevent deployment failures.

Could work? Thoughts?

andrewtavis:
Exploring a GitHub action seems to be the logical next step for me, @TeddyGavi! Let’s make sure that we make it depend on the maintainer checklist just to be sure that that information is conveyed to them first :)

andrewtavis:
I'm going to close this, @TeddyGavi. Let's pick up the discussion in #228 :) First things first is a bit of research to get the action running on the branch, and from there this should be pretty clear cut.

Issue #262, opened 2023-08-12
Fix TS errors in misc components


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue is related to #211 in which we've been documenting the current TS errors that activist is facing. As of now the [components](https://github.com/activist-org/activist/tree/main/frontend/components/form) directory has a few miscellaneous `.vue` files that have the following error output when running `yarn nuxi typecheck` in the `frontend` directory:

```
Errors  Files
     1  components/landing/LandingContent.vue:162
     4  components/modal/ModalImage.vue:9
     9  components/PageBreadcrumbs.vue:53
     2  components/selector/SelectorLanguage.vue:76
     4  components/sidebar/left/SidebarLeftFooter.vue:17
```

Would be great if we could fix the TS errors in each of these files :)

andrewtavis:
CC @TeddyGavi and @luuu-xu 😇  Too many CCs, but I figure you two would like to follow these and maybe help coordinate the work if someone wants to work on some of these :)

luuu-xu:
Let me handle this TS issue since it has breadcrumbs which I consider as my baby now, haha 😛

andrewtavis:
Thank you, @luuu-xu! Looking forward to getting all these TS finalized :) :)

andrewtavis:
140e62d finalizes this, @luuu-xu :) :) Added back the name of the local, and I also fixed the issue where the local name would be broken onto two lines when it's in the header.

Thanks for the work here! Hope to see you in the sync, or if not as I wrote on Matrix let's set up a call!

PR #292, opened 2023-08-25
Fix: fixes TS errors in misc components

luuu-xu:
### Contributor checklist

---

### Description

Fixes TS errors for several misc components listed as below:

1. `PageBreadcrumbs`
2. `ModalImage`
3. `LandingContent`
4. `SidebarLeft`
5. `SidebarLeftFooter`
6. `SelectorLanguage`

Although now TS errors are all gone, I am not happy about the work-arounds relating i18n's `locales` interface. For example, in `SelectorLanguage`, the original error was about `locales.value.filter` because locales has an union of `string[] | LocaleObject[]`. 

I researched for some time and found only a walkaround as in my PR, and the result is that now the `availableLocales` has a type of `Array<string | LocaleObject>`, so another function `getLocaleCode` is added. 

I don't think this is an elegant solution so maybe you can take a look at it?

### Related issue

- #262

### DIFF
--- a/frontend/components/selector/SelectorLanguage.vue
+++ b/frontend/components/selector/SelectorLanguage.vue
@@ -42,8 +42,8 @@
         <ul class="px-2 py-2">
           <NuxtLink
             v-for="locale in availableLocales"
-            :key="locale.code"
-            :to="switchLocalePath(locale.code)"
+            :key="getLocaleCode(locale)"
+            :to="switchLocalePath(getLocaleCode(locale))"
           >
             <MenuItem v-slot="{ active }">
               <li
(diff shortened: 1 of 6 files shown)


andrewtavis:
Thank you for this, @luuu-xu! I’ll try to get to this review before the dev sync tomorrow :) :)

frontend/components/selector/SelectorLanguage.vue
```diff
@@ -54,7 +54,7 @@
                   'group flex w-full [word-spacing:0.5em] items-center rounded-md pl-4 pr-2 py-2 text-sm',
                 ]"
               >
-                {{ locale.name }}
+                {{ getLocaleCode(locale) }}
```
andrewtavis, line 57: This here needs to be `.name`. I can make a similar function for it just so that we're using the same methods of accessing the local info.

andrewtavis (approved):
I'd say the current workaround is fine for now, @luuu-xu. We have lots to do, and if this works for a component that we hopefully won't be touching too much, then let's not stress too much about it :) You'd also be welcome to create a good first issue for this and see if someone eventually wants to pick it up where you and I could then review 😊

Issue #288, opened 2023-08-23
Add needed strings to localizable JSON file


andrewtavis:
### Terms

- [X] I have searched [open localization issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+is%3Aopen+label%3Alocalization)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to go through and add in the localizable strings that are currently not in [i18n/en-US.json](https://github.com/activist-org/activist/blob/main/frontend/i18n/en-US.json). All strings that are on the platform should be within this file such that they can be localized.

felipolis:
Let me see if I understood. Should I look for strings within the project that are not in the i18n/en-US.json file and add these strings to this file? I'm interested in working on this issue

andrewtavis:
Yes, @felipolis, that’s exactly what we’re looking for here. Feel free to send along whatever keys you think are best in a PR and I’ll fix them if they need to be 😊 Thanks for your willingness to help!

felipolis:
Working on it. Thank you!!!

andrewtavis:
Thank you, @felipolis!

felipolis:
@andrewtavis I've just made a pull request. I would be grateful if you could take a look. I did my best :D

andrewtavis:
@felipolis, let’s definitely remember to do an edit to the [contributing guidelines](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md) to account for the conventions we set out here :) I’d be happy to do this myself, but let’s just make sure it gets done as this will be important for others to understand 😊

andrewtavis:
716c0a5 finishes this off, @felipolis 😊 Thanks for all the work here! Please let me know what you think of the [localization styleguide entry](https://github.com/activist-org/activist/blob/main/STYLEGUIDE.md#localization)!

Feel free to open another issue if you find some things, and we can used this closed issue to discuss the style guide if need be :) :)

andrewtavis:
Thinking about the system one more time, @felipolis, I'm thinking some further changes might make the text keys a bit more understandable:

- Instead of `"section-#-paragraph-#-pt-#"` we can do `"section-#-paragraph-#-#"`
- Instead of `"section-#-list-#-item-#-subitem-#"` we can do `"section-#-list-#-item-#-#"`

[...]

PR #300, opened 2023-08-26
Restructuring the i18n module in the entire project

felipolis:
### Contributor checklist

---

### Description

I was reviewing the project and noticed that the en-US.json file is quite disorganized, containing many jumbled labels without following a specific pattern. Given my familiarity with i18n in my work environment, I chose to apply the configuration standard that I use daily. This standard is based on the directory hierarchy structure. For example, if there is a title within the HTML in the file /pages/supporters/join.vue, in the i18n file, it would be referenced as "pages.supporters.join.title". I extended this pattern throughout the entire frontend of the project.

Furthermore, I replaced ALL standalone strings found in pages and components with i18n calls, which will facilitate translation into other languages. Although it took a significant amount of time, I believe this update will greatly enhance the project's internationalization.

I suggest that contributors be guided to adhere to the i18n standard rules when creating or editing components, in order to maintain code organization.

### Related issue

- #288


andrewtavis:
This is really great stuff, @felipolis! Will obviously take a bit to review, but I can say now that I’m very happy that you’re the one who picked up #288 😊

felipolis:
I'm Happy to help :D

andrewtavis:
Hope to see you in the dev sync in an hour, @felipolis! No stress if you can't make it this time :)

felipolis:
Sure!

frontend/components/grid/GridGithubShields.vue
```diff
@@ -7,7 +7,7 @@
     >
       <Icon name="cib:github" size="2em" />
       <div class="pb-1 ml-4 place-self-center">
-        <div class="font-semibold text-xs leading-4 ml-0.5">Visit us on</div>
+        <div class="font-semibold text-xs leading-4 ml-0.5">{{ $t("components.grid.grid-github-shields.visit") }}</div>
```
andrewtavis, line 10: Noting here that this could also be applied for `Fork` and `Star` labels below. I can also get to this though :)

frontend/components/landing/LandingSplash.vue
```diff
@@ -6,14 +6,14 @@
       <h1
         class="w-10/12 text-3xl font-bold leading-snug text-center lg:font-bold sm:text-4xl md:text-5xl xl:text-6xl md:max-w-lg xl:max-w-2xl sm:leading-snug md:leading-snug xl:leading-snug font-display"
       >
-        {{ $t("landing-header") }}
+        {{ $t("components.landing.splash.header") }}
```
andrewtavis, line 9: Note that given naming conventions this should be `components.landing.landing-splash.header` :) Similarly below.

frontend/components/landing/LandingTechBanner.vue
```diff
@@ -5,17 +5,17 @@
     <h2
       class="w-10/12 text-2xl font-semibold text-center sm:text-3xl sm:w-full xl:text-4xl"
     >
-      {{ $t("landing-page-github-header") }}
+      {{ $t("components.landing.tech-banner.github-header") }}
```
andrewtavis, line 8: Similarly here we'd have `components.landing.landing-tech-banner.github-header` :)

frontend/components/sidebar/left/SidebarLeftFooter.vue
```diff
@@ -55,28 +55,28 @@
             >
               <SidebarLeftSelector
                 iconURL="bi:calendar-check"
-                btnText="New event"
+                :btnText="$t('components.sidebar.left.footer.new-event')"
```
andrewtavis, line 58: Should be `components.sidebar.sidebar-left-footer.new-event` and similar changes are needed here :)

frontend/components/sidebar/left/SidebarLeftIndex.vue
```diff
@@ -95,63 +95,63 @@ interface SidebarLeftSelectorType {
 
 const organizationButtons: SidebarLeftSelectorType[] = [
   {
-    btnText: "sidebar.about",
+    btnText: "components.sidebar.left.index.about",
```
andrewtavis, line 98: Maybe it does make sense to not include the full directory structure for all of this, as in this case we'd have `sidebar` and `left` in there multiple times. I'll think on this, but also for non-tech localizers it might make sense to make it as simple as possible.

frontend/components/WebsiteHeader.vue
```diff
@@ -33,8 +33,8 @@
           <div class="flex items-center space-x-3 lg:space-x-4 xl:space-x-6">
             <SelectorTheme />
             <SelectorLanguage />
-            <BtnLabeled :cta="true" label="get-in-touch" linkTo="/help/contact" fontSize="sm" class="hidden lg:block" />
-            <BtnLabeled :cta="true" label="get-in-touch" linkTo="/help/contact" fontSize="xs"
+            <BtnLabeled :cta="true" label="components.website-header.get-in-touch" linkTo="/help/contact" fontSize="sm" class="hidden lg:block" />
```
andrewtavis, line 36: Would it make sense for something like this that's generally used on many pages to maybe shift it to `components.btn-labeled.label.get-in-touch`? So maybe if we have a component that's being used within another, the localized strings for it as "options" would be in that part of the JSON?

frontend/pages/about/activism-network.vue
```diff
@@ -3,12 +3,11 @@
     class="text-light-text dark:text-dark-text bg-light-content dark:bg-dark-content"
   >
     <Head>
-      <Title>{{ title }}</Title>
+      <Title>{{ $t("pages.about.activism-network.title") }}</Title>
     </Head>
-    <h1 class="p-4">This will be a page about our partner organizations</h1>
+    <h1 class="p-4">{{ $t("pages.about.activism-network.description") }}</h1>
```
andrewtavis, line 8: I'll likely remove this one and other placeholder ones so we don't localize things that ultimately are going to be removed :)

frontend/components/grid/GridGithubShields.vue
felipolis: You're right.. I didn't realize lol

frontend/components/landing/LandingSplash.vue
felipolis: yes, I'll fix that. It's just that I had thought that the "landing" would be very repetitive, but it's better to leave it complete anyway

frontend/components/landing/LandingTechBanner.vue
felipolis: I'll fix that

frontend/components/sidebar/left/SidebarLeftFooter.vue
felipolis: I'll fix that

frontend/components/sidebar/left/SidebarLeftIndex.vue
felipolis: Its true.. What do you think should I put there for now?

frontend/pages/about/activism-network.vue
felipolis: Do you think should I remove these labels for these pages and left them static for now?

frontend/components/WebsiteHeader.vue
felipolis: The issue is that btnLabeled component can receive different labels. In this case, the text is from the WebsiteHeader. However, if another component instantiated the BtnLabeled, the path would be that component.

andrewtavis: Awake and checking this again 😊 But for things like `BtnLabeled` couldn't we just switch the last part of the key? So we could have:

- `components.btn-labeled.label.get-in-touch`
- `components.btn-labeled.label.learn-more`

And anywhere that that same text it used we could then apply it? As of now we'd need to potentially apply the same label that's technically on the same component, but it's just on different pages 🤔

frontend/pages/about/activism-network.vue
andrewtavis: I'd say it'd likely be best, but I can also get this if need be 😊

frontend/components/sidebar/left/SidebarLeftIndex.vue
andrewtavis: I'd say it'd be easier if we leave out the full directory structure. Components are named in the way that they are such that the name reflects the directory in which they're found :) If I were to see the following two examples I think I'd know they're the same, so let's keep it easy on ourselves and pick the short:

- `components.sidebar.left.sidebar-left-index.about`
- `components.sidebar-left-index.about`

If a person wanted to find it in the code, they should be able to do so easily with both versions. Consistency would be key though 😊

andrewtavis: Especially with `components` in the key, we'd definitely be able to know where it is :)

frontend/components/WebsiteHeader.vue
andrewtavis: What I'm thinking is that the keys would be defined in relation to the lowest component on which they're found, or specifically in relation to the component in which they're applied. Yes we're currently within `WebsiteHeader` here, but the component is technically `BtnLabeled`.

frontend/pages/docs/index.vue
```diff
@@ -3,12 +3,11 @@
     class="text-light-text dark:text-dark-text bg-light-content dark:bg-dark-content"
   >
     <Head>
-      <Title>{{ title }}</Title>
+      <Title>{{ $t("pages.docs.index.title") }}</Title>
```
andrewtavis, line 6: Btw, I really like how you also went through and are doing the localization on a deep level like for the page header and such. This is so great :) :)

andrewtavis (changes requested):
Overall this is really great, @felipolis 👏😊 Something to realize is that we don't need to stress too much about getting this perfect, as there will be changes. What I'd argue for is the following:

- We define things on the lowest component level as discussed in [this comment](https://github.com/activist-org/activist/pull/300#discussion_r1306622037)
  - I think that for development it also will be easier as if someone's making a new key they don't have the minor cognitive load of "Ok, where am I in the website right now? Check file name..." vs. "This key is on this component that I'm working on".
  - Maybe more something for me than others at this point 😇
- We name keys in a way that repeats from the directory structure are not repeated in the name of the key as discussed in [this comment](https://github.com/activist-org/activist/pull/300#discussion_r1306623283)
- Would also make sense to revert the placeholder texts back to static so that if we do start active localization before some of their pages are worked on, that someone isn't then localizing something that'll never be live (as discussed [here](https://github.com/activist-org/activist/pull/300#discussion_r1306573070))

andrewtavis:
I'm very happy to get to any of the above points, @felipolis! Would be great to get this merged today as there will be quite some conflicts for anything that comes after 😊

felipolis:
Thanks so much for the feedback, @andrewtavis :D I tried to use this morning to correct the notes you said. Could you take a look?

andrewtavis (approved):
Looks all good to me, @felipolis! I'm going to go through and do a sweep of the files here and make some changes, and then will write to you in #288 when I'm all done 😊

felipolis:
Sounds good! Thanks!!!

PR #301, opened 2023-08-26
improved backend/apps/models.py

kurtisolo:
Improved backend/apps/models.py

### Contributor checklist

---

### Description

- Every field without a default value has been temporarily set to null=True.
- Fields "creation_date" and "last_updated" have been standardized so that Django manages them automatically:
    creation_date = models.DateTimeField(auto_now_add=True)
    last_updated = models.DateTimeField(auto_now=True)
- I added a brief comment as a reminder in models.py

### Related issue

- #174

### DIFF
--- a/backend/apps/models.py
+++ b/backend/apps/models.py
@@ -2,14 +2,25 @@ from django.contrib.auth.models import AbstractUser
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
 
+'''
+Considerations:
+
+- Every field without a default value has been temporarily set to null=True.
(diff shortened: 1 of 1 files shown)


andrewtavis:
Thanks for this, @kurtisolo! Will check it later today :)

andrewtavis (approved):
Thanks for the improvements here, @kurtisolo! Appreciate the comment for documentation as well, and am looking froward to us working on this a bit more and figuring out all the issues such that it can be removed and replaced with all the other needed comments 😊

PR #304, opened 2023-08-26
#228 test to see if the typecheck will work

andrewtavis:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

Note: this is not meant to be merged and is just a check to see if the typecheck can work.

### Related issue

- #228

PR #309, opened 2023-08-26
#228 test to see if the typecheck will work

andrewtavis:
### Contributor checklist

---

### Description

Note that this is a test to see if the typecheck works.

### Related issue

- #228

### DIFF
--- a/frontend/pages/index.vue
+++ b/frontend/pages/index.vue
@@ -4,7 +4,6 @@
       <Title>{{ title }}</Title>
     </Head>
     <LandingSplash />
-    <LandingCommunityBanner />
     <div
       class="flex flex-col mt-4 mb-2 sm:mb-4 sm:mt-10 xl:mt-16 xl:mb-8 2xl:mt-24 2xl:mb-12"
     >


andrewtavis:
Heyooo @TeddyGavi! 😊 So this has been a bit of a mess of trying things in the workflow and then opening new PRs, but I think this might work. What we were going for here was to be able to check the TypeScript errors while still being able to see a deploy preview from Netlify, and that's exactly what we have here 🥳 See the [workflow file](https://github.com/activist-org/activist/blob/main/.github/workflows/pr_ts_typecheck.yaml) for details.

Thing is that the deploy preview needed to be triggered manually, but maybe it's just because Netlify was sick and tired of getting requests from this branch as #304 did originally get a preview and none of the ones after did :)

Let's see how this goes when the next PR comes through and then if all goes well we can close #228 🚀

andrewtavis:
Closing this and we'll wait on another PR :)

TeddyGavi:
@andrewtavis Yes! This is exactly what I imagined. I did not fully understand how to implement it and I will study what you have done. Thank you for this and keeping me in the loop. 👍

andrewtavis:
Happy to keep you in the loop, @TeddyGavi! 😊

Issue #293, opened 2023-08-25
 Enforcing Standard Structure in Django Models


momanyisamuel:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Maintaining a consistent and well-structured codebase is crucial for the maintainability and scalability of activist. In activist-org's Django backend, it's essential to ensure that the `models.py` files adhere to a standardized structure to enhance code readability and streamline future development efforts.

**Proposed Solution:**
To address this issue, I propose implementing a standard structure for the `models.py` files across activist-org's Django backend. By adhering to a consistent structure, we can minimize confusion, make it easier for developers to navigate the codebase, and promote best practices.

**Standard Structure Guidelines:**
Below is an example of the proposed standard structure for `models.py` files. This structure is meant to provide a clear and organized layout for defining Django models and is based on the following references

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Models

```

from django.db import models

class MyModel(models.Model):
    # Fields
    #Use UUID for id fields
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False) 
    name = models.CharField(max_length=255, unique=True)
    description = models.TextField()

    # Timestamps
    created_at = models.DateTimeField(auto_now_add=True)
    updated_at = models.DateTimeField(auto_now=True)

    #Metadata
    class Meta:
        verbose_name = "My Model"
        verbose_name_plural = "My Models"

    #Methods
    def __str__(self):
        return self.name

```
The task is to review and update existing `models.py` file  and ensure that it is in line to the proposed standard structure - other structure improvement suggestions are welcome. This may involve reorganizing code and adding missing fields or methods.

LATER - There is also need to create a style-guide to document the standards which will be a good reference for other developers

### Contribution

Happy to support someone on this issue or work on it myself

to-sta:
I would like to work on this issue :)

andrewtavis:
Finally a backend issue for you, @to-sta 😉 Assigning now :)

to-sta:
I have question regarding this issue. Some models have a field `index` like below. 

```python
class OrganizationTopic(models.Model):
    index: models.IntegerField(null=True)
[...]

PR #315, opened 2023-08-28
Models standard structure

to-sta:
### Contributor checklist

---

### Description

In this PR the standard model structure was implemented to all models inside `backend/apps/models.py`. 

The `index` for each model was changed to `id` and set as primary key.
```python
# previously
index = models.IntegerField(null=True)
# now
id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False) 
```
Timestamps have been grouped and seperated by a blank line on each model. 

Example: 
```python
class Role(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    is_custom = models.BooleanField(null=True)
    description = models.TextField(max_length=500)

    creation_date = models.DateTimeField(auto_now_add=True)
    last_updated = models.DateTimeField(auto_now=True)
    deprecation_date = models.DateField(null=True)

    def __str__(self):
        return self.name
```
`def __str__:` have been added to each model.

Some fields were assigned `:` instead of `=`. Now all fields are assigned correctly and checked by making migrations. 

Example:
```python
class TopicFormat(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    index: models.IntegerField(null=True)
    topic_id: models.ForeignKey(Topic, on_delete=models.CASCADE)
    format_id: models.ForeignKey(Format, on_delete=models.CASCADE)

    def __str__(self):
        return self.index
```
I deleted the `class Meta:` for all models since the default plural made sense to me.

```python
class Meta:
    verbose_name = "organization"
    verbose_name_plural = "2. Organizations"
```
Django will add a `s` to the end by default. For cases like `Child` it would be by default `Childs` instead of `Children`. But the models we have are all correct by default (to the best of my knowledge) like `Event` -> `Events`. This can ge reverted tho in no time :).

<ins>Comments:<ins>
I would suggest to deleted the initial migrations file `migrations/0001_initial.py`. Since we db is not in production yet.

### Related issue

- #293

### DIFF
--- a/backend/apps/models.py
+++ b/backend/apps/models.py
@@ -3,7 +3,7 @@ from django.contrib.auth.models import AbstractUser
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
 
-'''
+"""
 Considerations:
 
 - Every field without a default value has been temporarily set to null=True.
@@ -12,11 +12,11 @@ Considerations:
     last_updated = models.DateTimeField(auto_now=True)
 - All fields have on_delete=models.CASCADE: this needs to be reviewed, as SET_NULL is preferable in many cases.
 - More comments should be added to improve the readability and understanding of the code.
-'''
+"""
 
 
 class Support(models.Model):
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False) 
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     index = models.IntegerField(null=True)
     supporter_type = models.IntegerField(null=True)
     supporter_entity = models.IntegerField(null=True)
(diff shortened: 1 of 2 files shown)


andrewtavis:
Thanks for the PR, @to-sta! No stress on the failed TS test, but I think you knew that 😊 Will give it a look in the coming days.

andrewtavis (approved):
Clean and lots of great fixes/additions here, @to-sta! I'm going to do a pass while I have Figma open to make sure that the columns are all there and in the right order, but I agree that `creation_date` fields should go to the end. Thanks so much!

Issue #316, opened 2023-08-28
Black formatting check


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In the `styleguide` [black](https://github.com/activist-org/activist/blob/main/STYLEGUIDE.md?plain=1#L123) is mention for formatting python code. Would it make sense to add a github workflow for it to ensure its enforced. 

### Contribution

I could help with the implementation.

andrewtavis:
Sounds interesting, @to-sta! Curious how this will work in a workflow :) Assigning you now 😊

andrewtavis:
91af592 sends the minor edits I mentioned, @to-sta :) We can close this when we see it's working, and let me know what your thoughts are for a similar workflow for the frontend!

to-sta:
@andrewtavis i will think about the frontend 😊. I also want to check [flake8](https://flake8.pycqa.org/en/latest/) additionally to `black` for the backend. Since black just checks the formatting but does not lint. Might be useful in the future to catch unused imports and so on.

I am also stoked to see the next PR 😃

andrewtavis:
All sounds good, @to-sta! Once this is done I’ll edit the maintainer checklist point that @TeddyGavi is working on in #314 so that we have one that’s for confirming that all checks have passed :)

to-sta:
@andrewtavis Iet's close this. I will make a new one for the frontend.

andrewtavis:
Sounds good, @to-sta! Thanks for the work on this one and looking forward to the next!

PR #308, opened 2023-08-26
#228 test to see if the typecheck will work

andrewtavis:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

Note that this is just a test to see if the typecheck works.

### Related issue

- #228

PR #317, opened 2023-08-28
Added Black formatting Github Action

to-sta:
### Contributor checklist

---

### Description

This PR adds a GitHub Action workflow that checks if python code is formatted with black. The workflow targets the `backend` folder and excludes migrations. I have excluded migrations since they are autogenerated by django.

I have tested the black command:
```powershell
black ./backend --check --verbose --exclude "migrations/" 
```
Here is the result:
```
...
would reformat C:\Users\tobia\Documents\Projects\contrib\activist\activist\backend\apps\models.py
Oh no! 💥 💔 💥
1 file would be reformatted, 14 files would be left unchanged.
```
So currently the Github Action should fail because the models.py is not formatted with black.

I tried to test the Github Action locally with [act](https://github.com/nektos/act) but ran into a known [issue](https://github.com/nektos/act/issues/1765) that would have required me to install a 14gb image for testing. So i skipped testing :). Guess we have to give this a test via PR.

### Related issue

- #316

### DIFF
--- a/.github/workflows/pr_python_formatting_check.yaml
+++ b/.github/workflows/pr_python_formatting_check.yaml
@@ -1,0 +1,22 @@
+name: pr_python_formatting_check
+on:
+  pull_request_target:
+    branches:
+      - main
+    types:
+      - opened
+  workflow_run:
+    workflows: [pr_maintainer_checklist]
+    types:
+      - completed
+
+jobs:
+  black:
+    name: Run PR Python formatting check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: psf/black@stable
+      with:
+        options: --check --verbose --exclude "migrations/"
+        src: ./backend


andrewtavis:
Thanks @to-sta! Happy to test via PR - just ask #304, #305, #308, #307 and #308 😇🤦‍♂️😅😅

Will merge this in in the coming days :)

andrewtavis (approved):
Is looking good, @to-sta! Let's see how it works :) Do you think we need a formatting check for the frontend? We're generally using Prettier for that. 

Minor changes I'll do to this are capitalize the name throughout for consistency with the others and add a last line to the file. FYI this is something that you might want to check out in your env setup! It's generally [recommended to always include one](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file), and I have VS Code setup to add it automatically :)

Issue #323, opened 2023-08-31
add frontend formatting check


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In the styleguide [prettier and headwind](https://github.com/activist-org/activist/blob/main/STYLEGUIDE.md?plain=1#L123) is mention for formatting frontend code. Would make sense to implement a workflow to ensure its conform with the styleguide hence prettier and headwind.

### Contribution

I can help with creating a GitHub workflow. I am sure i am able to setup workflow with prettier but i have to do some research on how to incorporate headwind as well.

andrewtavis:
f80d9a5 sent through fixes just now, @to-sta!

andrewtavis:
Note that we have some errors being reported in the most recent PR (#326), @to-sta: https://github.com/activist-org/activist/actions/runs/6041410258/job/16394314324

Is this something we should try to fix on our end, or would you say we should try to ignore something on the Prettier side?

to-sta:
It worked 😃.

Here are warnings:
```
[warn] frontend/components/grid/GridGithubShields.vue
[warn] frontend/components/Loading.vue
[warn] frontend/components/media/MediaMap.vue
[warn] frontend/components/PageBreadcrumbs.vue
[warn] frontend/components/sidebar/right/SidebarRight.vue
[warn] frontend/composables/useFormCheckboxRadio.ts
[warn] frontend/composables/useFormSetup.ts
[warn] frontend/pages/home/index.vue
[warn] frontend/tsconfig.json
```

These should be all solved by formatting the files with prettier. I tried the first one and it only changes the formatting.

Here are the errors:
```log
Error:  frontend/types/event.d.ts: SyntaxError: A property signature cannot have an initializer. (5:18)
Error:    3 |   tagline: string;
Error:    4 |   organizer: string;
Error:  > 5 |   type: string = "act" | "learn";
Error:      |                  ^^^^^^^^^^^^^^^
Error:    6 |   topic: string;
Error:    7 |   description: string;
Error:    8 |   getInvolvedDescription: string;
Error:  frontend/types/feed.d.ts: SyntaxError: A property signature cannot have an initializer. (2:22)
Error:    1 | export interface Feed {
[...]

PR #324, opened 2023-08-31
added frontend formatting check and stand. steps

to-sta:
### Contributor checklist

---

### Description

This PR adds a github workflow to check frontend formatting with [prettier](https://prettier.io/). The workflow file `pr_frontend_formatting_check.yaml` was added. It targets the frontend folder and checks the formatting. Files are not changed but this can be changed in the long run with `dry: False`. 

Similar to #316 i tried to use [act](https://github.com/nektos/act) to run the GitHub action locally but ran into a known issue.

<ins>Additional</ins>:
I standardize the steps inside pr_python_formatting_check.yaml. Minor change that makes the github actions logs more readable. We have to check this with a PR again 😃.

<ins>Suggestions</ins>:
I named the file `pr_frontend_formatting_check` could also be `pr_prettier_formatting_check` (or something else 😄) but i do like the frontend and backend distinction and would suggest to rename `pr_python_formatting_check` to `pr_backend_formatting_check.`

### Related issue

- #323

### DIFF
--- a/.github/workflows/pr_backend_formatting_check.yaml
+++ b/.github/workflows/pr_backend_formatting_check.yaml
@@ -1,4 +1,4 @@
-name: pr_python_formatting_check
+name: pr_backend_formatting_check
 on:
   pull_request_target:
     branches:
@@ -12,7 +12,7 @@ on:
 
 jobs:
   black:
-    name: Run PR Python Formatting Check
+    name: Run PR Backend Formatting Check
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/pr_frontend_formatting_check.yaml
+++ b/.github/workflows/pr_frontend_formatting_check.yaml
@@ -1,4 +1,4 @@
-name: pr_python_formatting_check
+name: pr_frontend_formatting_check
 on:
   pull_request_target:
     branches:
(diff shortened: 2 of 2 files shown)


andrewtavis (approved):
Looks great to me, @to-sta! Thanks for setting all this up 😊 As before let's wait for the next PR and then we can judge how it's all working and close the issue 🚀

I'm of the opinion that we should keep it at `dry: True`, but let me know what your thoughts are. To me this is a simple community responsibility where if it's not formatted we ask them to install the needed formatters, which further will help them contribute in the future via an appropriate environment. Unformatted code also would be an indication that there are other problems in the person's setup, so it gives us an opportunity to check in :)

to-sta:
@andrewtavis just checked the GitHub workflow on my side and made minor changes.
The `working-directory` key value pair did not work as intended but we can also use the `prettier_options` to set this up correctly.

```yaml
jobs:
  prettier:
    name: Run PR Frontend Formatting Check
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Run Prettier
        uses: creyD/prettier_action@v4.3
        with:
          dry: True
          prettier_options: ./frontend --check
```

I do agree on keeping `dry: True` since this is a not a big issue. Wrongly formatted code won't break the build 😄. Just mention this because i saw this in the styleguide:

> ...We'll eventually set it up so that code is autoformatted within the PR flow...

andrewtavis:
I'll correct this and remove the line above, @to-sta!

Issue #318, opened 2023-08-29
View lower on back button press navigation


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

We have a minor bug that the view of the screen is shifting down when a user presses the back button - i.e. the user navigates to a new page, then goes back, and then their view on the original page is lower than it was before.

I think that the following could be related to the [Loading.vue](https://github.com/activist-org/activist/blob/main/frontend/components/Loading.vue) component. The height of the shift is roughly that of the screen, so maybe this component is loading, shifts the view down, and then disappears. 

Note: this is on Chrome and Firefox at least.

henrque:
Hey, I'm new to the open source community. I'd like to take on this task. Could you assign it to me?

andrewtavis:
Hey @henrque! Thanks for reaching out and for your interest in activist :) Assigning you now, and please let me know if you have any questions 😊

andrewtavis:
Just checked it live as well and can confirm that #326 fixes this 😊 Thanks again, @henrque! 🚀

PR #326, opened 2023-08-31
fixing loading component

henrque:
### Contributor checklist

---

### Description
When user navigates to a new page, then goes back, and then their view on the original page is lower than it was before.
This was caused because the loading component was not fixed, making the page going down.
To solve this problem I fixed the loading component and setting z-index to be above others components.

### Related issue

- #318

### DIFF
--- a/frontend/components/Loading.vue
+++ b/frontend/components/Loading.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    class="flex items-center justify-center h-screen bg-light-header dark:bg-dark-header"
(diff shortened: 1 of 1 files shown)


andrewtavis:
Thanks for the PR, @henrque! Don't worry about the failed checks as the TS errors aren't from you and we working on the formatting check now 😊 I'll review this soon!

@to-sta, you have any ideas on the frontend check here? We're getting a formatting error on some of the types files 🤔

Issue #322, opened 2023-08-30
Refactoring Initial Django Project into Multiple Apps


momanyisamuel:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The current structure of the Django project is monolithic, as the project grows and to enhance maintainability, readability, and modularity, as discussed in #174 activist-org proposes refactoring the project into multiple apps. This will involve splitting the existing backend codebase into several apps, each with its own dedicated responsibilities. The proposed structure is as follows:

1. Authentication App (Support, User):
Responsible for managing user authentication and support functionalities.

2. Contents App (Resource, Task, Topic):
Handles resources, tasks, and topics within the project.

3. Groups & Orgs App (Organization, Group, Movement):
Deals with organizational and group-related functionalities.

4. Events App (Format, Role):
Manages event formats and roles within the application.

Each app will follow a similar structure:

`
app_name/
│
├── migrations/
│
├── templates/
│ └── app_name/
│
├── admin.py
│
├── forms.py
│
├── apps.py
│
├── models.py
│
├── tests.py
│
├── views.py
│
├── urls.py
│
├── signals.py
│
└── serializers.py

`

### Contribution

Happy to support someone on this issue or work on it myself

kurtisolo:
If you'd like, I can take care of it! Just one thing: should the 4 apps be placed in the project's main folder or contained within the "apps" subfolder? Typically, having them in the main folder is fine, but just these days, I had to modify the folder structure in another Django project, and it was a nightmare.

So, if we want the apps grouped in a folder, it's now or never!

andrewtavis:
Thanks @kurtisolo! Assigning you now :)

To me putting them in the apps folder makes more sense to really mark that we’re using this structure 😊 If you’d suggest main, then we can go that route as well, but I have no particular preference for it.

kurtisolo:
I'm almost done with this. I decided to keep the apps in the main directory, as this is thestandard practice so it's more suitable for an open source project.

I called the apps: 
- authentication
- contents
- entities (I was undecided on what to name this app, so I went for an inclusive term. If you have any other ideas or preferences, please let me know)
- events

If there are no objections, I can make the PR soon!

andrewtavis:
[...]

PR #330, opened 2023-09-01
fixed models sequence

kurtisolo:
### Contributor checklist

---

### Description

I tried to reorder the models as suggested, but in some cases, it wasn't possible because the sequence was enforced. I also added an index at the beginning of each module; I believe this, combined with the 'follow-the-references' approach, should make the code readability sufficiently clear. Let me know if you think it can be further improved!

### Related issue

- #322


andrewtavis:
Thank you, @kurtisolo! Will review soon :) :)

andrewtavis (approved):
Thanks, @kurtisolo! I'll do some minor edits, but we should be able to close this up then :)

PR #331, opened 2023-09-01
Prettier config and workflow setup

to-sta:
### Contributor checklist

---

### Description

This PR is related to the recently merged PR #324. It builds on the GitHub workflow but instead of using a predefined Github Action `creyD/prettier_action@v4.3` we manually install prettier an run it. 

Prettier was installed via npm and configuration files were added. 

I ran a test of prettier and the GitHub workflow with act. Both worked and only [warn] showed up, where most of changes are the end of line sequence from `crlf` to `lf`. Prettier uses `lf` by default, we could overwrite the default in the `.prettierrc`.

```json
"endOfLine": "crlf"
```
I did not run `npx prettier . --write` to make the changes yet.

### Related issue

- #323

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "prettier.configPath": "frontend/.prettierrc",
+    "prettier.ignorePath": "frontend/.prettierignore"
+}
--- a/frontend/.prettierrc
+++ b/frontend/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": false
+}
--- a/.github/workflows/pr_frontend_formatting_check.yaml
+++ b/.github/workflows/pr_frontend_formatting_check.yaml
@@ -17,8 +17,14 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Run Prettier
-        uses: creyD/prettier_action@v4.3
-        with:
-          dry: True
-          prettier_options: ./frontend --check
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+      
+      - name: Install prettier
+        run: npm install prettier
+      
+      - name: Run prettier
+        working-directory: ./frontend
+        run: |
+          npx prettier . --check
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # Visual Studio Code
 .vscode/*
+!.vscode/settings.json
 !.vscode/extensions.json
 
 # Django and Python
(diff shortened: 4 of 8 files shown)


frontend/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "trailingComma": "es5",
```
andrewtavis, line 2: What are we doing for trailing commas with this? In looking at it a little bit more, I think that I might have been convinced that my hereto dislike of trailing commas at the end of lists and arrays might be unwarranted. Might be preferable even as it'd give us nicer git diffs and also would making extending lists easier (I've not enjoyed all the JSON work with the L10N files where I need to add the comma myself all the time). How do you feel about enforcing trailing comma addition?

frontend/.prettierrc
```diff
@@ -0,0 +1,6 @@
...
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": false
```
andrewtavis, line 5: Yessss to all of the above :) :)

andrewtavis:
And doing a bit of research, `crlf` is two bytes instead of one, so I'm ok with keeping the prettier default. It would just be a matter of those of us on Mac and Linux going through and removing/adding in the end line again in some cases to make sure that we get the same type of break?

frontend/.prettierrc
to-sta: I don't have a strong opinion on this. Personally i do not write trailing commas and let the formatter take care of it. My default setting in VScode was also "es5".

Just checked the docs for the default and had changed:

> Default value changed from es5 to all in v3.0.0

I am happy with any of the three options.

andrewtavis: So as of now is it setup to add the last comma to it, @to-sta? If not, can that be swapped in a commit? Happy to merge once that's done :) :)

to-sta: Yeah, it will add a trailing comma.

andrewtavis: Thanks, @to-sta :) Has been a long day of traveling and just wanted to check :)

andrewtavis (approved):
Thanks for all the work here, @to-sta! Merging this in, and let's see how the next PRs interact with it 😊 Am realizing as well that there's now a `package-lock.json` file here, which I'll remove quickly after merging :)

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "prettier.configPath": "frontend/.prettierrc",
+    "prettier.ignorePath": "frontend/.prettierignore"
+}
```
wkyoshida, line 4: Sorry, real quick - was looking at some of the `prettier` additions.

Could having some of these configs in the `.vscode` configs cause some issues? Some individuals may not be using VSCode. Or if they are, could this conflict with any `settings.json` files that folks may already have? Just wondering

to-sta: Added an issue for it #363 and will work on this in the next days😃

Issue #291, opened 2023-08-25
a11y: SidebarLeftFooter's tab control order needs to be fixed


luuu-xu:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

While I was working on fixing TS errors in `SidebarLeftFooter`, I tried using tab and found out the tab order can be improved.

The current behaviour: when tab selects either "Create", "Info" and "Username", now presses enter key, the selected button is clicked, and expanded. but the tab focus is now focusing on the last button "Username".

The correct behaviour: when the selected button is entered, the **tab focus should instead focus on the first of these according sub buttons**. Example: when "Create" is entered, sub buttons "New Event", "New Organization", "New Group" and "New Resource" are shown, and the tab focus should be on "New Event".

### Contribution

I would love to work on this myself!

andrewtavis:
Thanks for bringing this up, @luuu-xu! Assigned 😊 Let me know if you need any support :) :)

andrewtavis:
Closed via #333! Thank you, @luuu-xu 🙌

PR #333, opened 2023-09-01
Fix: SidebarLeftFooter tab control order fix and misc fixes

luuu-xu:
### Contributor checklist

---

### Description

1. `SidebarLeftFooter`: fixes tab control order misbehaviour, now pressing enter key on the DisclosureButton will move focus to the first child element in DisclosurePanel.
2. `HeaderAppPage`: fixes TS props error where props `header` and `tagline` was required before, but the normal use case is that either `organization`, or `event`, or `header + tagline` is passed into this component.
3. `nuxt.config.ts`: fixes console error messages about our `swiper-container` and `swiper-slide` custom html elements in `MediaImageCarousel` component. See this [StackOverflow thread](https://stackoverflow.com/questions/71601714/how-to-set-compileroptions-iscustomelement-for-vuejs-3-in-laravel-project). The error messages in console was like this [one](https://laracasts.com/discuss/channels/vue/how-to-set-compileroptionsiscustomelement-for-vuejs-3-in-laravel-project).

### Related issue

- #291

### DIFF
--- a/frontend/components/header/HeaderAppPage.vue
+++ b/frontend/components/header/HeaderAppPage.vue
@@ -27,8 +27,8 @@ import type { Event } from "../../types/event";
 import type { Organization } from "../../types/organization";
 
 const props = defineProps<{
-  header: string;
-  tagline: string;
+  header?: string;
+  tagline?: string;
   organization?: Organization;
   event?: Event;
 }>();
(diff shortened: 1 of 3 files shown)


andrewtavis:
Hey @luuu-xu, one comment on this: we should make it so that when a user closes the dropdown rather than opens another one that the focus stays on the dropdown that it's currently on. As of now it goes to one of the other ones 😊 So tab to `Create`, open it and it goes directly to `New event` (awesome, btw!), then tab back to `Create` and close it, but then it jumps to `Username` :)

luuu-xu:
@andrewtavis I totally agree on your suggestion! I somehow missed this not great ux. Now it is fixed!

andrewtavis (approved):
Wonderful work, @luuu-xu! Thanks for the quick fix as well 😊 With this and the routing that I'll review now we're making so many strides in terms of navigation!! 🚀

Issue #348, opened 2023-09-06
Backend Django issues


to-sta:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Currently if we run the django server it won't start. The errors can be seen below:
Fixing the errors is a prerequisite for testing other backend tasks #329.

```
Watching for file changes with StatReloader
Performing system checks...

Exception in thread django-main-thread:
Traceback (most recent call last):
  File "C:\Users\tobia\AppData\Local\Programs\Python\Python310\lib\threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "C:\Users\tobia\AppData\Local\Programs\Python\Python310\lib\threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\tobia\Documents\Projects\contrib\activist\activist\venv\lib\site-packages\django\utils\autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "C:\Users\tobia\Documents\Projects\contrib\activist\activist\venv\lib\site-packages\django\core\management\commands\runserver.py", line 133, in inner_run
    self.check(display_num_errors=True)
  File "C:\Users\tobia\Documents\Projects\contrib\activist\activist\venv\lib\site-packages\django\core\management\base.py", line 556, in check
    raise SystemCheckError(msg)
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
<class 'entities.admin.OrganizationAdmin'>: (admin.E108) The value of 'list_display[2]' refers to 'application_id', which is not a callable, an attribute of 'OrganizationAdmin', or an attribute or method on 'entities.Organization'.
authentication.Support.supported_type: (fields.E304) Reverse accessor 'SupportEntityType.support_set' for 'authentication.Support.supported_type' clashes with reverse accessor for 'authentication.Support.supporter_type'.
        HINT: Add or change a related_name argument to the definition for 'authentication.Support.supported_type' or 'authentication.Support.supporter_type'.
authentication.Support.supporter_type: (fields.E304) Reverse accessor 'SupportEntityType.support_set' for 'authentication.Support.supporter_type' clashes with reverse accessor for 'authentication.Support.supported_type'.
[...]

PR #351, opened 2023-09-06
Backendfix

to-sta:
### Contributor checklist

---

### Description

In this PR i fixed a few minor issues that prevented django from running the server properly. I also added makemigrations and migrate to the docker-compose.yaml. Without it we would have need to keep a initial migration for the user model to run the server. Another benefit of adding makemigrations and migrate is that we are now able to check if there are issues within the models.py if we make changes. 

I  tested it locally (with a virtual enviroment) and i ran docker-compose to check if the changes worked. 
Now it's possible to access the routes like the admin page for example:
`http://localhost:8000/v1/admin/`

Let me know if you want the dockerfile to hold all the cmd for django instead of the docker-compose file. That was just easier for me :). I am not sure how this would be done in the dockerfile given that django needs the postgres db connection before running the commands.

### Related issue

- #348


momanyisamuel (approved):
Thanks for these changes, looking alright @to-sta

andrewtavis:
@momanyisamuel, did you mean to merge? Do you have that ability? Maybe as tests have failed only I can?

momanyisamuel:
yeah i thought approving would merge @andrewtavis please go ahead i was just about to reopen the issue since that did not happen

andrewtavis:
I’ll merge it in tomorrow or later on tonight :) I’m enjoying a beer right now as you told me to 🍻😇

momanyisamuel:
By all means enjoy @andrewtavis pardon the interruption hehe

backend/authentication/models.py
```diff
@@ -33,8 +33,6 @@ def __str__(self):
 class Support(models.Model):
     supporter_type = models.ForeignKey("SupportEntityType", on_delete=models.CASCADE)
     supporter_entity = models.IntegerField(null=True)
-    supported_type = models.ForeignKey("SupportEntityType", on_delete=models.CASCADE)
```
andrewtavis, line 36: These two lines shouldn't be removed, right? Maybe the way that this table is set up is flawed, but we have `supporter` and `supported`. The type is either org or user, and the entity would be the corresponding id so that we can keep things as integers throughout.

to-sta: Thanks for the review @andrewtavis. In that case we have to supply a kwarg `related_name` to resolve the issue.

andrewtavis: Ah so it's problematic in so far as they're both referencing the same foreign key, @to-sta? Makes sense. If we can make that change we should be good here 😊

andrewtavis: Checks are looking great, btw! Looking forward to finishing up the TS fixes so we're back to all green ✅

andrewtavis: For this @to-sta, we'd have `related_name="supporter"` and `related_name="supported"` respectively? Just checking the docs a bit and that's what first came to mind :)

to-sta: Makes sense :). I will commit the change now.

to-sta:
To make sure that everything is correct I queried the support table.

```postgres
activist=# SELECT * FROM  authentication_support;
 id | supporter_entity | supported_entity | supported_type_id | supporter_type_id 
----+------------------+------------------+-------------------+-------------------
(0 rows)
```
Looks all good 😊

andrewtavis (approved):
Thanks for getting to the changes so quickly, @to-sta! 😊 Appreciate you testing it as well 🙏

Issue #295, opened 2023-08-25
Creating Serializers with Standard Field Validation


momanyisamuel:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

To ensure data integrity and maintain consistent data quality across activist-org's Django Rest Framework (DRF) api, we need to implement serializers that enforce standard validation for fields. By enforcing validation rules consistently across the project, we can prevent incorrect or invalid data from entering our system and enhance the reliability of activist-org's API.

Proposed Task:
The proposed task involves creating custom serializers - given all the models have been defined - that enforce standard validation rules for fields. By implementing these rules within the serializers, we can ensure that incoming data adheres to predefined criteria before being processed and stored.

LATER - we can look at implementing custom validators

### Contribution

Happy to support someone on this issue or work on it myself

thiagogquinto:
Hey @momanyisamuel , I am new in open source community, I would like to take a look in the source code before I decide to develop this task, so I will assign myself, if something changes I remove muself then, ok?

andrewtavis:
Hey @thiagogquinto! Thanks for your interest in activist :) I'll assign you, and please let us know if there's anything we can do to help!

thiagogquinto:
Hey @andrewtavis, sorry for taking so long, I had some problems, but I just want to update and tell you that I am working on the issue,

andrewtavis:
Hey @thiagogquinto 😊 Not a problem :) I appreciate you communicating with us on the progress! Please let us know if you want to discuss it further or need some help.

kurtisolo:
I can help with this, if needed :) (sorry for my absence btw @andrewtavis )

andrewtavis:
Thanks for offering to help, @kurtisolo, and no stress on any absence :) Let's give @thiagogquinto a bit more time to work on this :) @thiagogquinto, @kurtisolo would also be a great resource to ask for help on this 😊 Take your time and we'll get this done! 🚀

@kurtisolo, we'll have a few more backend issues from @momanyisamuel later on today :) Would be great to have you in the sync tomorrow, btw!

thiagogquinto:
[...]

PR #358, opened 2023-09-08
Create serializers with validators 

thiagogquinto:
### Contributor checklist

---

### Description

This *Pull Request* implements serializers with validators for all the models already defined. 

### Related issue

- #295


andrewtavis:
Thanks for this, @thiagogquinto! I'm a bit busy today and the next few days, but we'll check this for you and provide some feedback 😊

CC @momanyisamuel and @kurtisolo

kurtisolo:
In my opinion, it looks good; I haven't tested it, but I don't see any obvious issues. The addition of utils also makes sense to me. Maybe it could be done differently, but for now, it seems fine. Perhaps I would suggest using:
```
 class Meta:
     model = EventAttendee
     fields = '__all__'
```
for each class in serializers.py to reduce the amount of code and make it more readable?

thiagogquinto:
Thanks @kurtisolo, you are absolutely right, this way looks much better.

backend/authentication/serializers.py
```diff
@@ -1,23 +1,89 @@
...
+            )
+        return data
+class SupportSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Support
+        fields = '__all__'
+
+    def validate(self, data):
+        
+        if data["supporter_type"] == data["supported_type"]:
```
andrewtavis, line 25: This one isn't needed as they can be the same if say a user is supporting another user. The entity cannot be the same for sure, but then for this we need to keep in mind that we need to not repeat ids for organizations for those of users :)

I can remove this. No stress! 😎

andrewtavis:
Basic question on my part: is it standard that validation errors are incomplete sentences? I'd assume that these should follow Python full-line comment standards with capitalization and a period at the end (inline is lower case and no period as is present). Will merge regardless, but in case there's an explicit preference in the Django community I'll likely go through and change these :)

backend/entities/serializers.py
```diff
@@ -1,18 +1,189 @@
...
+
+        return data
+
+
+class OrganizationMemberSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = OrganizationMember
+        fields = "__all__"
+
+    def validate(self, data):
```
andrewtavis, line 68: This one here is also not necessary. Would be great to get feedback on the various roles we'll include for orgs, but generally we just have these three right now as special roles with more permissions :)

andrewtavis (approved):
Sorry for a bit of a wait on this review, @thiagogquinto! Really appreciate all the care that you've put into this PR, and hope to have more contributions from you in the future given your knowledge of what we're building in the backend now! 🚀 I'll go through now and remove the two mentioned validations that aren't necessary. This is really great! 😊

Issue #329, opened 2023-09-01
Look into isort and mypy for backend environment


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

It would be good to check the packages [isort](https://github.com/PyCQA/isort) and [mypy](https://github.com/python/mypy) to see if we should add them to the backend dependencies. Maybe `black` will already cover `isort`, but adding in `mypy` for static typing would likely be very worthwhile 😊

### Contribution

Happy to discuss this with people and either help implementation or do it myself!

to-sta:
I am interested in working on the mypy portion of this issue 😃.

andrewtavis:
Me too, @to-sta 😝😝😅😊 But you can take it :) Can you just confirm that with `black` we don’t need `isort`?

to-sta:
`black` does have an overlap with `isort` but does not replace the use of `isort`. Additionally i would also recommend the use of `flake8` or [ruff](https://github.com/astral-sh/ruff ) and to configure all 3 in a pyproject.toml file for example.

I think it would make sense to seperate mypy and isort (potentially black, ruff) into another issue. Since they can be worked on independently of each other.

If you want to work on mypy then i could switch to working on isort, black and ruff 😃.

In the end they all come together in the backend GitHub workflow 😃.

andrewtavis:
Would `mypy` also have a workflow run to check for it? I mostly just want to try to implement the types Python code :) We can focus on `mypy` in this one and then you could make another issue for the others? 😊

andrewtavis:
I’m also fine with ruff over flake8, btw :)

to-sta:
Yes, we can just add `mypy` to the backend GitHub workflow and instead of looking for a pre-defined GitHub action for it we can install via pip and run it in the workflow.

Then i will start working on `mypy` and create another issue for `black`, `isort` and `ruff` afterwards.

andrewtavis:
Would you say this issue is good to close, @to-sta? Or are we waiting to see how it works? Also, do you want to talk a bit at the dev sync about the recent changes? I'd be happy to update the schedule with a section for you to go over some of the updates 😊

to-sta:
[...]

PR #367, opened 2023-09-10
Type hints / mypy errors

to-sta:
### Contributor checklist

---

### Description

This PR adds type annotations to the python backend. It solves the current errors from running mypy.

I ran mypy till there were no errors anymore:
`Success: no issues found in 42 source files`

Additional: i ran the `docker-compose up --build` to check if build works as expected.

### Related issue

- #329

### DIFF
--- a/backend/authentication/admin.py
+++ b/backend/authentication/admin.py
@@ -3,7 +3,7 @@ from django.contrib import admin
 from .models import User
 
 
-class UserAdmin(admin.ModelAdmin):
+class UserAdmin(admin.ModelAdmin[User]):
     list_display = ("username", "first_name", "last_name", "email", "creation_date")
 
 
--- a/backend/authentication/models.py
+++ b/backend/authentication/models.py
@@ -26,7 +26,7 @@ class SupportEntityType(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     name = models.CharField(max_length=255)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.name
 
 
@@ -40,8 +40,8 @@ class Support(models.Model):
     )
     supported_entity = models.IntegerField(null=True)
 
-    def __str__(self):
-        return self.id
+    def __str__(self) -> str:
+        return str(self.id)
 
 
 class User(AbstractUser):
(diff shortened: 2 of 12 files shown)


backend/authentication/admin.py
```diff
@@ -3,7 +3,7 @@
 from .models import User
 
 
-class UserAdmin(admin.ModelAdmin):
+class UserAdmin(admin.ModelAdmin[User]):
```
andrewtavis, line 6: I deleted some of the `[Class]` arguments in the merge conflict update as it looks like they weren't used in other places in the PR for #295. Let's keep this in mind and add them back in if necessary. Sorry, just trying to get through all of these and see the new state of things, but we can do a clean up at the end 😊

backend/authentication/models.py
```diff
@@ -1,89 +1,89 @@
...
+        "SupportEntityType", on_delete=models.CASCADE, related_name="supporter"
+    )
+    supporter_entity = models.IntegerField(null=True)
+    supported_type = models.ForeignKey(
+        "SupportEntityType", on_delete=models.CASCADE, related_name="supported"
+    )
+    supported_entity = models.IntegerField(null=True)
+
+    def __str__(self) -> str:
+        return str(self.id)
```
andrewtavis, line 44: We need `str()` here as we have this as a numeric field, so this is necessary here but isn't for `return self.username`?

backend/authentication/models.py
```diff
@@ -1,89 +1,89 @@
...
+    verification_method = models.CharField(max_length=30, null=True)
+    verification_partner = models.ForeignKey(
+        "User", on_delete=models.SET_NULL, null=True, blank=True
+    )
+    social_accounts = ArrayField(models.CharField(max_length=255), null=True)
+    total_flags = models.IntegerField(default=0)
+    creation_date = models.DateTimeField(auto_now_add=True)
+    deletion_date = models.DateField(null=True)
+
+    def __str__(self) ->str:
```
andrewtavis, line 64: Space needed (I'll get it).

backend/authentication/admin.py
andrewtavis: These were added back in in ef1c8a8 :)

backend/authentication/models.py
andrewtavis: Fixed in ef1c8a8 :)

backend/authentication/models.py
to-sta: @andrewtavis exactly! It's necessary for all non string fields (IntegerField, DateField,...). The __str__ method is special case where the `return` datatype is already determined to be a `str`. What we specifiy as return value will be shown in the admin view (`localhost:8000/v1/admin/`).

Issue #360, opened 2023-09-09
Implement DRF ViewSets for CRUD Operations on Activist Resources


momanyisamuel:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This is an issue to optimize activist-org API structure by utilizing Django Rest Framework's (DRF) ViewSets for CRUD (Create, Read, Update, Delete) operations on activist resources. This will help streamline activist-org's codebase and enhance the maintainability of the API endpoints.

### Contribution

Happy to support someone on this issue or work on it myself

kurtisolo:
I can work on this if no one steps up :)

andrewtavis:
Happy to assign, @kurtisolo! :)

kurtisolo:
In `entities/views.py`, there are some Django class-based views. For consistency, I would modify them to use DRF views. Is that okay?

momanyisamuel:
@kurtisolo yes you can modify them, I created those before we added DRF to the project 👍

andrewtavis:
Closed via #378 and a bit of formatting in 145d08a 😊 Thanks for the commit @kurtisolo and for the issue @momanyisamuel! 🙌 Would be great if the both of you could think of a few more issues for the backend for code night. I should be able to switch over and help out more soon, but this will likely be next week at the earliest :)

Issue #215, opened 2023-07-07
Improve a11y specifically to reach AA/AAA colour contrast for labeled buttons


luuu-xu:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

For our labeled buttons that is in brand orange colour, having white text inside does not meet WCAG colour contrast standards, see below:

<img width="327" alt="Screenshot 2023-07-06 at 21 57 35" src="https://github.com/activist-org/activist/assets/97932191/432aa843-1088-4b33-854a-26f470b4c122">

Changing the brand colour could be one solution, or we can have black text inside. 

And for our link within text, the colour blue is on the light side, we can make it a bit darker to reach AA rating, see below:

<img width="327" alt="Screenshot 2023-07-06 at 21 58 37" src="https://github.com/activist-org/activist/assets/97932191/5ebf3d23-5712-422b-92b8-dadf47a371fd">

### Contribution

I can make these easy changes once the design is considered and reached a conclusion.

andrewtavis:
Reposting the tools that I generally use for this, @luuu-xu :)

- https://coolors.co/ for palettes and getting darker/lighter shades of colors
- https://polypane.app/color-contrast for a quick contrast check

andrewtavis:
With each of these color changes we need to consider multiple backgrounds, specifically `content`, `distinct` (for cards and the sidebar content) and `header` (web and sidebar header/footer). Let's start with the link colors as they're maybe a bit more simple :) The alpha is definitely the part that's causing problems as it looks like when I checked I didn't add that in.

Based on checking some colors we could change `light-link-text` to `005CB8` while maintaining the alpha of 0.9, which would then give us AA accessibility over each of the background? 

If this sounds good you'd be welcome to open a PR. I've updated the [designs](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=805%3A231&mode=design&t=ApCWorDmDSuUVMt6-1) to use the new color and it's looking good :)

andrewtavis:
By the looks of it our dark mode link text’s blue is already AA, so this PR would just be to fix the light mode color :)

luuu-xu:
@andrewtavis Thank you for your ideas!

[...]

PR #378, opened 2023-09-14
Add viewsets

kurtisolo:
### Contributor checklist

---

### Description

So, I've fixed several things, and I hope I remember them all.

- `BASE_DIR/urls.py`: I added 4 different paths that link to their respective apps.
- `urls.py` of the 4 apps: I added the '`app_name`' value for more clarity; it's a good practice when using the '`include`' method.
- `OrganizationViewSet`: In `entities/view.py`, I converted the pre-existing functions to DRF, retaining the same functionalities. I also added some '`get_object_or_404`' which never hurts.

**Note**: consider this as a starting point; the views will be updated as new API requests become necessary.

**Note 2**: I noticed that all `ModelSerializers` serialize all the fields of the models. This may be fine as a starting point, but as the endpoints are used, it will be necessary to evaluate, one by one, which fields should actually be exposed in the endpoints, and which should remain confined to the backend for security reasons. Perhaps I'm stating the obvious, but I wanted to point it out.

**Last note**: I fought with Docker today, so I haven't actually tested as much as I wanted to. Hope everithing is ok. Damn whale.

### Related issue

- #360


andrewtavis (approved):
Thanks for this, @kurtisolo! Let me know if you have some ideas for more issues that need to be made for the backend. Now that #215 is done, I should be able to more actively shift to helping with the backend development 😊

Issue #365, opened 2023-09-10
Shift to [id]/about on [id] when screen expands


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We currently have direct redirects to the `[id]/about` page for both events and organizations that is triggered above the medium breakpoint. Below this breakpoint we are directed to a page that has options for navigation. An issue is that if the screen is resized we get a stretched version of this page and are not directed to the `[id]/about` page. Would be great to get this redirect implemented!

### Contribution

Happy to support someone on this. This issue is for now reserved for @felipolis who expressed interest in dev sync 😊

felipolis:
I would love to work on this, @andrewtavis . Could you assign me?

andrewtavis:
Can do, @felipolis! Let me know if you need some support :)

andrewtavis:
Closed by #381 and minor changes in 4df81a4 to fix comment nits 😇 Thanks for the work here, @felipolis! Looking forward to the next issue! 🚀😊

PR #381, opened 2023-09-14
Redirecting to /about after resizing to a medium device

felipolis:
### Contributor checklist

---

### Description

Regarding the /events/[id] and /organizations/[id] pages, I made an improvement for the user experience. Now, when the user views these pages in a window with a width smaller than or equal to 640px and then resizes it to a larger width, I've implemented an event listener to track the size change.

This way, if the user resizes the screen to a width greater than 640px, they will be automatically redirected to the /about page. This ensures a smooth and intuitive transition for the user when they enlarge the window, maintaining the consistency and usability of the website.

### Related issue

- #365

### DIFF
--- a/frontend/pages/events/[id]/index.vue
+++ b/frontend/pages/events/[id]/index.vue
@@ -136,11 +136,11 @@ const handleResize = () => {
 }
 
 onMounted(() => {
-  // Verify that the user is on a mobile device
-  handleResize()
-
   // Add event listener to handle resizing
   window.addEventListener('resize', handleResize)
+
+  // Verify that the user is on a mobile device
+  handleResize()
 })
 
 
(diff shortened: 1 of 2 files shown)


felipolis:
@andrewtavis , just finished it :D

frontend/pages/events/[id]/index.vue
```diff
@@ -127,13 +127,21 @@ const eventButtons: MenuSelector[] = [
...
-function redirectBasedOnScreenSize() {
+const handleResize = () => {
   if (window.innerWidth > 640) {
+    window.removeEventListener('resize', handleResize)
     navigateTo(`${id}/about`);
   }
 }
+
+onMounted(() => {
+  // Add event listener to handle resizing
```
Comment (andrewtavis) [inline frontend/pages/events/[id]/index.vue:139]: Minor "nit" (nitpick) that there should be a period at the end here for full comments. I can add these myself though, @felipolis! We tend to do full sentences including punctuation for comments that are their own line and lower case the first letter/no punctuation for inline comments (Python styleguide, but it seems to be gaining traction elsewhere).

andrewtavis (approved):
Works like a charm, @felipolis! Thanks for the dedication to this feature! 🙌😊

Issue #383, opened 2023-09-15
Update admin.py


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Recently the `models.py` files have been updated but the `admin.py` files have not reflected the changes yet. The following files could be updated:

- `backend/authentication/`
- `backend/content/`
- `backend/entities/`
- `backend/events/`

I think the UserAdmin class inside `backend/authentication/` can be used as a guide to generate the rest.

### Contribution

I am happy to help or work on it myself 😄

andrewtavis:
Thanks for creating the issue and offering to work on it, @to-sta! I think it makes sense to combine this with #382 as well, so I'll head over and assign that one as well 😊

andrewtavis:
Closed via #384 :) Thanks for creating and working on the issue, @to-sta!

Issue #382, opened 2023-09-15
Update backend with new db schema entries


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

A few new entries have been added to the [database schemas on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377%3A33014&mode=design&t=sAtZhV2c6WKZH0eV-1) and need to be added to the backend. They include:

- The `Resources` table now has a `private` column
- The `Users` table now has `private` and `high_risk` columns
- The `Organizations` table now has a `high_risk` column

These columns represent very important aspects of activist that allow for private content and browsing as well as allowing users and organizations to eventually change the UI and other aspects of the platform in response to them needing more security.

### Contribution

Happy to work on this myself or guide someone who has interest in working on it :) This is also a `good first issue` for someone who wants to learn a bit more about activist's growing backend 😊

andrewtavis:
Thanks for offering, @to-sta! Assigning you now :)

andrewtavis:
Closed via #384 :) Appreciate you working on this, @to-sta!

PR #384, opened 2023-09-15
Update models

to-sta:
### Contributor checklist

---

### Description

In this PR, I extend the models.py tables Resources, Users and Organization according to the [database schemas on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377%3A33014&mode=design&t=sAtZhV2c6WKZH0eV-1). 

And added updated the admin.py files. 

I ran docker and created a superuser to view the admin page. All models are registered.

image

There are a lot of models now 😄, might make sense to restructure this further before merge or down the line.

Another option i like in case we know there will be more frequent changes in the models for a while, is to use a for loop inside the admin.py to iterate over the models and register them one by one. 

```python
import django.apps

models = django.apps.apps.get_models()

for model in models:
    try:
         admin.site.register(model)
    except admin.sites.AlreadyRegistered:
         pass


```
### Related issue

- #383 
- #382

### DIFF
--- a/backend/authentication/admin.py
+++ b/backend/authentication/admin.py
@@ -1,11 +1,14 @@
 from django.contrib import admin
+from django.contrib.auth.models import Group
 
 from .models import Support, SupportEntityType, User, UserResource, UserTask, UserTopic
+
+# Remove default Group
+admin.site.unregister(Group)
 
 admin.site.register(User)
 admin.site.register(UserResource)
 admin.site.register(UserTask)
 admin.site.register(UserTopic)
-
 admin.site.register(SupportEntityType)
 admin.site.register(Support)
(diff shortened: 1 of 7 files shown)


andrewtavis:
Thanks, @to-sta! I'll get to this tomorrow hopefully :) :)

to-sta:
Had another look into this and I think we should remove `Groups` under AUTHENTICATION AND AUTHORIZATION. This comes by default with Django. We could unregister it, since we do not use it.

andrewtavis:
Sounds good to me, @to-sta :) Feel free to send along another commit!

andrewtavis (approved):
Thanks for the changes here, @to-sta! Getting much closer to having a functioning backend with these 😊😊

Issue #391, opened 2023-09-22
Create sign in page


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create the `pages/sign-in` page that would thus translate to `activist.org/sign-in`. Designs for this page can be found in the [designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=6481%3A8298&mode=design&t=QtIT1amRX0lQy9zG-1) and are also shown below.

![Sign In - Username, Password and Verification](https://github.com/activist-org/activist/assets/24387426/521e2b7b-49f6-40b7-9b7e-babd7ee0f0e4)

![Mobile Sign In - Username, Password and Verification](https://github.com/activist-org/activist/assets/24387426/fa4d2e8e-35f0-48be-a23f-55b59c134c5d)

This issue could be worked on during the 23/9/23 Code Night 🌙 along with the issue for sign up #392 :)

### Contribution

Happy to work on this or support people who have interest in working on it!

ahmedy00:
Hi @andrewtavis 
I would like to develop this feature, as I will be involved in the project for the first time and as I am not a very experienced developer, I presume that this may take some time. If you don't see this as a problem, could you assign this to me? 
Thanks.

andrewtavis:
Hey @ahmedy00 👋😊 Great to have you here :) You're more than welcome to work on this, and no stress on not being the most experienced. Lots of people are learning by working on activist. I'll assign this issue to you, and please let me know if you need any assistance!

andrewtavis:
Hi @ahmedy00 👋 Wanted to say that @thiagogquinto will be working on the other similar issue to this, #392. Maybe you two could collaborate if there’re any issues 😊

ahmedy00:
@andrewtavis Ok it will be great to work together. Also I'm pushing some changes for this issue and I would like to if you tell me what's the mistakes and next steps

ahmedy00:
@andrewtavis Could you authorise me for contribute to the project

andrewtavis:
[...]

Issue #392, opened 2023-09-22
Create sign up page


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create the `pages/sign-up` page that would thus translate to `activist.org/sign-up`. Designs for this page can be found in the [designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=3228%3A22748&mode=design&t=QtIT1amRX0lQy9zG-1) and are also shown below.

![Sign Up 1 - Username, Password and Verification](https://github.com/activist-org/activist/assets/24387426/f271fe89-6bda-40fa-895c-e5bb698497f4)

![Mobile Sign Up 1 - Username, Password and Verification](https://github.com/activist-org/activist/assets/24387426/8ec03e70-cfa9-436f-a5c9-f3657a5209b7)

This issue could be worked on during the 23/9/23 Code Night 🌙 along with the issue for sign in #391 :)

### Contribution

Happy to contribute to this or support someone who wants to work on it!

thiagogquinto:
Hi @andrewtavis , I would like to work on it.

andrewtavis:
Sounds good, @thiagogquinto! Assigning you now :) Feel free to discuss with @ahmedy00 on what the plans for #391 are 😊

thiagogquinto:
Thansks @andrewtavis, just to be clear, I will  learn *Vue* by the issue, so if the task is important, feel free to assign to somebody else that wants to work on the issue,

andrewtavis:
There are enough examples in the code base for how to use Vue and Tailwind in the frontend that I’m confident you’ll be able to figure it out, @thiagogquinto 😊 Thanks for letting me know though, and ping me if you need assistance! You’re also welcome to send along a PR and we can talk about how it’s looking prior to further commits :)

ahmedy00:
@thiagogquinto Hey friend, If you want, we can have a voice meet about this issue and we can share our experiences with each other.

thiagogquinto:
> @thiagogquinto Hey friend, If you want, we can have a voice meet about this issue and we can share our experiences with each other.

[...]

PR #411, opened 2023-09-28
Create sign-in page.

ahmedy00:
### Contributor checklist

---

### Related issue

- #391


andrewtavis:
Thank you, @ahmedy00! First glances and everything's looking great 😊 Really appreciate you also including the i18n keys and even making a section in the JSON for #392! Happy to have your support here! 

Some minor comments:
- The content background color on the right side isn't being applied for dark mode
- We need a header applied to the page as as of now it's `activist • activist`
- Can we make the logo on the left a bit smaller and pop it up a bit?
  - I realize that it's centered, but sadly the way the eye works is that centered looks a bit low

Really wonderful first contribution! And thanks so much for already being such a communicative member of the community. Really great to see! 🚀

frontend/components/form/FormTextField.vue
```diff
@@ -0,0 +1,56 @@
...
+      :placeholder="placeholder"
+      :type="refInputType"
+    />
+    <span v-if="isIconVisible" @click="changeInputType" class="cursor-pointer">
+      <Icon name="mdi:eye" size="28px" />
+    </span>
+  </div>
+</template>
+
+<style scoped></style>
```
andrewtavis, line 56: No need to add in the style block if it's not being used :) And we normally do template before script, which I know is maybe a bit against what the normal ordering is. Would be good if you could share some feedback on why that's the standard, but for me it's always easier to think of the HTML first 😊

I can remove this myself. No stress!

andrewtavis (approved):
As stated in the call we had, @ahmedy00, this is really wonderful :) :) I'll bring it in and do some minor edits and let you know what it is that I changed, but overall I'm very happy with how this turned out! 😊

Again, really sorry for the review time. I was waiting till I had some time to really dive into it all post being out for a bit, but I should be more available for future work!

We've got the dev sync in two hours, btw! Details are in the Development channel on Matrix. Would be great to have you there!

Issue #398, opened 2023-09-22
Create top selector component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to create the top of page selector that will serve as a sub menu navigation for organization and event pages. The designs for it can be seen [on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2641%3A16666&mode=design&t=QtIT1amRX0lQy9zG-1) and can further be seen below:

![Top Selectors](https://github.com/activist-org/activist/assets/24387426/aee3d54a-0129-4292-a56d-56b08e8f6da7)

The name of this component could be something like `SelectorTopBar`, with it needing to extend the full width of its container. It should take a prop that defines the options, should also be styled for light and dark mode and include hover actions.

### Contribution

This is a great `good first issue` for someone who'd like to work on their first component, and can also be worked on as a part of our 23/9/23 Code Night 🌙😊 Happy to support someone if they'd need some assistance in working on this :) :)

sayik:
Hi Andrew, can I work on this issue?

andrewtavis:
Sure thing, @sayik! Assigning you now :) Let me know if I can do anything to help!

andrewtavis:
```ts
const props = defineProps({
  selectors: {
    type: string[],
    default: [""],
  },
  selected: {
    type: string,
    default: "",
  },
});

if props.selected === "" {
    props.selected = selectors[0]
}
```

andrewtavis:
```html
:class="{
  'bg-light-highlight dark:bg-dark-highlight': s === selected, 'bg-light-header dark:bg-dark-header': s !== selected
}"
```

andrewtavis:
Hey @sayik! fffaec1 finishes this up and also does some variable renames as I decided that this is better named `MenuTopBar` and the other selectors should be moved to a dropdown directory. 

Get well soon my friend! 🍵🌿

PR #415, opened 2023-10-04
#398 Add selector top bar

sayik:
### Contributor checklist

---

### Description

SelectorTopBar.vue was added, the file has a for loop, which checks for available top bar elements to display in organisation settings and event settings page.  "SelectorTopBar" tag was added to organisation settings file and event settings file accordingly.

### Related issue

- #398

### DIFF
--- a/frontend/components/selector/SelectorTopBar.vue
+++ b/frontend/components/selector/SelectorTopBar.vue
@@ -1,0 +1,18 @@
+<template>
+  <div v-if="selectors.length > 0" class="w-full h-fit">
+    <div class="flex flex-row"> 
+      <button
+        v-for="t in selectors"
+        class="w-full justify-center bg-light-header dark:bg-dark-header flex px-3 py-1 border-l-[1px] border-y-[1px] border-light-text dark:border-dark-text text-light-text dark:text-dark-text hover:bg-black/10 dark:hover:bg-white/30"
+      >
+        {{ t }}
+      </button>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+defineProps<{
+  selectors?: string[];
+}>();
+</script>
--- a/frontend/pages/events/[id]/settings.vue
+++ b/frontend/pages/events/[id]/settings.vue
@@ -1,4 +1,5 @@
 <template>
+  <SelectorTopBar :selectors="['Allies', 'Supporter', 'Supporting']"/>
   <div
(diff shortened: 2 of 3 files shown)


andrewtavis:
Thanks, @sayik! Will get to this soon :) :)

frontend/components/selector/SelectorTopBar.vue
```diff
@@ -0,0 +1,18 @@
...
+      </button>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+defineProps<{
+  selectors?: string[];
+}>();
+</script>
```
andrewtavis, line 18: For here we need a blank line at the end of the file, @sayik :) I'd set up VS Code to always add one via your settings.json file 🙃

frontend/components/selector/SelectorTopBar.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <div v-if="selectors.length > 0" class="w-full h-fit">
+    <div class="flex flex-row"> 
+      <button
+        v-for="t in selectors"
+        class="w-full justify-center bg-light-header dark:bg-dark-header flex px-3 py-1 border-l-[1px] border-y-[1px] border-light-text dark:border-dark-text text-light-text dark:text-dark-text hover:bg-black/10 dark:hover:bg-white/30"
+      >
```
andrewtavis, line 7: @sayik, one think I'm realizing is that we need to add styling for whichever option is selected. Do you want to work on this, or should I? We'd need another prop like say `selected` that's a string, and then we can do a check for if `s === selected` and change the styling if this is the case. How does this sound?

frontend/components/selector/SelectorTopBar.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <div v-if="selectors.length > 0" class="w-full h-fit">
+    <div class="flex flex-row"> 
+      <button
+        v-for="t in selectors"
```
andrewtavis, line 5: I think that `s` makes more sense here :)

frontend/components/selector/SelectorTopBar.vue
andrewtavis: For this we'd do a conditional `:class` like the following:

```
:class="{
  'bg-ETC': s === selected,
  'bg-ETC': s !== selected
}"
```

You can find examples of how to do this in many places in the code base by searching for `:class="` :)

andrewtavis (approved):
Heyo @sayik! As discussed, I'll merge this in and complete it for you 🤗

PR #417, opened 2023-10-07
Update CONTRIBUTING.md

gitgoap:
### Contributor checklist

---

### Description

The spelling of "focusing"  is more widely used than "focussing."

I changed this in 2 places.

This was not a major change so I did not raise an issue to fix this.


andrewtavis (approved):
Thanks for the fix here, @gitgoap! 🚀  We chatted in the dev sync we have and agree it should be this :)

PR #425, opened 2023-10-08
Standardize type declarations to lowercase for front end containing d…

RengarajR404:
The above PR is for standardising the type declarations inside defineProps.

### DIFF
--- a/frontend/components/card/CardMetric.vue
+++ b/frontend/components/card/CardMetric.vue
@@ -14,10 +14,10 @@
 
 <script setup lang="ts">
 defineProps<{
-  text: String;
-  number: String;
-  textColor: String;
-  borderColor: String;
-  backgroundColor: String;
+  text: string;
+  number: string;
+  textColor: string;
+  borderColor: string;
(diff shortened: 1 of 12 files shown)


andrewtavis:
Thank you @RengarajR404! I'll do a final review later today, but this looks to be what we're after. I appreciate the quick response to the other PR's review 🙏

andrewtavis (approved):
Thanks for this, @RengarajR404! 🙏 Will bring this in and make some minor changes to close the issue. I'm realizing that the way we're doing default values needs to change dramatically as we're using the JavaScript `String` etc, and changing that to `string` doesn't actually work. What we need is to use [props with defaults](https://vuejs.org/guide/typescript/composition-api.html#typing-component-props) :)

Will implement this quick and we'll close it out! 🚀

RengarajR404:
Feel free to contact me if you need anything.
Happy to help in a good cause.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Issue #400, opened 2023-09-22
Create sign in/sign up tooltip (or general tooltip with variants)


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to create a component `TooltipSignIn.vue` that would prompt the user that they need to sign in or sign up. The designs for this can be found [on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2037%3A3102&mode=design&t=QtIT1amRX0lQy9zG-1) and can further be seen below:

![Mode=Light, Type=Sign in](https://github.com/activist-org/activist/assets/24387426/bb493b18-c2da-4efe-a0ab-99658f504a4f)

This component would comprise of a text, two `BtnLabeled` components, would need a shodow and would further be reactive to light and dark mode 😊 What would be even better would be if we could create a base tooltip component that then can be changed based on a passed text and buttons to a `<slot>`, which I'd be happy to discuss with the potential contributor!

### Contribution

This is a great `good first issue` for someone who'd like to make a first component for us, and can also be picked up by a member of the community if they're looking for something to work on 😊 Happy to support someone who has interest in working on this! This can also be worked on as a part of our 23/9/23 Code Night 🌙🚀

marcosquadross:
I would like to work on this issue.

marcosquadross:
I would like to work on this issue.

andrewtavis:
Sounds good, @marcosquadross! Thanks for your interest in activist and your willingness to help 😊 Would you have interest in the base component as well? By that I mean the part I added here:

> What would be even better would be if we could create a base tooltip component that then can be changed based on a passed text and buttons to a `<slot>`

For that we'd make to components, `Tooltip.vue` and `TooltipSignIn.vue` where the latter would just have the appropriate buttons added to a `<slot>` as I described :)

Another tip on this: `card-style` can be used to apply the backgrounds and borders 🙃

Assigning you, and let me know if you want to do `Tooltip.vue` too :) Happy to provide more assistance as well if needed!

hoon610:
Hi! If @marcosquadross is specifically working on the `TooltipSignIn.vue` could i work on a base component?

[...]

PR #428, opened 2023-10-08
Add BaseTooltip component and update it to the search bar

hoon610:
### Contributor checklist

---
Screenshot 2023-10-08 150810
Screenshot 2023-10-08 150841

### Description

Add a BaseTooltip component.

Tested by adding into the search bar component.  I kept the styling classes invisible and mt-8 in the invocation of BaseTooltip in the SearchBar component, since I assume those styling features are specific to its location. I didn't mess with the control and command k tooltips that are commented out because I do not have a mac to test them on.

All other styling fits the designs on Figma except the text-color for dark mode. I could not find the exact color in the stylesheet. I removed the 'Header' placeholder before commiting the changes. Let me know if you need me to change anything or have any concerns!

### Related issue

- #400

### DIFF
--- a/frontend/components/SearchBar.vue
+++ b/frontend/components/SearchBar.vue
@@ -27,11 +27,10 @@
         <div
           class="flex px-2 py-[0.125rem] text-sm text-center rounded-md has-tooltip bg-light-highlight dark:bg-dark-highlight text-light-special-text dark:text-dark-special-text"
         >
-          <span
-            class="invisible px-2 py-1 -mt-8 rounded shadow-md shadow-zinc-700 bg-light-menu-selection dark:bg-dark-menu-selection w-max text-light-content dark:text-dark-content tooltip"
-          >
-            {{ $t("components.search-bar.slash-tooltip-label") }}
-          </span>
+        <BaseTooltip 
+        class="invisible -mt-8"
+        :text="$t('components.search-bar.slash-tooltip-label')"
+        />
           <p class="-mt-[0.075rem]">/</p>
         </div>
         <!-- <div
@@ -80,6 +79,7 @@
 <script setup lang="ts">
 import { useMagicKeys, whenever } from "@vueuse/core";
 import { ref } from "vue";
+import BaseTooltip from "./tooltip/BaseTooltip.vue"
 
 const route = useRoute();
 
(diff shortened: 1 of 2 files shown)


andrewtavis:
Thanks for the PR here, @hoon610! No stress on the TS error - that was my bad from a recent commit 😊 I'll try to get to this soon!

andrewtavis (approved):
Looks really great, @hoon610! I'm going to play around with it a bit and will likely decrease the text size, but overall this is super helpful and I really enjoyed discussing how this could work with you! 😊

andrewtavis:
Ah, one quick thing I forgot that I thought of at first glance a bit ago: I'll likely change the name to just `Tooltip` as we should be following the directory structure in naming. Maybe `TooltipBase` could also work, but might go for the simplest option :)

frontend/components/tooltip/BaseTooltip.vue
```diff
@@ -0,0 +1,21 @@
...
+        </slot>
+        <slot name="text">
+          <div class="responsive-h5 text-light-text dark:text-dark-text" v-if="text">{{ text }}</div>
+        </slot>
+        <slot name="button" class="block"></slot>
+    </div>
+  </template>
+  
+  <script setup lang="ts">
+  import { defineProps } from 'vue';
```
andrewtavis, line 14: As we're in Nuxt we don't actually need to do this import :)

frontend/components/tooltip/BaseTooltip.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+    <div class="flex flex-col responsive-px-2 responsive-py-1 rounded shadow-md shadow-zinc-700 bg-light-distinct dark:bg-dark-distinct w-max text-light-content dark:text-dark-content tooltip">
+        <slot name="header">
```
andrewtavis, line 3: These extra slots are also not needed, but no stress. Easy to take them out :) :)

frontend/components/SearchBar.vue
```diff
@@ -80,6 +79,7 @@
 <script setup lang="ts">
 import { useMagicKeys, whenever } from "@vueuse/core";
 import { ref } from "vue";
+import BaseTooltip from "./tooltip/BaseTooltip.vue"
```
andrewtavis, line 82: Similarly here - no need for the import within Nuxt. We need to import from extensions, but all components are auto loaded 🚀

hoon610:
Thanks for your help! I really enjoyed working with you as well!

PR #430, opened 2023-10-09
mypy workflow setup

to-sta:
### Contributor checklist

---

### Description

In this PR i have added mypy (Static Type Checking for python) to the backend workflow. This is the last missing piece for the issue #329. 

I tested the cli command locally and then run the workflow via [act](https://github.com/nektos/act).

<ins>Comments:</ins>
While looking into the workflows,  i was thinking that we should rename the files now.  pr_backend_formatting_check does linting and static type checking. I think pr_ci_backend would make sense. I will open an issue for that and a few more minor things regarding workflows.

### Related issue

- #329

### DIFF
--- a/.github/workflows/pr_backend_formatting_check.yaml
+++ b/.github/workflows/pr_backend_formatting_check.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Thanks for the edits here @to-sta as well as the new issues to keep the setup improvements rolling! 🚚😊

PR #438, opened 2023-10-10
Enforce order of Vue SFC

ahmedy00:
…nd custom rules

### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -1,0 +1,13 @@
+module.exports = {
+  extends: ["eslint:recommended", "plugin:vue/vue3-recommended"],
+  rules: {
+    "vue/component-tags-order": [
+      "error",
+      {
+        order: ["template", "script", "style"],
+      },
+    ],
+    "vue/multi-word-component-names": "off",
+    "vue/no-empty-component-block": "error",
+  },
+};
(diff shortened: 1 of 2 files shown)


frontend/.eslintrc.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
```
ahmedy00, line 1: @andrewtavis On my IDE(pycharm) the word `module `gives the following error: 
ESLint: 'module' is not defined.(no-undef). Lets check with another IDE and if error still exist we can add the following code to top of file: 
/*eslint-disable no-undef*/
or we can discuss about better solutions

andrewtavis:
Thank you @ahmedy00! Get some sleep good sir! :)

frontend/.eslintrc.js
```diff
@@ -0,0 +1,13 @@
...
+  extends: ["eslint:recommended", "plugin:vue/vue3-recommended"],
```
ahmedy00, line 2: I have included the rules suggested by eslint in the project to avoid problems, we can check the recommended rules with a 'tick' next to them on the [rules page](https://eslint.org/docs/latest/rules/) and we can discuss possible problems

andrewtavis (approved):
Thank you for looking into this, @ahmedy00! Really happy to add this to our already strong dev env 💪 Also thanks for the extra options added in here. Both make sense!

Issue #445, opened 2023-10-13
Eslint rules does not gives warning for VS Code


ahmedy00:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

The new rules added in issue #424 work properly in Jetbrains IDEs (Pycharm, Intellij IDEA), but do not show the necessary warnings in VS Code.

ahmedy00:
@andrewtavis Could you assign it to me?

andrewtavis:
Sure thing, @ahmedy00! Appreciate you opening this up and looking into this further! 🙏

ahmedy00:
Hey @to-sta, when you have time I would like to discuss about this bug and benefit from your experience

to-sta:
@ahmedy00 i looked a bit into the issue and found out that when i add the following into the `.vscode/settings.json` file:

```json
{
  "eslint.validate": []
}
```
That vscode is starting to complain about the order. This is not an optimol solution though. 

@ahmedy00 are you on matrix?

@andrewtavis At the moment we run `formatting` and `type` checks on the frontend, might make sense to add a linter as well (to gain independence of the IDE issues 😄).

andrewtavis:
Makes sense, @to-sta! Would you make an issue for that? :)

andrewtavis:
Closed via #448 and #452 🚀 Thanks for this, @ahmedy00 and @to-sta!

PR #448, opened 2023-10-14
BugFix: Eslint rules does not gives warning for VS Code

ahmedy00:
### Contributor checklist

---

### Related issue

- #445

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "eslint.validate": [
+    "javascript",
+    "javascriptreact",
+    "vue"
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
   "eslint.validate": [
     "javascript",
-    "javascriptreact",
+    "typescript",
     "vue"
   ]
 }
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -3,7 +3,7 @@ module.exports = {
   extends: [
     "eslint:recommended",
     "plugin:vue/vue3-recommended",
-    "plugin: vue/base",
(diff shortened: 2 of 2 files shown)


andrewtavis:
Hey @ahmedy00 👋 Thanks for this! 😊 Don't forget to fill out your [PR description](https://github.com/activist-org/activist/pull/448#issue-1943136825) as this can help others who might want to review and aren't exactly sure what the context is :) More of a problem for the future, but good to get into the practice now 😇

andrewtavis:
This solves the issue that the error isn't showing up, @ahmedy00 😊 Issue is that the dev env really starts screaming issues at me including the `module.exports` one in `eslintrc`. Are you still getting this error? I feel like this is likely a Nuxt issue? Should we try [nuxt/eslint-config](https://github.com/nuxt/eslint-config/tree/main)? If so, would we need `eslint-plugin-vue`, or would that be optional potentially?

ahmedy00:
> This solves the issue that the block order error isn't showing up, @ahmedy00 😊 Issue is that the dev env really starts screaming issues at me including the `module.exports` one in `eslintrc`. Are you still getting this error? I basically have an error in every `.vue` file when I check this out.
> 
> I feel like this is likely a Nuxt issue? Should we try [nuxt/eslint-config](https://github.com/nuxt/eslint-config/tree/main)? If so, would we need `eslint-plugin-vue`, or would that be optional potentially?

andrewtavis:
Let me think on this @ahmedy00 and see if adding the Nuxt eslint plugin would solve these errors :)

andrewtavis (approved):
Hey @ahmedy00, we'll be merging in #452 right after this that sends along new configuration files and fixes a ton of the errors we were seeing 😊 Thanks for the initial work on this! 🚀

Issue #446, opened 2023-10-13
GitHub Workflow for eslint


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

General:
On the backend side of things we have the whole suite already: formatter, linter and static type checking. It is also configured to work well together. 

On the frontend side we are still missing the linter. 

The task is to add an `eslint` step to the `pr_ci_frontend` workflow. In case this or any other step fails in the workflow it should continue to run.

Prerequisite is the merge of PR #444 and solving the issue #445

### Contribution

I am happy to help someone or work on this myself :smile

to-sta:
I added another issue #449 about configuring eslint. This should be resolved before we start merging a PR with a new workflow.

andrewtavis:
Closed by #466 🚀 Let me know if you think anything else is needed as far as workflows, @to-sta!

PR #466, opened 2023-10-16
eslint workflow added and contine-on-error added

to-sta:
### Contributor checklist

---

### Description

In this PR, I added the eslint to the frontend check. I also realized that we have added continue-on-error to the frontend check but not to the backend check. That can be seen in the PR #465, where ruff fails due to import sorting and mypy will not be executed anymore.

I tested ran the command in the frontend docker container.

### Related issue

- #446

### DIFF
--- a/.github/workflows/pr_ci_backend.yaml
+++ b/.github/workflows/pr_ci_backend.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   backend:
-    name: Run PR Backend Formatting Check
+    name: Run PR Backend Check
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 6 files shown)


andrewtavis (approved):
Thank you, @to-sta! I was about to send along the formatting changes myself 😊 Those changes came from my autoformat I'm pretty sure. Do I need to setup my code formatter to be ruff?

Appreciate the ESLint workflow as well :) Will close the issue, and let me know if you think there's anything else as far as workflows that we need to look into!

to-sta:
> Thank you, @to-sta! I was about to send along the formatting changes myself 😊 Those changes came from my autoformat I'm pretty sure. Do I need to setup my code formatter to be ruff?

If you the pre-commit hooks installed then you do not need to worry about anything 😉. Will be taking care off before its submitted :). But the VScode extension can also help :). 

You can install pre-commit globally with pip install pre-commit or conda if its available there, and install the hooks with pre-commit install. That's it.

Issue #471, opened 2023-10-17
Swagger UI endpoint is broken


to-sta:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

While working on #450 I discovered that the `/v1/schema/swagger-ui/` does not work anymore. I went thru the commit history to find out what cause the issue and figured out that its caused by:

```python
import django_stubs_ex
django_stubs_ext.monkeypatch(extra_classes=(viewsets.ModelViewSet,))
```

It somehow override the DEFAULT_SCHEMA_CLASS that we are setting in the REST_FRAMEWORK dictionary.

```bash
>>> from rest_framework.settings import api_settings
>>> api_settings.DEFAULT_SCHEMA_CLASS
<class 'rest_framework.schemas.openapi.AutoSchema'>
```

When I remove django_stubs_ex, than we are getting the right result.
```bash
>>> from rest_framework.settings import api_settings
>>> api_settings.DEFAULT_SCHEMA_CLASS
<class 'drf_spectacular.openapi.AutoSchema'>
```

I found a open issue on [djang-stubs](https://github.com/typeddjango/django-stubs/issues/1750) with the same problem. Any ideas on how we could solve that? 

We could wait and see if the issue gets resolved or in case we need SwaggerUI soon, we could adjust mypy. We could change the settings on generics and remove the monekypatch(). 

We could also discover issues like this with testing our endpoints. The page currently returns a HTTP 500 status code.
```bash
django_backend  | [17/Oct/2023 20:52:22] "GET /v1/schema/ HTTP/1.1" 500 108779
```

to-sta:
I found a solution and will make PR right away. Let's keep the issue open till this [djangorestframework-stubs issue](https://github.com/typeddjango/djangorestframework-stubs/issues/299#issuecomment-1343952084) gets resolved.

andrewtavis:
Makes sense to leave this open, @to-sta :) PR is merged and we can keep an eye on this 🙃

PR #474, opened 2023-10-18
reload drf api settings to solve Swagger UI bug

to-sta:
### Contributor checklist

---

### Description

This PR solves the bug that breaks the Swagger UI endpoint `/v1/schema/swagger-ui/`. This can be considered  a workaround, since we are reloading the drf api_settings after calling the monkeypatch function.  

There is currently a open issue on djangorestframework-stubs that is related to this bug. Let's keep the issue #471 open to keep track of the changes.

I have tested the endpoint to make sure it works.

### Related issue

- #471


andrewtavis (approved):
Thanks for this, @to-sta!

Issue #463, opened 2023-10-15
Create CardDiscussionEntry component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to create a component that would be used to display the various discussions for an organization and on other parts of the platform. The component would be made within the [components/card](https://github.com/activist-org/activist/tree/main/frontend/components/card) directory, with its designs being found [here on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1955%3A1111&mode=design&t=jTUMfp3ZcqY6KqEK-1). The light mode version of this component can further be seen below:

![Mode=Light, Mobile=False, Type=Discussion, Reduced=False](https://github.com/activist-org/activist/assets/24387426/65627e1d-32a9-4a8a-a480-7bbc716abe70)

Generally the design is similar to GitHub issues, but is a card rather than a list item :)

### Contribution

Happy to support someone who wants to work on this! This is a `good first issue` that would be ok for someone new to the community to pick up, and is also of course welcome for the current dev community if someone's looking for something to work on 😊

andrewtavis:
Sure thing, @GOWTHAM2K1 :) I'd suggest that you check the search result components within the card directory for this one 😊 Let me know if you need some help!

andrewtavis:
Ok, @GOWTHAM2K1 😊 Please give the changes in 004bfb2 a look through, but this is now closed :) There wasn't a mobile version of this in the designs, so I went ahead and formatted it to what I figure is ok for now. From there we're good to proceed on the PR for #460 🚀

Issue #460, opened 2023-10-15
Create CardDiscussionInput component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create a component that houses the write a discussion comment UI. It roughly matched the GitHub comment UI, and can be found [here on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2642%3A19615&mode=design&t=jTUMfp3ZcqY6KqEK-1) as well as below:

![Discussion Input](https://github.com/activist-org/activist/assets/24387426/c12dea1a-61c9-4c99-a069-30d4b83cca29)

This issue doesn't necessarily need to include the full functionality of this component! We can certainly separate it into other issues after an initial implementation is created here 😊

### Contribution

This issue is more involved than other issues and should likely be worked on by someone who has knowledge of the code base. Happy to support a new contributor though! Also happy to pick it up myself at some point 😊

GOWTHAM2K1:
hey can i do it??

andrewtavis:
I'm happy to assign you, @GOWTHAM2K1, but before you start working on it can we document the steps we'll take to make this component? This is definitely more involved than other issues and is definitely not easy. How would you go about working on this?

andrewtavis:
Assigned :) Let's please make a plan before coding starts though so we can agree on how best to build this 😊

GOWTHAM2K1:
> I'm happy to assign you, @GOWTHAM2K1, but before you start working on it can we document the steps we'll take to make this component? This is definitely more involved than other issues and is definitely not easy. How would you go about working on this?

So I would firstly build the component as per design and then individually start embedding functionality to the elements in the components Individually.  And also I am a beginner in vue framework so i would like to know the best way to go froward .

andrewtavis:
Maybe it makes sense that for this we just build the component with all the buttons and then we can add the functionality in other issues :) How does this sound, @GOWTHAM2K1?

GOWTHAM2K1:
Yes, it would make the process much quicker.I will submit pr as soon as possible.

andrewtavis:
[...]

PR #484, opened 2023-10-20
card discussion component

GOWTHAM2K1:
---

---

### Description

created card discussion component as closely as possible to the design . Please let me know if anything needs to be changed. And thanks for your guidance for hacktoberfest.

- #ISSUE_NUMBER
#463

### DIFF
--- a/frontend/components/card/search-result/CardDiscussionEntry.vue
+++ b/frontend/components/card/search-result/CardDiscussionEntry.vue
@@ -1,0 +1,59 @@
+<template>
+    <div class="flex flex-col w-full md:flex-row">
+        <div class="justify-content-centre w-min md:w-min">
+            <BtnLabeled class="w-small flex mt-1" :cta="true" linkTo="/" :label="discussion.supporters" fontSize="sm"
+                leftIcon="bi:arrow-up" iconSize="1.25em"
+                ariaLabel="components.btn-labeled.support-organization-aria-label" />
+        </div>
+        <div class="flex-col pt-3 space-y-3 md:pl-4 lg:pl-6 md:space-y-4 md:grow md:pt-0">
+            <div class="flex flex-col justify-between md:flex-row">
+                <div class="flex items-center justify-center space-x-2 md:space-x-4">
+                    <h2 class="font-bold responsive-h3">
+                        {{ discussion.name }}
+                    </h2>
+                    <div class="w-min md:w-min">
+                        <BtnLabeled class="w-small flex mt-1" :cta="true" linkTo="/" :label="discussion.category"
+                            fontSize="sm"  iconSize="1.25em"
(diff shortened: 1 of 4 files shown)


andrewtavis:
Thank you @GOWTHAM2K1! I'll review this soon :) :)

frontend/components/card/search-result/CardDiscussionEntry.vue
```diff
@@ -0,0 +1,59 @@
...
+                            class="border rounded-full w-fit border-light-section-div dark:border-dark-section-div bg-light-content dark:bg-dark-content">
+                            <Icon name="bi:person" size="1.75em" />
+                        </div>
+                        <p class="ml-2">Author</p>
+                    </a>
+                    <div class="ml-2 flex items-center">
+                        <MetaTagDate :date="event?.date.toLocaleDateString()" />
+                        <p class="mt-0.8 ml-1">Date and Time</p>
+                         
+                    </div>
```
andrewtavis, line 41: Looking at the file here and the overall coloration of the elements, @GOWTHAM2K1, it looks like you're missing a tag in here somewhere. You see how the `<script>` tag also doesn't have syntax coloring. My expectation is that the missing tag is being added by Vue/Nuxt/the browser. I'll go through and find where it's missing 🙃

frontend/components/card/search-result/CardSearchResult.vue
```diff
@@ -27,6 +27,12 @@
     >
       <CardSearchResultUser :user="user" :isPrivate="isPrivate" />
     </div>
+    <div
```
andrewtavis, line 30: I'm going to take this part out of here as well. This component is supposed to be `CardDiscussionEntry` based on the issue, which I get was confusing from the designs. It's styled like a search result, but we should keep all of this and what you're working on in #460 in a `discussion` directory as that to me seems like a more fitting categorization 😊

frontend/components/card/search-result/CardSearchResult.vue
```diff
@@ -35,13 +41,16 @@ import type { Event } from "../../../types/event";
 import type { Organization } from "../../../types/organization";
 import type { Resource } from "../../../types/resource";
 import type { User } from "../../../types/user";
+import type { Discussion } from "../../../types/discussion";
+import CardDiscussionEntry from "./CardDiscussionEntry.vue";
```
andrewtavis, line 45: A quick note here that as with the other issue we do not need to import components because Nuxt imports them automatically, but we do need the type :) Obviously no worries as these reviews are coming so close to one another!

andrewtavis (approved):
This is looking very good, @GOWTHAM2K1! Big thing is just renaming everything and getting it independent from the search result one, but no stress 😊 I'll get that done shortly!

Issue #260, opened 2023-08-12
Fix TS errors in components/form directory


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue is related to #211 in which we've been documenting the current TS errors that activist is facing. As of now the [components/form](https://github.com/activist-org/activist/tree/main/frontend/components/form) directory has the following error output when running `yarn nuxi typecheck` in the `frontend` directory:

```
Errors  Files
     4  components/form/FormButtonStyledCheckbox.vue:14
     3  components/form/FormButtonStyledRadio.vue:16
     3  components/form/FormCheckbox.vue:7
     6  components/form/FormCheckboxGroup.vue:9
     3  components/form/FormRadio.vue:5
     6  components/form/FormRadioGroup.vue:6
     3  components/form/FormSearch.vue:8
```

Would be great if we could fix the TS errors in each of these files :)

andrewtavis:
CC @TeddyGavi, @luuu-xu and @Mradosi 😊

Mradosi:
> CC @TeddyGavi, @luuu-xu and @Mradosi 😊

Ok, I will handle it :)

andrewtavis:
Thank you, @Mradosi! Hope you've been having a nice weekend so far :) :)

to-sta:
I happy to help here with the simply ts errors 😄. There is a recurring  error caused by :id="uuid"

```bash
components/form/FormButtonStyledCheckbox.vue:14:7 - error TS2322: Type 'number' is not assignable to type 'string'.
14       :id="uuid"
```
`id` is expecting a `string` but `uuid` is a `number`. We could solve this by converting the UUID into a string. 

```ts
let UUID = 0;

export default function useUniqueID() {
  const getID = () => {
    UUID++;
    return UUID;
  };

  return {
    getID,
  };
}
```

andrewtavis:
Thanks @to-sta! @Mradosi, do you have an estimate on this? Would be good to get this closed by Saturday so we can have the workflows being fully functional for the Code Night contributions :)

andrewtavis:
[...]

PR #439, opened 2023-10-10
Replace useUniqueId for uuid package

rigojr:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

- Remove `UseUniqueID` composable for [uuid](https://www.npmjs.com/package/uuid) package.
- When removing `UseUniqueID` composable, a minor memory leak inserted by the use of global variable is deleted `let UUID = 0`.

### Related issue

- No formal issue related, it was talk here https://github.com/activist-org/activist/issues/260#issuecomment-1732129561

PR #487, opened 2023-10-21
chore: move uuid to dependecy

rigojr:
### Contributor checklist

---

### Description

- Move `uuid` package to `dependencies`

### Related issue

- related issue https://github.com/activist-org/activist/issues/260#issuecomment-1732129561

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -27,7 +27,6 @@
     "nuxt": "npm:nuxt3@latest",
     "nuxt-icon": "^0.2.6",
     "typescript": "^5.1.3",
-    "uuid": "^9.0.1",
     "vue-tsc": "^1.8.8"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


rigojr:
Hi @andrewtavis  just to let you know that I take a really deep dive in the type issue reported here https://github.com/activist-org/activist/pull/439#issuecomment-1758527082 and I really could not find any issue.

On my end:
image
image

I don't think this is a "It works on my machine ¯\_(ツ)_/¯" problem, the `@types/uuid` should do the work and it seems that is working (on my machine 😆). To be in the same page, these are my current setups:

image
image

And my current VS extensions are the suggested in the documentation plus others, so I don't think this is a VS extension problem, let's get it deep on this issue through Matrix.

andrewtavis (approved):
Thanks for the change here, @rigojr! I agree that this seems to be something better left for a call. Let's continue to discuss on Matrix and we'll figure it out!

Issue #489, opened 2023-10-22
CI workflows are running on main, not branch


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Something that’s been noticed by myself and other contributors is that currently the workflows for the repository are running on the current main branch of the code and not the code of the branch for PRs. This is problematic as we’re getting passed tests that then lead to merged code that then leads to failed tests. Looking through the CI process and deriving what’s causing this is a high priority to assure that no new errors are being introduced :)

andrewtavis:
Ping @rigojr :) Let me know if you have some ideas on this, and thanks for reporting! Also ping @to-sta who might have some ideas 😊

to-sta:
In order to run on the pr branch we need to switch to pull_request but then we lose access to GitHub secrets and variables (for security reasons). 

In order to use pull_request we have to supply env. variables in another way. I will have a look into this.

andrewtavis:
There's definitely more work to be done on testing, but I feel like this issue at the very least is closed at this point. We can reopen if there are further problems, and new issues for anything that's needed for mypy would be welcome 😊

PR #486, opened 2023-10-21
Card discussion component

GOWTHAM2K1:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

Completed card discussion input component

- #ISSUE_NUMBER
- #460

PR #490, opened 2023-10-22
fix(ci): correct event type "pull_request"

wkyoshida:
### Contributor checklist

---

### Description

If I'm understanding correctly, I think the trigger event type that we would want is `pull_request` rather than `pull_request_target`. 

From the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target):

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the `pull_request` event does. 

### Related issue

- #489

### DIFF
--- a/.github/workflows/pr_ci_backend.yaml
+++ b/.github/workflows/pr_ci_backend.yaml
@@ -1,6 +1,6 @@
 name: pr_ci_backend
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - main
     types: [opened, reopened, synchronize]
--- a/.github/workflows/pr_ci_frontend.yaml
+++ b/.github/workflows/pr_ci_frontend.yaml
@@ -1,6 +1,6 @@
 name: pr_ci_frontend
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - main
     types: [opened, reopened, synchronize]


to-sta:
@wkyoshida when we switch to `pull_request`, we lose access to GitHub secrets.

We might have to adjust the workflow for this to work. I have to look into this.

to-sta:
We can add environment variables to the workflow and then we can switch to use `pull_request`. 

```
env:
  SECRET_KEY = 'Add a secret key for the django backend here'
  DEBUG = True
  ....

```
Might make sense to place a comment in there "for testing purpose only".

wkyoshida:
Right :thinking: 
@to-sta sorry, remind me again the cases of how we were thinking of using secrets in the workflows?

to-sta:
> Right 🤔 @to-sta sorry, remind me again the cases of how we were thinking of using secrets in the workflows?

We are actually do not use any github variables or secrets in any of the workflows. I think the always had `pull_request_target` in workflows there? @andrewtavis.
I did not question that 😄.  

In order to run mypy we need the enviroment variables SECRET_KEY and DEBUG. But i just saw that we actually have a setup that defaults to DEBUG=False and generates a secret key inside settings.py everytime we run it. 

We can just switch to pull_request :).

to-sta (approved):

to-sta:
After thinking about this for a bit, I would suggest to remove the default value for DEBUG and the SECRET actually, because this could cover up errors in our setup.

```python
DEBUG = int(os.environ.get("DEBUG", default=0))
SECRET_KEY = get_random_secret_key()
```

And since assign SECRET_KEY twice in settings.py we overwrite it. 

@wkyoshida what you think?

wkyoshida:
Oh I see, gotchu! Just saw the `mypy.ini` file :+1: 

I'll be honest though - I haven't used `mypy` before, so I'm a bit unclear still how the env vars get used in the static type checks. Although, it does appear that the `mypy` step in the workflow is having issues working with env vars as seen in the [logs](https://github.com/activist-org/activist/actions/runs/6601379889/job/17932356293?pr=490#step:7:51). Noticed this same issue in most of the other open PRs and all of them erroring on an empty `DATABASE_NAME` env var (as seen in the log). Leads me to believe there might be a different issue as well going on with `mypy`.. :disappointed:  

Would it make sense for us to get this PR merged so we can at least get folks now running their CI on the correct commits? Then also open an issue to investigate why `mypy` is having issues with env vars? I can open it and assign to you (I assume that would make sense as you've been working with it so far, but let me know otherwise)

to-sta:
I agree we should merge this PR. The issue with mypy is already adressed [here](https://github.com/activist-org/activist/pull/486#issuecomment-1774063105)

andrewtavis (approved):
Thank you both for the discussion here! 🙏 Glad to have this covered so we're not getting blindsided from now on :)

wkyoshida:
And @to-sta glad you already figured out the mypy issue!! :muscle::rocket:

Issue #491, opened 2023-10-22
Backend fixture for superuser


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We also have to create a superuser for the django backend in order to access the django admin dashboard. 

Task:
Create a superuser fixture and adjust dockerfile to read the fixture everytime. 

### Contribution

I will work on this myself.

andrewtavis:
Closed via #492. Thank you, @to-sta!

PR #492, opened 2023-10-22
fixture superuser added / load fixture in docker

to-sta:
### Contributor checklist

---

### Description

This PR introduces a fixture for a superuser in the Django backend. The superuser is read from the docker-compose file, enabling access to the Django admin panel without the need to execute` python manage.py createsuperuser` each time.

```
user: admin
pw: admin
```

This will improves the developer experience. 

I executed docker-compose up --build to ensure the changes are valid and the user and pw combination works.

### Related issue

- #491

### DIFF
--- a/backend/fixtures/superuser.json
+++ b/backend/fixtures/superuser.json
@@ -1,0 +1,33 @@
+[
+  {
+    "model": "authentication.user",
+    "pk": "83f1174f-1f22-46c8-a542-f4814d97bbf3",
+    "fields": {
+      "last_login": "2023-10-22T11:58:17.946Z",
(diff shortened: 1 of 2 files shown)


rigojr:
Howdy dummy question, @to-sta how do you discriminate this configuration from production environment?

to-sta:
> Howdy dummy question, @to-sta how do you discriminate this configuration from production environment?

At the moment there is only a development environment. But as soon as we have a prod environment, we could split the docker-compose file into prod and dev.

andrewtavis (approved):
Thanks for keeping this in mind and getting it done, @to-sta! 🙌😊

Issue #456, opened 2023-10-15
Create FormViewSelector component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to create a view selector component within the [form directory](https://github.com/activist-org/activist/tree/main/frontend/components/form). The designs for this component can be found [here in the Figma designs](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1922%3A1810&mode=design&t=jTUMfp3ZcqY6KqEK-1), and is also found below:

![View Selector](https://github.com/activist-org/activist/assets/24387426/2a19a41d-f76b-47ed-9430-0da237bf8620)

How the props for this component would be defined is still up for discussion, and ideas would be welcome!

### Contribution

Happy to support someone with this issue! This is a great `good first issue` for someone who wants to join the community and also for a contributor who's looking for something to work on 😊

bmartins-42:
Hello, if this issue is still available, I would gladly work on it :)

andrewtavis:
Hey @bmartins-42! Happy to assign you 😊 Please let me know if there's anything I can do to assist :)

bmartins-42:
Thank you, @andrewtavis.

Maybe you could explain me a little the purpose of this component, so I can understand the usage.

andrewtavis:
Sure thing, @bmartins-42! I'd say on this let's focus on the bottom ten selectors, with that being four options that have list or map view and six that have list, map or calendar view. This would basically be the filter that you would click in the sidebar to change the view of the right side of the screen. So say that we're on a list view of events and we click the map: we'd then switch to a map, and similarly for the calendar. The best place to see this is on [the events page on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2504%3A14117&mode=design&t=1iLaGoS7r2LOTewF-1). For filtering organization results we'd just need a list view and a map view. 

As far as props we can maybe pass selector options? `option1: string`, `option2: string`, and `option3?: string` where if the first option is null then we'd only have two buttons. We then determine the icons and the results based on what the passed option strings are like "list", "map" and "calendar" :)

How does this sound?

[...]

PR #495, opened 2023-10-24
Added FormViewSelector

bmartins-42:
### Contributor checklist

---

### Description

Added a `FormViewSelector` that is taking `v-model`, `option1`, `option2` as required and `option3` as optional

image image

### Related issue

- [#ISSUE_NUMBER](https://github.com/activist-org/activist/issues/456)

### DIFF
--- a/frontend/i18n/en-US.json
+++ b/frontend/i18n/en-US.json
@@ -204,6 +204,9 @@
     "loading": {
       "img-alt-text": "The activist icon: a lower case letter a with a half circle pointing down and to the left in the center."
     },
+    "marker-private": {
+      "private": "Private"
+    },
     "media-map": {
       "error-message": "There was a problem with the map service.",
       "img-alt-text": "Map displaying a pin at the location of this event.",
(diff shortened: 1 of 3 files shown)


bmartins-42:
- I do not have a lot of experience with A11Y so let me know if the compatibility has to be reworked.
- I also not sure about the color I used so let me no if I need to change them

andrewtavis:
Thanks for this, @bmartins-42! I'll try to get back to you with a review soon :)

frontend/i18n/en-US.json
```diff
@@ -342,8 +345,12 @@
       "contains-numbers": "Contains numbers",
       "contains-special-chars": "Contains special characters (!#$^*)"
     },
-    "marker-private": {
-      "private": "Private"
+    "view-selector": {
+      "title-aria-label": "Select a view type",
```
andrewtavis, line 349: This is really really wonderful, @bmartins-42 😊 Thanks for the focus on accessibility!

andrewtavis (approved):
Really great work here, @bmartins-42 😊 Just tried it out and it's perfect. Hope that the conversion over the Headless UI wasn't too annoying! Would be great to get your feedback on Headless. If you enjoyed using it, then there are a few more conversion issues that were just made, and it'd also be great to see if there are other components that we should be using it in!

Issue #432, opened 2023-10-09
Starting with testing


to-sta:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

# Let's start with testing!

We can use this issue to discuss **testing** in general and spin off **new issues** to work on.

![swiss_cheese_model](https://github.com/activist-org/activist/assets/49518857/86daece7-3894-484f-b341-f5668c12e252.png?raw=true)

Here is an illustration based on the [swiss-cheese-model](https://en.wikipedia.org/wiki/Swiss_cheese_model) of our current **development process** (excluding the testing barrier 😄). The image illustrates how bugs could get into our codebase and what barriers are in the way.

## Topics
1. Splitting development and production.
2. Testing
3. GitHub Workflows

## 1. Splitting development and production
That would involve splitting the `settings.py` file, for example, into `base.py`, `development.py`, and `production.py`, as well as separating the `requirements.txt`. We could use the `env. variable DEBUG` to switch between settings. The development settings should contain the location for the fixtures and a seperate test database. 

## 2. Testing
To start, I would say we should run `coverage.py` to find out where we need tests and then start implementing them. We could work on fixtures asynchronously. For end-to-end testing, we could use [playwright](https://playwright.dev/python/).

## 3. GitHub Workflows
After implementing tests, there are a few additions that can be made to improve the development process. First, we can add test coverage to determine where we need to add new tests. We also have to execute tests in the workflows to find issues that arise during changes. Additionally, I think we should also add a workflow for the build process to find dependency issues.

Since we are have not reached production, i would exclude logging and performance testing in general for now.
Let me know if i have missed something? Or if you have any suggestions. 

@andrewtavis @wkyoshida

wkyoshida:
> To start, I would say we should run `coverage.py` to find out where we need tests and then start implementing them.

Do we already have any? Not sure if I'm just grossly missing them :laughing:, but maybe safe to say that all-around tests needed?

> First, we can add test coverage to determine where we need to add new tests

+1 yep! This is great. What I've often done is have the workflow check if the coverage has gotten worse after the incoming changes. This could just be a warning in the PR or it could fail the check if we'd like to too :eyes::astonished:

to-sta:
[...]

PR #496, opened 2023-10-24
Test setup and first tests inside content app

to-sta:
### Contributor checklist

---

### Description
In this PR, I added initial tests for the `content` app. The tests check the __str__ method of the models. A base setup for coverage is added to `pyproject.toml`.

In factory.py are all the models to create objects for models that can be used for tests. [factory_boy](https://factoryboy.readthedocs.io/en/stable/) uses the [faker](https://faker.readthedocs.io/en/master/) package to generate fake names, words, and so on. That let's us create tests without the use of fixtures.

<ins>Issues:</ins>
* Currently there is no support for generic types with factory_boy, therefore I excluded the factory.py file and its imports from being checked by mypy.
* I excluded TopicFormat, since it has a ForeignKey to another app (events). There is only a placeholder `TopicFormatFactory` at the moment. We can go back to this after creating tests for events app or find another solution.

  ```python
  class TopicFormatFactory(factory.django.DjangoModelFactory):
      """Placeholder
      We do not have Factory for events.Format, therefore can not use a Subfactory yet.
      format_id = models.ForeignKey("events.Format", on_delete=models.CASCADE)
      """
      pass
  ```

<ins>Added development dependecies:</ins>
Are added to requirements-dev.in and compiled in requirements-dev.txt
* pytest
* pytest-cov
* pytest-django
* factory_boy

I have tested the factories within the django shell. I also ran pytest locally and with docker. Some tests require database access and won't work outside of docker. We could use a sqllite database for local testing in the future, when we split settings.py into dev and prod.

### Related issue

- #432

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,10 @@ __pycache__
 venv/
 .mypy_cache
 
+# Testing
+.coverage
+cov_html/
+
 # Nuxt dev/build outputs
 .output
 .nuxt
--- a/backend/content/factory.py
+++ b/backend/content/factory.py
@@ -1,0 +1,21 @@
+import datetime
+import random
+
+import factory
+
+from .models import Resource
+
+
+class ResourceFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = Resource
+
+    name = factory.Faker("name")
+    description = factory.Faker("paragraph")
+    topics = factory.Faker("words", nb=1)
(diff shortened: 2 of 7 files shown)


andrewtavis (approved):
All looks fine to me, @to-sta! Thanks for setting this all up 😊 I'll go ahead and unblock #505 now that this is merged :)

Issue #498, opened 2023-10-24
Convert CardLegalDisclaimer to Headless UI disclosure


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue focuses on converting over [CardLegalDisclaimer](https://github.com/activist-org/activist/blob/main/frontend/components/card/CardLegalDisclaimer.vue) so that the opened field functions using [Headless UI Disclosure component](https://headlessui.com/vue/disclosure). This will add a lot of great accessibility functionality to the component and would make it more in line with other components we have that use Headless UI.

### Contribution

This is a great `good first issue` for someone wanting some experience with Tailwind and Headless UI. Would be happy to support! It's also of course available to the dev community if someone's looking for a new issue 😊

bmartins-42:
Hello @andrewtavis 

Could you assign me this task

andrewtavis:
By all means, @bmartins-42! Please let me know if you need any assistance! Should be pretty standard for you though 😊

andrewtavis:
7a08617 finalizes this, @bmartins-42 😊 I thought it was best to have the whole thing as a button as before and to also allow the user to close it by clicking on the text. Also has the `focus-brand` now :) Seems to be working really well! Thanks for the support! 🚀

PR #504, opened 2023-10-26
chore: refactor component with HeadlessUI

bmartins-42:
### Contributor checklist

---

### Description

- The component now uses HeadlessUI to manage state and A11Y
- Fixed small issue with text in dark theme

### Related issue

- #498

### DIFF
--- a/frontend/components/card/CardLegalDisclaimer.vue
+++ b/frontend/components/card/CardLegalDisclaimer.vue
@@ -1,16 +1,17 @@
 <template>
-  <div
-    @click="toggleExpand()"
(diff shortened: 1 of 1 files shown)


bmartins-42:
About the `disclaimer` props should it be required ?

andrewtavis:
Thanks for this, @bmartins-42!

> About the `disclaimer` props should it be required ?

I'd say yes. If we don't have a disclaimer then we should have a general one :)

For the [PR comment](https://github.com/activist-org/activist/pull/504#issue-1963251719), I edited it to check the checkbox and also to show how we usually use the `- #ISSUE_NUMBER` - replace the text with the number in question and because it's a list item this will convert to the full name of the issue 😊

andrewtavis (approved):
Thanks @bmartins-42! I'll do a quick edit of this so that more styling is in the `DisclosureButton` such that when it's focussed it's not just the icon and text, but rather the full area :) Beyond that, I'm realizing that this should also have `focus-brand` applied to the Tailwind so that its focus colors are the correct ones :)

Issue #510, opened 2023-10-26
Firefox opens general page search on `/`, doesn't jump to activist search bar


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

For the search bar that's in the sidebar we have the hotkey `/` selected. An issue with this is that for Firefox there appears to be a default `/` hotkey to search the page. This leads to the user basically not being able to use this hotkey. 

I'm not really sure how we'd go about changing this so it would work, so suggestions would be very welcome!

esslamben:
From what I've seen most websites seem to use cmd + k/ ctrl + k as it was the default keybind to use the search bar in browsers. Its also what most doc websites use. https://react.dev/  https://docusaurus.io/ etc. How about we swap it to that as the default?

andrewtavis:
This is a great thought, @esslamben :) We've been thinking for a while that there will ultimately be two kinds of search on activist in the same way that there is on GitHub: `/` as a hotkey for the normal search within the general area of the platform where the user finds themself (in `/events` and search for events, etc), and `cmd + k` / `ctrl + k` as a jump to as it works on other websites. We haven't yet made the designs for this yet, but as with GitHub it'd be a modal where what's being searched for would then be pattern matched specifically based on the names of events or organizations.

In checking Firefox out a bit more, GitHub actually seems to have disabled the `/` default page search hotkey in favor of their own that jumps to the search bar as we'd like to. Maybe we could figure out a way to that as well?

esslamben:
so it looks like you can just do a event.preventDefault to prevent the behavior. Gonna try it now see if we can mimic what github does

andrewtavis:
Nice, @esslamben! I'll assign this to you then :) Feel free to send along a PR when you get it working 😊

andrewtavis:
Closed by #512 :) Thanks for the work here and the discussion, @esslamben! Realized that I haven't done the designs for the `cmd + k` / `ctrl + k` UI yet. I'll get to those soon, and maybe you can implement if that'd be of interest 😊

esslamben:
[...]

PR #512, opened 2023-10-26
Fixed bug with slash hotkey getting picked up by firefox

esslamben:
### Contributor checklist

---

### Description

- Used a custom event handler to hijack what happens when slash is pressed
- Added an if statement around input focus as it was throwing an error even prior to the above change

### Related issue

- #510


andrewtavis:
Thank you, @esslamben! I'll review this in the coming days :)

andrewtavis (approved):
This is really awesome, @esslamben 😊 Wasn't expecting this to be solved so quickly! Thanks for the help here!

Issue #429, opened 2023-10-09
Discussion (and potential setup) of ability to self-assign issues


wkyoshida:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

With the activist community growing and hacktoberfest in full-swing, I've been wondering about the ability for folks to potentially self-assign themselves to unassigned issues. Some reasons for this are: 
- to free @andrewtavis from having to diligently moderate issues
- to give contributors a hopefully smoother/enjoyable contributing experience
- to give other contributors better up-to-date visibility on which issues are still up-for-grabs

Not sure if this is deemed a beneficial automation - so open to discussion :smile::speech_balloon: One idea, though, could be simply for a quick GitHub Action to listen for GitHub events of type [`issue_comment`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment) and trigger after a comment containing something like `/assign me` or `/assign @wkyoshida` (something similar to what is discussed [here](https://github.com/orgs/community/discussions/25389)). Then, the Action could assign the user to the issue. 

Other thoughts - we could make it so:
- this only works for still unassigned issues
- only one user per issue is assigned this way. Other users can still get assigned to the issue through the regular manual flow (this is important, for instance, during discussion issues where multiple people are involved as in #174)

Let me know :shrug:

andrewtavis:
Definitely is an interesting question, @wkyoshida 🤔 If I'm being honest, I do kind of like the personal touch of assigning issues to non-collaborators so that we can do more of a personal welcome and make it clear that we're happy to help. The load on me will also decrease as the org has more members (😇🙃🙃). I think that we also should maybe expand it out a bit more as far as providing rights to contributors before we send org invites so they can pick up issues. This would certainly help? 

Let me know what you think!

wkyoshida:
> I do kind of like the personal touch of assigning issues to non-collaborators so that we can do more of a personal welcome and make it clear that we're happy to help

Oh absolutely - that is definitely fair! :smile: 

I guess - I am also just thinking that doing both is a possibility as well? The `/assign *` ability doesn't mean that core org members cannot also still make sure to hop in and write a welcome message + offer assistance :rocket: 

Simply - I'm thinking the big thing is just giving the contributor a different experience**:
[...]

PR #520, opened 2023-11-01
refac: Remove unnecessary triggers and steps in GitHub Actions

wkyoshida:
### Contributor checklist

---

### Description

Working on #429 , I noticed a couple points in our existing workflows:
- Was wondering if the `Checkout` step was necessary in the `pr_maintainer_checklist.yaml`, since the workflow doesn't perform any steps actually involving the source code. Tried removing this step and tested in my fork:
   - Test PR: https://github.com/wkyoshida/activist/pull/1
   - Notice in the [logs](https://github.com/wkyoshida/activist/actions/runs/6714205076/job/18247071592) that the workflow does not perform the `Checkout` step and still succeeds :rocket: 

- Also, the `workflow_run` triggers in the `pr_ci_backend.yaml` and the `pr_ci_frontend.yaml` appear to be redundant, since there is already a trigger for when the `pull_request` is `opened`. The `pr_maintainer_checklist` workflow is `completed` mere seconds after the `pull_request` is `opened`, so this simply causes an unnecessary duplicate CI run.
   - Notice the duplicate `pr_ci_backend` and `pr_ci_frontend` runs for the PR yesterday created for :herb:  
      image
   - Notice in the test PR from above, no duplicate CI runs occur :grin: 
      image

### Related issue

- #ISSUE_NUMBER ? 
   - Unsure... #429 ? As I was working on that when uncovered these items.. or another issue can be created. Either or

### DIFF
--- a/.github/workflows/pr_ci_backend.yaml
+++ b/.github/workflows/pr_ci_backend.yaml
@@ -4,10 +4,6 @@ on:
     branches:
       - main
     types: [opened, reopened, synchronize]
-  workflow_run:
-    workflows: [pr_maintainer_checklist]
-    types:
-      - completed
 
 jobs:
   backend:
--- a/.github/workflows/pr_ci_frontend.yaml
+++ b/.github/workflows/pr_ci_frontend.yaml
@@ -4,10 +4,6 @@ on:
     branches:
       - main
     types: [opened, reopened, synchronize]
-  workflow_run:
-    workflows: [pr_maintainer_checklist]
-    types:
-      - completed
 
 jobs:
   frontend:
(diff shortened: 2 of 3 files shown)


wkyoshida:
Likely not dire in our current situation, but thought of this refac to avoid unnecessary GitHub Actions usage, as there is in fact a limit on used minutes.

andrewtavis:
Thanks for the consideration on this, @wkyoshida! :) Great to cut it down as we can 😊

andrewtavis (approved):
Makes sense and looks good to me, @wkyoshida! If there's something like this on the Scribe end, could you also open PRs? Thanks! 🙏

Issue #530, opened 2023-11-04
GitHub Workflow for find_repeat_i18n_values.py


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Task is to create a github workflow to run `find_repeat_i18n_values.py'.

### Contribution

Will work on this issue myself

andrewtavis:
Glad my little script is getting a bigger role, @to-sta! 😊

andrewtavis:
Should we rename is from `worker`, @to-sta?

to-sta:
@andrewtavis I get rid of <del>`cronjob`</del> in the titel. 
Yeah renaming the workflow would make sense. Any ideas on how to rename this? (maybe i18n_check?)

Here my thoughts on improving the workflow/script:

I think there are two ways of improving this:
1. Raising an error in the script
2. Capturing the output of the script in the workflow

<ins>1. Raising an error in the script:</ins>
We can implement a logic in the script to raise an error based on the repeat values. 

<ins>2. Capturing the output of the script in the workflow:</ins>
This would allows to create an Issue/PR message with the output. The downside of this is that we potential create a lot of issues if we have consecutive commits.

andrewtavis:
Have it renamed locally now to `i18n_value_check`, @to-sta :) Let me send along the most recent changes shortly and we can talk about it further. I decided to alter the script so that it will let us know what the keys of the repeat values are and also suggest the lowest level of key we can use for the renamed key-value pair 😊 We could potentially capture the output and send it to the PR as a message. Making a bunch of good first issues would be nice, but then we need to make sure that repeats aren't being merged :)

andrewtavis:
40b7944 and fb04997 update the workflow and [find_repeat_i18n_values.py](https://github.com/activist-org/activist/blob/main/frontend/i18n/find_repeat_i18n_values.py). It's not perfect, but will give us good suggestions and also lists the keys that need to be find all replaced throughout. General workflow I'm envisioning is find all replace to a common key that will then create a key conflict in the JSON, and then just delete one of the two keys 😊

andrewtavis:
[...]

PR #531, opened 2023-11-04
Cronjob

to-sta:
### Contributor checklist

---

### Description
Cronjob workflow for the find repeat i18n values python script. I gave it a test run with act.

@andrewtavis 
The workflow does not fail. I think we could automatically create an issue with the result (I think that should be possible 😅) or have conditionals in the workflow. Let me know what you think would be useful, than we can add this functionally to the workflow.

The script is set to run at 6pm, every 14 days on a Monday.

- #530

### DIFF
--- a/.github/workflows/worker.yaml
+++ b/.github/workflows/worker.yaml
@@ -1,0 +1,31 @@
+name: Find Repeat i18n Values
+
+
+            # ┌───────────── minute (0 - 59)
+            # │ ┌───────────── hour (0 - 23)
+            # │ │ ┌───────────── day of the month (1 - 31)
+            # │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+            # │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+            # │ │ │ │ │
+            # │ │ │ │ │
+            # │ │ │ │ │
+            # * * * * *
+on:
+  schedule:
+    - cron: '0 18 */14 * 1' # At 00:00 on Monday
+
+jobs:
+  worker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo content
+        uses: actions/checkout@v4
+
+      - name: setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: execute py script
+        working-directory: ./frontend/i18n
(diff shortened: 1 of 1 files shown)


andrewtavis:
What about running this on PRs, @to-sta? Definitely opening an issue automatically would be cool!

andrewtavis:
The thing is, say there's a big change to the codebase that we merge and then i18n needs to do some work right after. If we wait for 10+ days the repeats would be potentially annoying multiple people. Sorry if I didn't get this fully in the sync (long day 😅). Let me know what you think!

to-sta:
> The thing is, say there's a big change to the codebase that we merge and then i18n needs to do some work right after. If we wait for 10+ days the repeats would be potentially annoying multiple people. Sorry if I didn't get this fully in the sync (long day 😅). Let me know what you think!

haha I thought its too much running this on every PR 😄.

I thought running this on every PR might be excessive 😄. I believe that if we run this on every PR, we might end up running this script too often without creating value.

What we could do instead is run this script only when certain files are changed. Here’s an example for .js files.

Here is an example on `.js` files.
```yaml
on:
  pull_request:
    paths:
      - '**.js'
```

We could target the [en-US.json](https://github.com/activist-org/activist/blob/main/frontend/i18n/en-US.json) file and run the script whenever it changes.

andrewtavis:
This is exactly what we should do, @to-sta! 😊 The magic of coding 🌈✨😇 Do you want to send along the change for that and we can merge it in? I'll be editing the JSON tomorrow or the next day and we can then see that it passes when they're all unique.

to-sta:
> This is exactly what we should do, @to-sta! 😊 The magic of coding 🌈✨😇 Do you want to send along the change for that and we can merge it in? I'll be editing the JSON tomorrow or the next day and we can then see that it passes when they're all unique.

Switched from a cronjob to targeting the file directly. I also added `push` besides `pull_request` to catch changes that you commit directly. 

I still want to improve this a bit. The workflow will not fail, because we are not raising an error inside the python file or have a condition in the workflow. So we have to check the workflow run to see what has changed. Automatically creating an issue would make sense here (but consecutive commits would create a lot of Issues 😆). 

Let's improve this in another Issue 😄.

andrewtavis (approved):
Really great, @to-sta! I'll try to send along the i18n updates soon so we can test this :)

Issue #354, opened 2023-09-07
Common menu state component for left sidebar and mobile dropdown


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We need a common component that houses the state of the menus between the lift sidebar and mobile dropdown such that it's stored in one place and can seamlessly translate between layouts. The left sidebar index logic is found in [SidebarLeftIndex.vue](https://github.com/activist-org/activist/blob/main/frontend/components/sidebar/left/SidebarLeftIndex.vue), and the dropdown is being worked on in #279.

### Contribution

Happy to work on this or support someone who has interest! 😊

andrewtavis:
Note that this state should also be shared with the component for #278 (if it's decided to be maintained).

andrewtavis:
Ping @KyryloKaralyus 😊

andrewtavis:
Sure thing, @Robert-Kovalcuk! Assigning you now :) Let me know if you have ideas for this!

andrewtavis:
Hey @Robert-Kovalcuk 👋 Thanks again for the call :) :) Quick point further in favor of moving the sidebar component directory into the menu component directory is that we actually have [MenuLinkWrapper](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuLinkWrapper.vue) being referenced by [MenuMobileSelector](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuMobileSelector.vue) for [MenuMobileNavBar](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuMobileNavBar.vue) as well as [SidebarLeftSelector](https://github.com/activist-org/activist/blob/main/frontend/components/sidebar/left/SidebarLeftSelector.vue). Good that we're doing that, but indicative that maybe we need to reorganize 😇

andrewtavis:
[...]

Issue #346, opened 2023-09-05
Update sidebar menu selection on page reload


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In #343 changes were made to [SidebarLeftIndex](https://github.com/activist-org/activist/blob/main/frontend/components/sidebar/left/SidebarLeftIndex.vue) that updates the `selected` attribute on a menu option based on the URL: i.e. on [activist.org/events/123/about](https://activist.org/events/123/about) `About` is highlighted. A final step is to change this so that it updates when a person switches to a different page. As of now if a person selects another sub page within the sidebar then the old highlight is maintained.

### Contribution

This is a `good first issue` for someone relatively new to the platform or a regular contributor who wants to pick up a task. Would be happy to work on this myself or support someone!

thiagogquinto:
Hey @andrewtavis, could I work on this issue?

contiyo:
Hi @andrewtavis 

I am mainly a guy who writes Python scripts for the last couple of years, but I would like to start contributing to open source projects and this issue seems to fit what I had in mind for my 1st issue. 

Can I start working on that please ? 

Thanks,
Kostas

andrewtavis:
Hey @contiyo 👋 Thanks for your interest in activist! Would be great to get some help with this issue 😊 I’ll assign you now, and please let us know if you need some assistance!

andrewtavis:
Hey @contiyo 😊 Checking in with you to see if there's anything I can do to assist with this issue :) Let me know!

Robert-Kovalcuk:
Hello, I'm looking for a an good-first-issue for my university class, is this still available?:)

contiyo:
Hi @andrewtavis, 

Apologies, but I coulnd't fix that myself. Feel free to pass that to @Robert-Kovalcuk.

andrewtavis:
No stress, @contiyo! Feel free to request another issue another time 😊 Sure @Robert-Kovalcuk, you'd be welcome to work on this! I'll assign you now, and please let me know if you have any questions!

andrewtavis:
[...]

PR #535, opened 2023-11-09
#354 #346 composable state for menu/sidebar components.

Robert-Kovalcuk:
### Contributor checklist

---

### Related issue

- #354
- #346


andrewtavis:
Hey @Robert-Kovalcuk 👋 Confusingly this isn't the spaces at the end of the file alone, but it's actually reverting the i18n changes as I'd feared. Not really sure what to do here 🤔 The version of the frontend you copied over has the old i18n keys, so because of that it's overwriting the new ones as you copied it into a fresh fork of the repo. Easiest would likely be to only copy the files that are for this issue over and then I can go through and fix the i18n keys in them if need be?

andrewtavis:
The no such file or directory for i18n is also because your fork has the old version/name of that file :)

Robert-Kovalcuk:
But I remember that I pulled the commit with the i18n while we were on a call. Hmm, let's figure it out tomorrow?

andrewtavis:
Ya it's very confusing. I remember you did that too. But ya, we can check it all tomorrow 😊 Thanks and till then!

andrewtavis:
Closing this due to new version in #539 :)

Issue #507, opened 2023-10-26
Create factories and tests for the authentication app in the backend


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The task is to create test data (factories) with [factory_boy](https://factoryboy.readthedocs.io/en/stable/) for the models in the authentication app. 

```bash
Name                            Stmts   Miss  Cover
---------------------------------------------------
...
authentication\models.py           49      6    88%
...
```
Have a look at the `content/factory.py` and `content/tests.py` to get started.

You can run `pytest` inside backend directory and a report terminal and html report will be generated. The html report shows you excatly what needs to be tested. The html report will saved in the folder "htmlcov".

### Contribution

Happy to help someone with this task.

IsaacDMz:
Hey, i would love to work on this task. Can assign this to me?

andrewtavis:
Thanks for your willingness to help, @IsaacDMz! Let us know if you need any support 😊

andrewtavis:
FYI @to-sta, I always mark `good first issue` with `help wanted` as there are some issue aggregators that parse for one or the other, but not both :) Obviously not always the other way around though 🙃 Thanks for making all these!

andrewtavis:
Closed by #540 :) Thanks @IsaacDMz! Really appreciate the work here 😊 Feel free to drop by the dev sync on Saturday if you have time. Details for that will be in `Development` on Matrix :)

PR #540, opened 2023-11-12
#540 Feat(Authentication) create tests and factories in authentication

IsaacDMz:
### Contributor checklist

---

### Related issue

- #507

### Note

I added a 'DEFAULT_PAGINATION_ORDERS_OBJECTS': False in the REST_FRAMEWORK settings to not require ordering on endpoints where order_by is optional

### DIFF
--- a/backend/authentication/factories.py
+++ b/backend/authentication/factories.py
@@ -1,57 +1,74 @@
 import factory
-from .models import SupportEntityType, Support, User, UserResource, UserTask, UserTopic
+
+from .models import Support, SupportEntityType, User, UserResource, UserTask, UserTopic
+
 
 class SupportEntityTypeFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = SupportEntityType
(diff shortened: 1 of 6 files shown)


to-sta:
@IsaacDMz thanks for factories :). Looks very good and do not worry about the ci_backend check. I will exclude the factories files from the mypy check later on. There are no stub files for factory_boy and its no that important. 

I made some minor adjustments. When you are using the string representation of another factory you need to be explicit with the file path. 

```python
topic_id = factory.SubFactory("content.factories.TopicFactory")
```

There is still an issue with UserFactory verification_partner field. It should use an instance of itself (UserFactory), but it works for build() at the moment. 

Will create a seperate issue on this later on.

andrewtavis (approved):
Thanks for the adjustments, @to-sta, and thanks @IsaacDMz for the initial work! We were really happy to see this when it came in. Great stuff 😊

Issue #499, opened 2023-10-24
Convert PageSubPageSelector to Headless UI tabs


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would focus on converting over [PageSubPageSelector](https://github.com/activist-org/activist/blob/main/frontend/components/page/PageSubPageSelector.vue) to use [Headless UI tabs](https://headlessui.com/vue/tabs) so that we can more easily access the options via the keyboard for increased accessibility. The styling should generally remain the same, but we'd just use Headless UI to make this component more in line with others that make use of Headless. 

### Contribution

This is a great `good first issue` for someone who wants to work with Tailwind and Headless UI! Would be happy to support, and members of the dev community are also of course welcome to work on this 🚀

andrewtavis:
@sayik, is this something you'd have interest in? Am realizing I screwed up the original scope here as we should be using Headless to make sure everything's accessible. Would be a good opportunity for you to check out this library and build on the past issues 😊 We can also do the backend one as well :) Just lemme know 🙃

nicholas-vecchio:
Hello! I am not sure if I am misunderstanding something or if the components have changed but the MenuTopBar component seems to not longer exist? Apologies if I'm simply confused.

andrewtavis:
Hey @nicholas-vecchio 👋 Sorry for the confusion. Had a call yesterday and realized that this component also wasn't particiularly well named. Updated the issue to reflect that we're talking about [PageSubPageSelector](https://github.com/activist-org/activist/blob/main/frontend/components/page/PageSubPageSelector.vue) 😊 Would you want to work on this?

nicholas-vecchio:
Yes I would @andrewtavis ! Thank you for the clarification.

andrewtavis:
Thanks for checking for clarification, @nicholas-vecchio! Assigned :) Please let me know if you need any support!

nicholas-vecchio:
Hello again @andrewtavis. Just for clarification, did you want to simply use the tabs for increased accessibility, and still have selecting the tabs route to a page link. Or would you simply like to use the TabPanels. Thank you again.

andrewtavis:
[...]

PR #541, opened 2023-11-14
Headless UI Tabs

nicholas-vecchio:
### Contributor checklist

- [☑️] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

Refactored the PageSubPageSelector to use HeadlessUI Tabs for increased accessibility. Manually handling pushes as without manual handling, the accessibility (aka using arrow keys and enter to navigate) will not function correctly. I believe implementing headless ui components without this accessibility made them redundant so I opted to implemented handling.

It implements localization support via the use of the localePath method.

### Related issue

- #499


andrewtavis (approved):
Seems like there are some ESLint complaints here, @nicholas-vecchio, but I can take care of them 😊 Thanks for the help here! Just tested it out and it's working great. I'll be making some page headers in the coming days and will definitely make use of this then :) :)

Issue #473, opened 2023-10-17
Allow for more than one type of media download for QR code component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

After #470 is merged the goal will be to add the ability to download the QR code from that links to an organization or event page. This work is covered by #472 :) This issue will be to finalize the QR code component by adding in a second button with dropdown options that will allow other filetypes to be downloaded. The default file type for clicking `Download QR code` will be switched to a `.png`, and the dropdown will allow for downloading `.png` (so they see it's an option), `.svg` and `.jpeg`.

- Please note that this issue is blocked by #353 and #472

A general image of how the CTA button for the component will look after this issue is:

![Screenshot_2023-10-18 00 59 21](https://github.com/activist-org/activist/assets/24387426/607c1a99-1cf9-4dde-9999-311a4b8f0d06)

### Contribution

Potentially @VinnyHS2 has interest in this one as well :) Happy to support someone who wants to work on this!

mjiruobe:
I'll try to solve this issue :-)

andrewtavis:
Thanks @mjiruobe! Let us know if we can do anything to support 😊😊

andrewtavis:
Hey @mjiruobe 👋 Quick note on this that I just realized: the QR code modal is currently not reachable via tabs. Could you also try to fix that? I can also make another issue once this is done if you'd prefer :) Accessibility wise it's good once we open it in so far as the close button is focused and we can close it with escape as well, but we just need to be able to get there. Loads of work to be done on this topic, but regardless would be good to get this one done 😊

andrewtavis:
Closed via #565 with some final work done in 6e637a9. Thanks for all the work here, @mjiruobe! 🚀

PR #565, opened 2023-11-18
QR-Code download as SVG, JPEG or PNG with selection over dropdownmenu

mjiruobe:
Hello,
I tried to solve the Issue 474: https://github.com/activist-org/activist/issues/473

This was the task:
I should implement a functionality, which allows a user to download the qr code of an event in different formats like svg, png or jpeg. PNG was already implemented by using html2canvas, but the disadvantage of this method was taking a screenshot and you can't use this as an svg.

So the solution was to build the qr code not with half html/css and half a graphic like before, instead i visualized the qr code completely as an svg including the logo, the white circle behind it, the border, the activist.org text with red hat display font and the background of it. The plain qr code already shipped as svg by the library qrcode-vue we used.

The font for the activist.org text was encoded as base64 because embedding google fonts or a web page as font provider is in my opinon the wrong decision for an activist page, which has high standards for data protection.

Problems which i did ran into:
- transform:translate seems to work different on different browsers? and this was necessary to use less calculations
- it seems like svg properties for text baseline are different in microsoft edge and ie than in other browsers
- fonts have to be included into an svg
- i used position:fixed for the dropdownbutton menuitemlist to make it overlap the modal on desktop; on mobile it had side effects but as there is no "modal" on mobile phones, so position:fixed can be excluded on phones (thanks coding night)
and some other design bugs

Please give me some code review, i will correct the changes (i think it will be necessary) and then i can create a documentation especially for the QrCode.vue because i did a lot of calculations.

I never worked with Vue.js before or on an open source project and I'm not really a web developer, so please the patience but advices are welcome

Thank you for the possibility to work on such a great project.

Greetings
Mike


andrewtavis:
Thanks @mjiruobe! Will give it a look as soon as I can 😊

andrewtavis:
Hey @mjiruobe 👋 Generally this is really really good :) :) Especially the SVG is so good! There are some minor things that I'll get to after merging it in, but I did want to check on if you have an idea on how to improve the quality of the activist `a` and `activist.org` in the PNG and JPEG exports. Is the conversion of SVG just not leading to high resolution? Could we increase the size of the base SVG such that PNGs and JPEGs that come from it are higher resolution?

Let me know if you have any thoughts on this! Happy to merge after and I can do some adjustments to the dropdown styling, etc 😊

andrewtavis:
I'll fix merge conflicts btw, @mjiruobe 😊

Note to myself:
- [ ] When this is done the modal needs `elem-on-card-style` applied to it

mjiruobe:
I fixed the quality loss bug with a specified higher resolution of the png and jpeg's. I also added the calculcated resolution to the dropdownmenu.

If there are any fixes, which i should do before you can integrate the changes, just tell me.

mjiruobe:
I see the BtnLabeled.vue needs a fix. I will go for this little issue soon.

andrewtavis:
Should be good from here, @mjiruobe! Thanks for the changes 😊😊

andrewtavis:
Ah the right rounding being removed? Sure if you can figure it out by all means :) Also happy to take a look myself.

andrewtavis:
Hey @mjiruobe 👋 Let me know what the status of this is. Happy to jump into the code and fix the buttons for you if need be! No stress 😎

andrewtavis (approved):
Got a bit carried away with fixing this, @mjiruobe, but I think we're in dramatically a better place for it all 😊 I realized that your implementation - though 100% sensical given the current architecture - really made it clear that we were relying on `BtnLabeled` to do too many things and that it was going to start getting even more difficult to maintain.

I thus made #642 to split `BtnLabeled` so that we could have a separate dropdown button from which this feature could function. The changes for that and all of the i18n key switches that it entailed are included in the most recent commit 🙌

Thanks so much for all your hard work on this issue! Really is functioning so well, was easy for me to make the needed changes, and activist really is so much better off given the new additions that this process made clear!

I'll write a bit about all this in the dev channel on Matrix now so people understand the changes :) :)

Issue #568, opened 2023-11-20
Address dev environment TS server not recognizing properties and types


wkyoshida:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue is to address reports of contributors sometimes encountering issues setting up their development environment for the frontend. Reports involve the TypeScript server in VSCode not recognizing properties and types. 

Will take this on :sunglasses:

andrewtavis:
Thanks for checking this out, @wkyoshida!

andrewtavis:
Closed via #575 😊 Thanks @wkyoshida! 🚀

PR #575, opened 2023-11-23
dev: Resolve dependencies & generate types when compose starts up

wkyoshida:
### Contributor checklist

---

### Description

I believe I may have found an option that addresses the issues we've been seeing with setting up the development environment.
- Removing the `/app/node_modules` volume we were using before to isolate the local `node_modules` and the `node_modules` in the container (they are now the same).
- Moving the `yarn install` and the `yarn dev` to outside the `Dockerfile` and into the `docker-compose.yml` so that the dependency resolution and the types generation happen when the stack starts up, as opposed to when the image is built. 

The two changes above make it so that the latest dependencies and types get resolved simply by starting up the stack, but also "stores" them locally. 

P.S.
- Retrospective - the previous setup was used to isolate the volume, since the `node_modules` in the container could get overwritten with a faulty/stale/empty local `node_modules` before. However, I believe this is now addressed with this change to do the dependency resolution in the `docker-compose.yml` instead.
- Also tried removing the legacy `npm rebuild esbuild` like we discussed, and it appears to be fine? 

### Related issue

- #568

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,9 +27,9 @@ services:
     build:
       context: ./frontend
     container_name: nuxt_frontend
+    command: sh -c "yarn install && yarn dev"
     volumes:
       - ./frontend:/app
-      - /app/node_modules
     ports:
       - "${FRONTEND_PORT}:${FRONTEND_PORT}"
       - "24678:24678"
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Makes sense, looks good and is working well for me locally, @wkyoshida 😊 Thanks for looking in to all of this!

Issue #469, opened 2023-10-17
Create password strength indicator for sign up/in


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create a progress bar that indicates password strength for the sign up and sign in pages. The designs for this component can be found on the [Sign Up / Sign In page on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=3228%3A22748&mode=design&t=XWA4AC78DOv1Jeo0-1), and can be seen below:

![Password](https://github.com/activist-org/activist/assets/24387426/582b24ed-3848-4dc3-838a-c61ee1610afc)

This component would be active on the first password entry field for signing up and the password entry field for signing in to remind the user of the strength of their password. We'll need to find a Vue/TypeScript extension of some kind to check strings for password strength and then return a percent to the user as the bar being filled in and changing colors from red to yellow to green.

### Contribution

Happy to support someone with this issue!

andrewtavis:
Ping @ahmedy00 who had interest in this 🙃

GOWTHAM2K1:
hey I am new to hacktoberfest can I do it

andrewtavis:
Thanks, @ahmedy00! Let me know if you need some support 😊

And please let me know if there's an issue you'd like to work on, @GOWTHAM2K1! Sorry for the confusion here. I literally started making the issue while we were on a call to split some work we're doing in #458, so this issue technically wasn't open.

GOWTHAM2K1:
> Thanks, @ahmedy00! Let me know if you need some support 😊
> 
> And please let me know if there's an issue you'd like to work on, @GOWTHAM2K1! Sorry for the confusion here. I literally started making the issue while we were on a call to split some work we're doing in #458, so this issue technically wasn't open.

Thank you for the clarification, No worries about the confusion.

andrewtavis:
I really have no idea what's up with those colors, @ahmedy00 🤔

andrewtavis:
I think we can close this and I'll try to assign them directly via hexes later on.

PR #583, opened 2023-11-26
#469 Password strength indicator

ahmedy00:
-zxcvbn package and types has been added
-Types configured in tsconfig.json file

### Contributor checklist

---

### Related issue

- #469


andrewtavis:
Great to have you back, @ahmedy00! 🙌 I'll get to this soon 😊

andrewtavis:
Do you want feedback on this, @ahmedy00, or should I work to finish up some things myself? Some things I'm thinking about:

- Let's have it be present even when the password hasn't been started yet as I don't like the jump for the rest of the content
- A bit of rounding of the corners would be good 😊
- Let's have some text under it as well as in the designs (right justified and `special-text` colored):
  - Password strength: invalid (bar color is `header` or `special-text`)
  - Password strength: very weak (bar color is  `action-red`)
  - Password strength: weak (bar color is  `cta-orange`)
  - Password strength: medium (bar color is  `pending-yellow`)
  - Password strength: strong (bar color is  `accepted-green`)
  - Password strength: very strong (bar color is  `text`)
- I think that having that last level where it goes to black for light mode or white for dark mode will be good 😎

Happy to work on these things myself if you're busy! Thanks for all this!

andrewtavis:
Thanks, @ahmedy00! Really excited to see the final version! 😊

andrewtavis:
Are we ready for a review here, @ahmedy00, or do you want to do a bit more? The workflow above failed because you haven't assigned the i18n keys as seen [here](https://github.com/activist-org/activist/actions/runs/7064013204/job/19231184442?pr=583), so maybe you're still planning on changing the text based on what the strength is?

ahmedy00:
@andrewtavis I think this is related to the dynamic use of the` `components.password-strength.${text}`` line, what do you suggest to fix this situation

andrewtavis:
Ah yes, I'm seeing it now. We can just do an explicit conditional where the keys are passed directly rather than using `${text}`? I know it's not the nicest way to code, but it is important that we maintain oversight over all the i18n strings.

andrewtavis (approved):
Really great, @ahmedy00 😊 I'll go through for a quick check as the colors are actually too transparent given the alphas. Let me have a look though and I'll send something along shortly!

PR #587, opened 2023-11-29
model field redundancy removed

bharath637462:
1. removed repeaded fields in models
2. created a common class and inherited.


to-sta:
hey @bharath637462 👋, thanks for contribution 😄. Makes our models all cleaner! I made some minor adjustments, since your implementation did not work right off the bat.

andrewtavis:
Just a quick note on this, @bharath637462: the [commits for this](https://github.com/activist-org/activist/pull/587/commits) show that you're sending commits from different users, with this typically meaning that there's an issue with how your email is set up for GitHub. I think that it should work for you to get credit for your commits as one of them does match your account, but I would make sure that the [email you use for GitHub](https://github.com/settings/emails) matches what you have for `git config user.email` in your local activist repo :) For later 😊

Issue #527, opened 2023-11-03
Expand a11y of CardTopicSelection


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

[CardTopicSelection](https://github.com/activist-org/activist/blob/main/frontend/components/card/CardTopicSelection.vue) was finished in #276, with overall the component working really really well 😊 One final thing that was noted was that some accessibility improvements could be made. Specifically:

- It would be nice if a user could tab from the text entry field to the tags (as of now the user needs to press the down arrow)
- Once on the tabs, it would be nice if they could be navigated between using the left and right arrow keys (as of now it only works with the up and down keys)

Note that the current implementation uses [Headless UI combobox](https://headlessui.com/vue/combobox). Moving away from this should be done only if it's impossible to add this functionality via the Vue `<script>` block.

### Contribution

Happy to support someone who has interest in working on this or get to it myself! 😊

andrewtavis:
CC @bmartins-42 who worked on #276 :)

andrewtavis:
Hey @bmartins-42 👋 Just talked this over with a coworker and we have a pretty good idea for this. Likely the best thing would have been to use an input and beneath it a [Headless UI tabs](https://headlessui.com/vue/tabs) component where the component is watching the input. This would allow us to skip from the text input to the options with a press of the tab key and then select options with left and right arrow keys. Beyond this the big thing is that apparently the tabs have an option for them being vertical, so we would just have this option enabled for mobile :) 

Let's discuss and maybe I can work on this while you look at the modal!

TeddyGavi:
@andrewtavis following up here from code night :) Happy to look into this :)

andrewtavis:
Thanks @TeddyGavi! So happy to have you back and working on this in particular 😊

andrewtavis:
5aac947 just fixes the styles as this is really proving to be difficult. What's missing:

- Whenever we tab or shift tab to the TabGroup the first element should be focussed
- If we're typing using the input and then hit tab then we should go to the first element in the TabGroup (as of now normally jumps to the next element)
- If we're on one of the tabs and hit shift tab then we should always go to the input
[...]

PR #589, opened 2023-11-30
refactor(card topic selection): to use headless ui tab component

TeddyGavi:
### Contributor checklist

---

### Description
[Screencast from 2023-11-29 05:02:57 PM.webm](https://github.com/activist-org/activist/assets/102203363/70b7f964-a5f9-4c6d-aa73-4381671fa9cb)

- refactor `CardTopicSelection` to use Headless UI `Tab` over `ComboBox` 
- `Tab` supports the use of the left and right arrow keys to traverse the tabs
- I worked to use a listener to override the default functionality of `ComboBox` 
- default operation as seen here https://github.com/tailwindlabs/headlessui/blob/6598db59b7c2d78d00bab4c9c9314a526ddb79e6/packages/%40headlessui-react/src/components/combobox/combobox.tsx#L656-L683
image
- However, I could not get the event to stop before hitting these cases
- Manually tested by creating a new organization on a locally running instance of the app

### Related issue

- #527


andrewtavis:
Just checking some expected behavior here, @TeddyGavi:

- It seems like `selectedTab` might need to be reset to the first index when we tab away from the tab group?
- We start on [the preview's create page](https://deploy-preview-589--activist-org.netlify.app/en/organizations/create/)
  - Tab down and we get to `Justice`
  - Tabbing again goes to the checkbox (expected)
  - Tab back to `Justice` and then right right to `Environment` and select it
  - Tab to the checkbox and back and then we're back to `Environment`
    - I'd expect to tab into the first thing always, but maybe I'm off base on this
  - Also once we're back on `Environment` and we switch to `Justice` with the arrow keys, if we tab then we go back to `Environment` where I'd expect it to go to the checkbox
  - Also if we deselect `Justice` then we lose focus on it, then tabbing to it gets to it

Let me know how the above sounds to you, and please provide your feedback on the thought process! Thanks for all the work till now as well 😊

TeddyGavi:
@andrewtavis I removed the `selectedTab` and set the default index to 0, with a sort() applied. 

Test this out and see if the tab and shift + tab works as expected :)

andrewtavis:
Seems to be doing a bit better, @TeddyGavi. Issue is that it's the first option on load, and if that option is deselected then we jump to where it is when we tab into the tabs rather than the first position 🤔 Beyond that I'm still getting cases where we can loose the focus and the user will be forced to tab out and tab back in to do anything else. Example for this is tabbing into the tabs, arrow keys to `Environment` and select it, arrow keys to `Education` and select it, and then arrow keys to `Activism` and select it -> `Activism` is deselected, but we've also lost the focus state.

I can take a bit more of a look at it later. Time to sleep 😴 Thanks for the dedication here!

andrewtavis:
Hey @TeddyGavi 👋 Let me know what your thoughts on this are :) Also happy to work to finalize it myself if you're a bit busy! I think it's right there and just needs some tweaking.

TeddyGavi:
Hey @andrewtavis I'm happy to have you finalize it!

andrewtavis (approved):
Bringing this in and will work to finalize this, @TeddyGavi! Thanks for the hard work to bring this to fruition! 🙏

Issue #559, opened 2023-11-18
Create date picker component for event creation flow


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would use [v-calendar](https://github.com/nathanreyes/v-calendar) to create a date and time picker for the event creation flow. This ultimately will be housed in a `CardDatePicker` component, which eventually will include a timeslot component, but for now will not. Designs for this can be seen [on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2001%3A1299&mode=design&t=hbrmPFJFE6NT0uV9-1) and can also be seen below:

![Time Inputs](https://github.com/activist-org/activist/assets/24387426/f0deffaa-5127-4e8d-8742-821a601d3fd1)

Again the time slot selection is not necessary at this time :)

### Contribution

Happy to support someone or work on this myself! Let's discuss how best to implement this, with it likely being best that we have the data picker in a component that then is used in the card 😊

PabloSSena:
Hello @andrewtavis, please assign me to this one :D

andrewtavis:
Nice @PabloSSena! Thanks for jumping on multiple issues! 😊

PabloSSena:
Hello @andrewtavis, the calendar needs to be exactly how it looks like in the figma project? because I saw in the Vcalendar docs that they have a model that attends what we want but it look like this 
![image](https://github.com/activist-org/activist/assets/58041720/319d797d-0be9-408b-8571-cec03f445ec8)

andrewtavis:
Hey @PabloSSena 👋 Let's do out best to style it, but for an initial implementation it's totally fine to use the base structure as it is there :)

PabloSSena:
Okayy I'll do my best here

andrewtavis:
Don't stress too much about it, @PabloSSena :) Big thing is getting the colors right so it's working with the rest of the page elements. We can figure out more exact styling later 😊

PabloSSena:
hello @andrewtavis, what i got untill this point is looking like this 
![image](https://github.com/activist-org/activist/assets/58041720/762e09ad-dff9-40f6-9233-6d4bf6455f7a)

[...]

Issue #560, opened 2023-11-18
Add event creation flow


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add in the event creation flow that populates the backend. Designs for this can be seen in the top [prorotype on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2001%3A1299&mode=design&t=hbrmPFJFE6NT0uV9-1) (the one for UNLOCK accelerator). Not all components for this finished at this time, with specifically #559 being a separate issue. 

### Contribution

I'd be happy to work on this or support someone who has interest! 😊 This is a needed flow for release, so getting to this sooner rather than later would be important 💪

PabloSSena:
Hello @andrewtavis i'm interested in this issue but i'm a bit confused about what is to do exactly, it is to implement only the front-end or also the back-end

andrewtavis:
For now just the front end, @PabloSSena :) I can maybe make the routes and the basic buttons at the bottom for navigation and you can add in the other parts for the pages?

andrewtavis:
Will get to that later today, @PabloSSena :)

PabloSSena:
I don't think I will be able to work on this today, but i can start tomorrow

andrewtavis:
Ok, @PabloSSena :) fd75503 added in [activist.org/events/create](https://activist.org/events/create). I changed the designs to not have multiple pages, as I think we'd be fine to have all of this be in one form that transitions from "page" to "page" as the user clicks through. What do you think about this rather than separate pages?

PabloSSena:
Hi @andrewtavis, one more time sorry for the late response,I agree that only one form is better, I worked in a place that we did something very similar to this but we used angular, i can't imagine how to this in vue without studying a little more

andrewtavis:
[...]

PR #591, opened 2023-11-30
Date picker event

PabloSSena:

### Related issue

My pull request have the purpose of adding a a DatePicker to the event creation workflow, It isn't funcional by the moment.
This is how it's looking like.

#559


andrewtavis:
Hmmm, very weird @PabloSSena. Now the original account's not in there. Your account's likely not gonna get credit for the original commits. Should we keep this one?

PabloSSena:
Yes we can keep @andrewtavis, no problem

andrewtavis (approved):
This is looking really good, @PabloSSena :) There are some minor things that would be good to change, but I think each can be individual issues.

- The date colors in dark mode aren't set appropriately (too dark or too high of an alpha, but I can fix this!)
- We should add functionality that the time selection goes away or is set to 00:00 to 23:59 and can't be changed when the all day button is pressed
- We should allow for selecting multiple days when that form button is selected
- The checkboxes should be our branded ones (I can also do this)

Will bring this in and send along the changes that create the pages I owe you for #560 😊

andrewtavis:
Will make the issues after I send along my changes :)

frontend/pages/events/index.vue
```diff
@@ -20,10 +20,15 @@
         :event="event"
       />
     </div>
+    <div class="pt-3 pb-6 space-y-6 md:pt-4 w-1/2">
+      <CardDatePicker></CardDatePicker>
```
andrewtavis, line 24: Ah btw, with Vue it's totally ok to do a self closing component tag, so we just need `<CardDatePicker />` :)

frontend/pages/events/index.vue
```diff
@@ -20,10 +20,15 @@
...
+    </div>
   </div>
 </template>
 
 <script setup lang="ts">
+import CardDatePicker from '~/components/card/CardDatePicker.vue';
```
andrewtavis, line 30: And with Nuxt there's no need to import the components as they're all loaded automatically ✨😊 Types we need to load, but components not :)

frontend/components/card/CardDatePicker.vue
```diff
@@ -0,0 +1,143 @@
...
+        trim-weeks
+        expanded
+        :attributes="attributes"
+        transparent
+        borderless
+        ></DatePicker>
+    </form>
+  </template>
+  
+  <script setup>
```
andrewtavis, line 84: Also no stress as TS wasn't causing any errors anyway, but for the `<script>` block we need it defined as `<script setup lang="ts">` to turn on TS :)

Issue #523, opened 2023-11-01
Add default values to organizations and other entities in the backend


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now the backend is missing default values for certain fields like organization members and supporters. These should be added, and generally a sweep of the current [database schemas](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377%3A33014&mode=design&t=W5dWwYl5XmQ8q05M-1) should be done to make sure that the fields are updated appropriately.

### Contribution

@sayik said in the last dev sync that he'd like to work on this 😊 Happy to support!

andrewtavis:
@sayik, can you write in here so I can assign you? :)

sayik:
Hi @andrewtavis can I work on this issue?

andrewtavis:
Thanks @sayik! Let me know what I can do to help! For now maybe look through the backend code to familiarize yourself and then we can look through the values that are missing together 😊 I'll also take this as an opportunity to update the schemas and we can add some new tables in another issue after this one!

andrewtavis:
Hey @sayik! Just checking post sync :) Plan is to send along an initial PR and we'll check in next week on a new issue for the changes?

andrewtavis:
I can also bring it down and format it for you, @sayik :) No stress 😊

andrewtavis:
Closed via #592 🥳 Thanks @sayik! @to-sta and I did some minor changes with the support of @wkyoshida as well. We're gonna look at the schemas more on Saturday as well 😊 Hope you can make it!

PR #592, opened 2023-12-01
523 add default values to models

sayik:
### Contributor checklist

---

### Description

I have made the following changes:

- Added default values to content, event, entities models based on the activist figma public design for models
- Tested updated models using django admin and checked whether tabels were getting inserted without errors.

---

### Related issue

- #523


backend/authentication/models.py
```diff
@@ -69,7 +69,8 @@ def __str__(self) -> str:
 
 class UserResource(models.Model):
     user_id = models.ForeignKey(User, on_delete=models.CASCADE)
-    resource_id = models.ForeignKey("content.Resource", on_delete=models.CASCADE)
+    resource_id = models.ForeignKey(
```
andrewtavis, line 72: Note that this line needs to be reformatted :)

Issue #582, opened 2023-11-26
Create base modal component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This component would create a base `ModalBase.vue` component in the [frontend/components/modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal) that others in this directory would inherit from such that there is a common styling for modals across activist. Note that work for this issue should likely be done before the recently made #580 and #581. The logic for this component could be extracted from the already existing modal components with a `<slot />` included such that the modal content can change :)

### Contribution

Happy to support someone with interest in working on this or work on it myself! 😊 This is a great `good first issue`, or also would be perfect for a current contributor who would like to pick up something new :)

andrewtavis:
Hey there @VinnyHS2 👋 Wanted to check in on if you had interest in this one along with #580 or #581 :) Figured any of these would be pretty easy for you to knock out, but also happy to discuss/do a call!

anthonyshibitov:
heyo, is this still up for grabs? more than happy to do this!

andrewtavis:
Ahhhhh shitttttt 😎 Great to hear from you, @anthonyshibitov! By all means you can work on this :) Lemme know if I can do anything to support 😊

anthonyshibitov:
> You coming to the sync on Saturday?

yessir!! i've got some catching up to do 🤠

andrewtavis:
Sickkkkk! See you there! 😊

andrewtavis:
51e61cb closes it up, @anthonyshibitov :) Image modals can now close on the BG click (the w-4/5 was restricting it, so I assigned it lower in the divs) 😊 I also made it so the slots are single tags as that's a bit more the style we're using. Aside from those some minor formatting. Thanks for this! Great to have you back 🥳

anthonyshibitov:
[...]

PR #595, opened 2023-12-02
Refactor ModalBase from existing modals

anthonyshibitov:
Refactor ModalBase from existing modals

ModalBase can accept an imageModal prop to change behavior and match existing functionality.

### Contributor checklist

---

### Description

Refactor modal functionality to a separate `ModalBase.vue` component.

`imageBase` can be added to the `<ModalBase>` component to mimic the functionality we currently have, where images and text modals behave (click image to close modal) and display (text modal has close button and title in a card, with the content within it) differently.

Note that this currently does not support i8n in the modal title (specifically, `{{ $t("components.modal-qr-code.header") }}`). I wanted to get some feedback on this before fleshing out the remaining portions, to make sure this design is what we're looking for. Let me know if this is okay, and I will commit another change to try and fix the i8n issue (I'm currently not sure how to do this, any ideas would be great lol 😀). If we want the modal to behave differently, I'll go back to the drawing board.

Also, note to myself, the bug where clicking to the left or right of the image modal does not close the modal still exists 🙃

### Related issue

- #582


andrewtavis:
Thanks @anthonyshibitov! I'll check this in the coming days and provide some feedback 😊

andrewtavis:
Looking good so far, @anthonyshibitov! I'm going to wait on #565 to get merged before bringing this in. Some thoughts on this:

- For the image maybe removing the active state when it's expanded would be good as when the user clicks the image to close it gets a border around it, but clicking the sides it won't (maybe will be solved by the change to get the left and right parts of the expanded modal to work to close the image)
- Looks like you're using npm to set things up rather than yarn, which is what it is, but I'd warn against installing any dependencies in this way as we've found that npm installed dependencies can break things
  - Regardless, can you revert the package-lock.json files 🙃 
- Re you're question on the i18n changes to this:
  - I think this could make sense, or maybe we can also just drop all the card contents including the title into s `<slot>`? Might be a bit weird with the close icon, but that way all the logic for the modal card would be in the slot and we wouldn't also have to assign some parts in the props.
  - Lemme know what you think! It'll work either way :)

andrewtavis:
I can take a look at it myself, but when I click the image to close it there's a border that appears around it, whereas when we'd click the sides there wouldn't be. Trying to make sure that the closing modal experience is the same :)

andrewtavis:
Are we good for a final review on this, @anthonyshibitov? :)

anthonyshibitov:
> Are we good for a final review on this, @anthonyshibitov? :)

yessir!

andrewtavis:
Wunderbar 😊 I'll get to it soon then :)

frontend/components/modal/ModalImage.vue
```diff
@@ -1,75 +1,42 @@
...
-            class="object-contain p-12"
-            :src="imageURL + '_dark.png'"
-            :alt="$t(imageAltText)"
-          />
-        </button>
-      </DialogPanel>
-    </div>
-  </Dialog>
+  <ModalBase imageModal>
+    <template #normalDisplay>
```
andrewtavis, line 3: This is some great stuff here, @anthonyshibitov :)

andrewtavis: Really well organized. The double multiple slots that are named makes this super easy to follow 😊

andrewtavis (approved):
Thanks for this, @anthonyshibitov! I'll try to play around to get the image boundaries to be clickable to close it :)

Issue #552, opened 2023-11-18
Create CardFAQEntry component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This is a very important issue that would be good to finish sooner rather than later. We've recently finished the designs for how FAQ entries will function, with the decision being that we'll do disclosures. Designs for this component can be seen on Figma [here](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=7088%3A92710&mode=design&t=hbrmPFJFE6NT0uV9-1) and further are pictured below:

![FAQ Question](https://github.com/activist-org/activist/assets/24387426/279f7daf-d270-4afa-a272-9997fb1eceaa)

We'd use [Headless UI disclosure](https://headlessui.com/vue/disclosure) for this, with one usage of this already being present on activist for the [CardLegalDisclaimer](https://github.com/activist-org/activist/blob/main/frontend/components/card/CardLegalDisclaimer.vue) component.

### Contribution

Happy to support someone who wants to work on this! 😊 This is a great `good first issue` for someone wanting to contribute to the activist community! 🚀

andrewtavis:
Big thing for this is that we have the Tailwind class `card-style` :)

mattburnett-repo:
writing in here

andrewtavis:
Thanks, @mattburnett-repo! Let me know if there's anything I can do to help :)

andrewtavis:
Hey @mattburnett-repo 👋 Checking in to see if there's anything I can do to support on this :) Hope you're well!

mattburnett-repo:
hey @andrewtavis Thanks for checking in. I've been really ill the past week and have fallen behind on a lot of things. I'll try to get this done before the weekend.

andrewtavis:
Sorry to hear that you've been unwell, @mattburnett-repo! 🍵🌿 Take your time :) Just checking in 😊

Hope to see you for the sync on Saturday!

andrewtavis:
[...]

PR #598, opened 2023-12-04
fix: Issue #552 add CardFAQEntry

mattburnett-repo:
### Contributor checklist

---

### Description

This PR added basic CardFAQEntry functionality per Issue #552 

### Related issue

- #ISSUE_NUMBER
#552


andrewtavis (approved):
Really nice, @mattburnett-repo 😊 Only thing I'd change here is that the whole disclosure isn't able to open it, but rather we have to click on the question. This is generally fine for usability with a mouse, but it looks a bit weird with a keyboard as the focus state is just on the question and then expands to the whole disclosure once it's open. Would be better if the whole component is focussed and then it just gets a bit more height when it's open (mostly the dramatic width shift is what I'm talking about). I'll bring this in and do what I assume is one line of Tailwind and we'll be all done! Thank you!

Issue #561, opened 2023-11-18
Add resource creation flow


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add in the resource creation flow to that would be activated via `Create -> New resource` from the left sidebar. Designs for this flow can be seen [on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2584%3A16751&mode=design&t=hbrmPFJFE6NT0uV9-1)  and further can be seen below:

![Resource Creation](https://github.com/activist-org/activist/assets/24387426/0153ee27-ddba-4b4e-ab61-24d89cbaeab2)

Components included in this would be the topic selection one, progress bar, and the form element could be custom or work from others in the code base 😊

### Contribution

Happy to work on this myself or support someone who has interest!

IsaacDMz:
Hey @andrewtavis can i work on this issue? I have some difficulties with front-end but i would like work on this.

andrewtavis:
Sure thing, @IsaacDMz! Let me know if there's anything we can do to support you with the issues you're having with the frontend! Assigning you now 😊

IsaacDMz:
@andrewtavis i have a question, this task is to create the Resoures button in the dashboard and redirect to resources page? Like this 
![image](https://github.com/activist-org/activist/assets/94393622/3608710d-f584-4212-9a3d-a28f8d7f76c7)
Or create the resources button and create the page as an example in figma?

andrewtavis:
Thanks for checking in, @IsaacDMz :) We want to have `Create > New resource` in the sidebar link to `/resources/create` with this page then being created as well :) As of now resources won't be enough of a focus to have a slot in the main menu selectors at the top of the sidebar. Thanks for checking in though!

You'd be welcome to do one PR that does the routing and creates a `/resources` directory in the frontend with `index.vue` and `create.vue` being pages that for the first PR just have the sidebar, and then we can do another PR that adds in the page contents to `/resources/create.vue`. All in one PR's also totally fine though 😊

IsaacDMz:
Hey @andrewtavis Firstly, I apologize for the delay in starting to work on this task, I was finishing up some college things.

[...]

PR #599, opened 2023-12-05
Feat/create resources page

IsaacDMz:
### Contributor checklist

---

### Related issue

- #561 

image


andrewtavis:
Thanks @IsaacDMz! I'll get to this soon :)

frontend/i18n/en-US.json
```diff
@@ -370,6 +370,19 @@
...
   "pages.events.team.tagline": "Event organizers",
+  "pages.resources.create.header": "Information",
+  "pages.resources.create.subtext": "Thanks for linking a resource to activist! This is how you can keep all the tools and information that you need to get active in one place. Please only link resources that you have the right to share with others. You can learn more about our terms and confitions for copyrighted materials on the resources page of the documentation",
+  "pages.resources.create.resource-name": "Name",
+  "pages.resources.create.resource-name-placeholder": "A name for the resource",
+  "pages.resources.create.link": "Link",
+  "pages.resources.create.link-placeholder": "A URL to the resource",
+  "pages.resources.create.description": "Short description",
+  "pages.resources.create.description-placeholder": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua",
+  "pages.resources.create.location":"Location",
```
andrewtavis, line 381: Note that this should be combined with the other "Location" i18n key. Ping @to-sta that an error for the workflow wasn't triggered :)

IsaacDMz: Should i use the organization location on this key, or add Location to global keys?

andrewtavis: You can check the Python script that was supposed to fail on the branch and should fail locally :) Run `python frontend/i18n/check_repeat_i18n_values.py` and it should list this values that have been repeated and also suggest what the lowest level of key should be.

andrewtavis: Looks like we can do `"pages._global.location": "Location"` :) Give it a check though 😊

andrewtavis: Nice @IsaacDMz :) Thanks! I'll get to this in the coming days. My home internet's a bit screwed up atm, and I don't like reviewing on my phone's hotspot 😅

andrewtavis (approved):
Very solid work, @IsaacDMz 😊 Thanks for coming in quick and fixing the i18n "issues" even though the workflow failed. Really appreciate all of this!

As we have the `/resources` being linked to via the breadcrumbs I'll go ahead and add a placeholder page in there. We can even do some quick page analytics and just do a list of the resources in order of how much they've been viewed for the time being 😊 Will close the issue when that's all done!

Issue #462, opened 2023-10-15
Create CardOrgApplicationVote


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create a component in the [components/card](https://github.com/activist-org/activist/tree/main/frontend/components/card) directory where that houses the ability for organization admins to indicate their support for other organizations to join activist. As this is a main component of how the activist community will expand, this issue is very very important 😊 The designs for this can be seen [on Figma on the organizations tab](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2642%3A20194&mode=design&t=jTUMfp3ZcqY6KqEK-1). One prop needs to be whether the vote is closed or not, with this then disabling the `BtnLabeled` components with right icon props passed for an up or down arrow.

### Contribution

This is a bit more of an involved `good first issue`, but I feel like it's definitely something that someone could jump into if they had interest. This component will doubtless evolve over time, but making an initial version now would allow us to start user testing it 😊 It's also of course open to the dev community if they want to pick it up 🚀 Happy to support!

rigojr:
Hi @andrewtavis I would like to take of this issue. I have issues to get into the dashboard, How can I get into?

andrewtavis:
Hey @rigojr 👋 Happy to assign you! Let me know what you mean by the dashboard. As of now we actually don't have the page in question for applications for organizations. That should be at `pages/organizations/[id]/application`, which could be added into this issue? The other parts of that page are the discussion card components that are being worked on right now :) 

Let me know if there's something else you mean though! Nice to hear from you 😊

rigojr:
Hi @andrewtavis, I thought that we have indeed a page were I can add this component and work on it.

I would like to make some questions about the design.
![image](https://github.com/activist-org/activist/assets/22437427/ca76a311-291e-4af5-8842-706eea8aa2ab)

- What does it represent those icons? Last three people that voted? or members of the organization?
![image](https://github.com/activist-org/activist/assets/22437427/9b15f3dc-d035-4b2a-862f-4982425143a0)

- I'm curious about how it works this poll section
[...]

PR #609, opened 2023-12-17
add card org application vote

rigojr:
### Contributor checklist

---

### Description

- Add the org application vote
- To test the feature use the following [http://localhost:3000/en/organizations/123/about?status=pending](http://localhost:3000/en/organizations/123/about?status=pending)
- A temporal param has been added for testing purpose, it'll remove as soon as the required information is available

Notes:
- The component has `upVotes` and `downVotes`, those are `ref` variable for testing purpose, this information should came as props.
- An event mechanism should be set between the component an their clients in order to communicate up and down votes, the `click` event could be forwarded.
- In the list of favor organizations, where the organization's icon is being shown, a better scope should be defined.
  - What happen if an organization do not have an icon set?
  - Should we limit this list to only 3 organizations? following the figma example?
  - How should we sort those organization display? what should be first and last?
- The disabled option need to be added in the `btnLabeled` component.
- Change made in `about.vue` for testing purpose should be removed.
 
Screen Capture
image

### Related issue

- #462


andrewtavis:
Thank you, @rigojr! I'll get to this soon 😊 Really happy to have this feature 🥳

rigojr:
Hi @andrewtavis I believe I could improve this implementation. Nevertheless, I would like to expand the information about it's scope, let me know what do you think about those notes I post in the description of the PR.

Apart of that what is the purpose of the `ariaLabel`? and the property `supporters` in `Organization` answer to what?

andrewtavis:
Answers to some of your questions, @rigojr 😊

> In the list of favor organizations, where the organization's icon is being shown, a better scope should be defined.
> - What happen if an organization do not have an icon set?

Would just be the default organization icon that we have.

> - Should we limit this list to only 3 organizations? following the figma example?

No the list shouldn't be limited. Maybe after a certain number of organizations we'd show more and then we can see a modal or expand the field to show all? Do you have any suggestions?

> - How should we sort those organization display? what should be first and last?

First organization is the first to upvote, and they'd follow in order from there :)

> The disabled option need to be added in the btnLabeled component.

Makes sense so we can disable the vote buttons! You'd be welcome to add this in here, but also I can add it after merging.

> Apart of that what is the purpose of the `ariaLabel`?

`ariaLabel` should be setting the `aria-label` for a11y for screen readers. I'm assuming you mean for the buttons? Is this not what you're seeing?

> and the property `supporters` in `Organization` answer to what?

Supporters is similar to say a follower on another platform. You'd get notifications from organizations and statements from the organization would also go out to you. When an organization comes onto the platform we'd want all organizations that have voted for it to automatically be supporters. This is also something for us to keep track of on the community end of things as if we see that an organization is being problematic and further that all of its original supporters are not supporting anymore, then it'd be a good indication that something's not right.

Let me know if you have further questions!

rigojr:
Hi Andrew, I just uploaded a final version of this PR, some key points I'd like to mention.
- I remember for the last tech sync we had that the organization status would be held by the PostgreSQL with numbers, we should add an `enum` to express that status in good type terms. It'll avoid future problems. Can you list me the possible options an organization status could have? I can add it.
- About the maximum count of organizations, I believe that the best approach is to limit to a certain number that depends on the device the web app is being shown since the screen could be larger or smaller, the number of organizations available for the first view should be limited by that number, we can add at the end of that list the organization icon with a bubble contained number (Like the red number that we had in iPhone when an application has notifications) in the upper right corner of the icon.
- I noticed that the `IconOrganization` doesn't support the dark theme.
- The vote button is being resized when the inner count changes, we should take care of that in the future.
- We could improve the `up-vote` and `down-vote` with two-way binding. However, I don’t think we should since the voting process could interact with external and probably async artifacts, so I’ll advise waiting.

I can take care of those things I noticed, no problem, I just want to let you know.

For finishing, some capture
**6 organizations with defined icons**
image

**15 organizations with defined icons, only 10 are visible, for now**
image

**6 organizations with undefined icons**
image

** 15 organizations with undefined icons, only 10 are visible**
image

**card org application vote component in about page when query param is not defined**
image

andrewtavis:
Hey @rigojr 👋 Thanks for the continued work here! Answers to your questions:

> Can you list me the possible options an organization status could have? I can add it.

They could be 1: Pending or 2: Accepted. We're encoding 1 with the defaults always, with the default being Pending and 2 is the only other option, which would mean that enough organizations have expressed interest in them coming on. We might do 3: Declined, but that's up for continued conversation and it sounds like from the sync it's not a priority to make a final decision on including it or not. Your feedback on whether we'd close an application or not would be welcome! Obviously not deleting the application, but whether we'd ever say "No" rather than "Looks like there's no interest" as a community.

> About the maximum count of organizations, I believe that the best approach is to limit to a certain number that depends on the device the web app is being shown...

Completely agree!

> I noticed that the IconOrganization doesn't support the dark theme.

It does! You just need to use `dark:fill-light-text` and it should work :) Hopefully we can change these later such that we can just use the text property for them, but as of now it's an SVG that's called like an icon, so we need to use `fill`.

Thanks for the other suggestions!

A final thing, can we put a light border around the images thats `border-light-section-div dark:border-dark-section-div`? That way if there is a white background we'll still see the organizations as squares in a similar way that we see the contributor GitHub icons on the [supporters showcase](https://activist.org/en/supporters/showcase).

Looking forward to the final review and the merge! 🚀

andrewtavis:
Quoting myself:

> Hopefully we can change these later such that we can just use the text property for them, but as of now it's an SVG that's called like an icon, so we need to use `fill`.

If you have some ideas on this it'd be great! Would be an issue I'd be happy to make and maybe you could work on it :)

rigojr:
Howdy, @andrewtavis 
- The `dark:fill-light-text` didn't work for me, I used `dark:fill-white` is that ok?.
image
image
- in other PR I'll add an `enum` for the Organization types and related refactors.
- I add `border rounded border-light-section-div dark:border-dark-section-div`
image
image

> If you have some ideas on this it'd be great! Would be an issue I'd be happy to make and maybe you could work on it
I'll think about it, I don't have a solution right away, We should keep it under the radar.

andrewtavis:
Thank you, @rigojr! Looks like we're ready for a review here :) I can see if I can figure out why the `dark-text` isn't working on my end, but regardless I'll try to bring this in by end of week 😊

andrewtavis (approved):
Looking great here, @rigojr! Really happy to have had you working on this. I'll bring it in and see if I can figure out the `dark:fill`, but aside from that we're looking great as far as the component is concerned. I think for the future we should redirect to the application in case the organization is pending, but we can check that out later 😊

Will write in the issue when I'm all done!

PR #615, opened 2023-12-24
fix(dev): Add FRONTEND_PORT env to configure port on compose

wkyoshida:
### Contributor checklist

---

### Description

Coming out of the [discussion](https://matrix.to/#/!CRgLpGeOBNwxYCtqmK:matrix.org/$5OLIJ-4DHZHS4U3DrDfkHrfkunWkiCsdEV9IcTxBInI?via=matrix.org&via=acter.global&via=chat.0x7cd.xyz) in the Matrix, decided to look into why trying to set a different front-end port was not working. Seems that we were missing configuring it in the command when starting up the stack. 

Before, `FRONTEND_PORT` was only being used for exposing the container ports.

### Related issue

- N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
     build:
       context: ./frontend
     container_name: nuxt_frontend
-    command: sh -c "yarn install && yarn dev"
+    command: sh -c "yarn install && yarn dev --port ${FRONTEND_PORT}"
     volumes:
       - ./frontend:/app
     ports:


andrewtavis (approved):
Thanks for thinking of this in reading through the conversation, @wkyoshida! @GrandSuccess87, this should fix the issue of the ports not being picked up that you were experiencing 😊 I've also made #617 where we can improve the setup docs a bit more :)

GrandSuccess87:
Wonderful, thank you for letting me know! I too was wondering if it's possible to run this without Docker. Looking forward to the outcome of #617 .

Issue #624, opened 2023-12-24
Setup Resource API


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Create a API for the resource model

### Contribution

I will work on this myself

Issue #620, opened 2023-12-24
Add image table and referencing columns to backend


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue focuses on adding in the image table to the backend in the [content directory](https://github.com/activist-org/activist/tree/main/backend/content). The schema for this can be found in the [designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=7917%3A1375&mode=design&t=eKjevthu2v68bHHV-1), with the `image` table being needed as well as changes to the other tables to add in the columns that it is linked to via foreign keys.

### Contribution

Happy to discuss and support with this as needed!

andrewtavis:
Ping @rigojr: would this be a backend issue that would be of interest? Feel free to write in here if so and I can assign! We'd talked maybe you working on the votes tables, but that feature is pretty far off at this point. This issue would give you a good orientation to the backend models though 😊

Also ping @to-sta: I figured adding this one in made more sense. We can do votes, tags and movements later next year. The only other table that's `Not implemented` now is location, and I'll make an issue for that now too :)

rigojr:
Let's do it, I can handle it.

andrewtavis:
Awesome, @rigojr! Let us know if there's anything we can do to support 😊

andrewtavis:
Closed via #363 🎉 Thank you, @rigojr! Let us know if you have any ideas on #670 😊

PR #625, opened 2023-12-24
Initial Resource API setup 🎅🎁🎄 

to-sta:
### Contributor checklist

---

### Description

In this PR, I have started implementing the first API for the Resource table. I had to make some changes to the table in order to make the privat field useful. I added the created_by to implement the logic for GET requests. 

In case a user is the creator for a resource and the privat field is set to True, the creator should still be able to retrieve the resource. Permissions are set to `IsAuthenticatedOrReadOnly`. 

Additionally, I have deleted a few validation utility functions inside the serializer, because the model itself already provided sufficient validation.

`GET, POST, PUT, PATCH` methods are setup with the viewset methods. I used the  

Tests have been manually via swagger for all the methods and endpoints that are covered by the ResourceViewSet. I would suggest to create unit tests for API later on, when we think it is stable. 

### Related issue

- #624

### DIFF
--- a/backend/content/factories.py
+++ b/backend/content/factories.py
@@ -15,8 +15,9 @@ class ResourceFactory(factory.django.DjangoModelFactory):
     url = factory.Faker("url")
     total_flags = factory.Faker("random_int", min=0, max=100)
     private = factory.Faker("boolean")
+    created_by = factory.SubFactory("authentication.factories.UserFactory")
     creation_date = factory.LazyFunction(datetime.datetime.now)
-    deletion_date = factory.LazyFunction(datetime.datetime.now)
+    last_updated = factory.LazyFunction(datetime.datetime.now)
 
 
 class TaskFactory(factory.django.DjangoModelFactory):
--- a/backend/content/models.py
+++ b/backend/content/models.py
@@ -18,7 +18,7 @@ from django.db import models
 from backend.mixins.models import CreationDeletionMixin
 
 
-class Resource(CreationDeletionMixin):
+class Resource(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     name = models.CharField(max_length=255)
     description = models.TextField(max_length=500)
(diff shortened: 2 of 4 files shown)


to-sta:
@andrewtavis Could you explain the field total_flags to me? What is it used for and how should it be incremented or set?

andrewtavis:
> @andrewtavis Could you explain the field total_flags to me? What is it used for and how should it be incremented or set?

Hey @to-sta! Thanks for all this :) `total_flags` would be used by people to indicate to use that a resource is problematic. Would be set to 0, and it's incremented by one for each person who flags the resource :)

to-sta:
> Hey @to-sta! Thanks for all this :) `total_flags` would be used by people to indicate to use that a resource is problematic. Would be set to 0, and it's incremented by one for each person who flags the resource :)

hmm if it's only a IntegerField, then are bad actor could incremented this multiple times because we are not keeping a record of who flagged which resource.

andrewtavis:
Very true! Shit. So are we needing a `flag` table then in a similar way that we have the `support` table?

to-sta:
Hmm total_flags is the sum of unique user_id and resource_id. So the table needs at least those two.

We could either query the total_flags or we could increment the total_flags count when an entry is made for the flag table (on save).

andrewtavis:
Thanks @to-sta! I'll also make a quick table in the designs.

andrewtavis:
For here we're good for a review then?

to-sta:
> For here we're good for a review then?

Yeah, all good for the review.

andrewtavis:
Hey @to-sta: FYI I threw together the issue right now as I've been making a bunch of "add table from schema" issues recently anyway 😇

- https://github.com/activist-org/activist/issues/626

Let me know what you think!

to-sta:
@andrewtavis will create Issues for the API's today 😃 and it's great idea if you get some full stack experience. Especially the API design is handy when you work on fetching data later on in the frontend.

@baibhavKumar1 was also interested in working on the upcoming backend issues.

andrewtavis:
All sounds good! For those tagged, please let us know if we should tag you in new issues!

baibhavKumar1:
I am working in backend, and happy to implement sir, but my stack has NodeJS instead of Django.

andrewtavis:
You'd be welcome to give it a try, @baibhavKumar1 :)

to-sta:
@baibhavKumar1 sry my fault, I tagged the username :sweat_smile: should have been @bharath637462

bharath637462:
hi @to-sta fianlly a task from backend what should i do come on come on

rigojr:
Happy to help, let me know if an API ticket is free for me. I'm about to start working on https://github.com/activist-org/activist/issues/620

CC: @andrewtavis  @to-sta

Issue #627, opened 2023-12-26
Organization API


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We are starting with creating the API's that will be used by the frontend in production soon 😃. 

We are using [django-rest-framework](https://www.django-rest-framework.org/) in the backend to create our API's. ViewSets in particular the [ModelViewSet](https://www.django-rest-framework.org/api-guide/viewsets/#modelviewset) are used for creating the different HTTP methods. Another benefit of using the ModelViewSet is that we are getting some validation from the model itself and only need to extend that in the serializer if necessary.

Have look at the [scheme](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377-96159&mode=design&t=8xoYzpakJ6ebt8Ie-4) and the models.py file. 

Task is to implement all these methods for the OrganizationViewSet that represent the different HTTP methods:

```python
def list(self, request):
    pass

def create(self, request):
    pass

def retrieve(self, request, pk=None):
    pass

def update(self, request, pk=None):
    pass

def partial_update(self, request, pk=None):
    pass

def destroy(self, request, pk=None):
    pass
```

The API generally is a combiniation of:

- views.py
- serializer.py
- models.py

Also feel free to suggest changes to the model, in case it is necessary for the logic. They are not set in stone 😃. In terms of testing, you can use Swagger UI for manuall testing (backend endpoint: `/v1/schema/swagger-ui/`).

>[!NOTE]
> Please use an appropiate http [status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) for your responses.

### Contribution

I am happy to help with this issue.

andrewtavis:
I'll take this one, @to-sta 😊 We can also make ones for organizations and FAQs as well to round out the main things from the start, and then maybe events. As you see fit :) :)

to-sta:
hmm I was thinking about starting with the main tables:

* resources
* organizations
* event
* topic
* task

What table/model represents the for example FAQ?

bharath637462:
@to-sta can i create api for 
resources
organizations
event
topic
task
below tables?

Jackblanket847:
def list(self, request):
    pass

[...]

Issue #628, opened 2023-12-26
Events API


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We are starting with creating the API's that will be used by the frontend in production soon 😃. 

We are using [django-rest-framework](https://www.django-rest-framework.org/) in the backend to create our API's. ViewSets in particular the [ModelViewSet](https://www.django-rest-framework.org/api-guide/viewsets/#modelviewset) are used for creating the different HTTP methods. Another benefit of using the ModelViewSet is that we are getting some validation from the model itself and only need to extend that in the serializer if necessary.

Have look at the [scheme](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377-96371&mode=design&t=8xoYzpakJ6ebt8Ie-4) and the models.py file. 

Task is to implement all these methods for the OrganizationViewSet that represent the different HTTP methods:

```python
def list(self, request):
    pass

def create(self, request):
    pass

def retrieve(self, request, pk=None):
    pass

def update(self, request, pk=None):
    pass

def partial_update(self, request, pk=None):
    pass

def destroy(self, request, pk=None):
    pass
```

The API generally is a combiniation of:

- views.py
- serializer.py
- models.py

Also feel free to suggest changes to the model, in case it is necessary for the logic. They are not set in stone 😃. In terms of testing, you can use Swagger UI for manuall testing (backend endpoint: `/v1/schema/swagger-ui/`).

>[!NOTE]
> Please use an appropiate http [status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) for your responses.

### Contribution

I am happy to help with this issue.

to-sta:
Do you want to work on this issue @bharath637462 ?

andrewtavis:
Yes @bharath637462, let's keep it to one API for now :)

andrewtavis:
We can work on others as soon as some are finished.

bharath637462:
yes @to-sta

andrewtavis:
Let's please finish the work for the current frontend issue before starting on this one as well, @bharath637462.

bharath637462:
@to-sta but event has already api written on it 
[...]

Issue #611, opened 2023-12-19
Move search component menu to the top right corner below medium breakpoint


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that would be great to have for the three dot kebab menu on search result components would be that it moves to the top right corner when the screen is at the small breakpoint or below (for mobile devices). These search components can be found on the [activist home page](https://activist.org/en/home/). The menu component that we need to move is [MenuSearchResult](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuSearchResult.vue). The change would need to happen in all of the [search result components](https://github.com/activist-org/activist/tree/main/frontend/components/card/search-result) where it is used. 

For example:

```
<MenuSearchResult searchResultType="event" />  # we want this to have class="hidden md:block" to show it

# We'd then have another MenuSearchResult component that is at the top right
<MenuSearchResult class="absolute md:hidden" searchResultType="event" />  # including some padding
```

### Contribution

Happy to support someone with interest in working on this!

andrewtavis:
An example of how the final components should look can be found on the [designs in Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=1955%3A1111&mode=design&t=EzhMhrFeHSEWeQVJ-1) 😊

bharath637462:
hey can i work on this?

andrewtavis:
Thank you, @bharath637462! Please let me know if you need any help 😊 Looking forward to working with you on this!

andrewtavis:
Note that for the `MenuSearchResult`, we need the `searchResultType` property to match the search result like event for[CardSearchResultEvent](https://github.com/activist-org/activist/blob/main/frontend/components/card/search-result/CardSearchResultEvent.vue) and the same for organizations, resources and users :)

andrewtavis:
Hey @bharath637462 👋 How's everything going here? Were you able to bring down the current changes and start working on a new implementation? :)

bharath637462:
yes @Andrew McAllister ***@***.***> i am  able to take
the current changes working on new implementation.

andrewtavis:
[...]

PR #629, opened 2023-12-26
organisation api completed

bharath637462:
### Contributor checklist

---

### Related issue

- #627


andrewtavis:
@bharath637462 we need to you to follow directions on what the work you're supposed to be doing is. #627 is assigned to me. You wrote into the issue asking if you could work on it, and you were not given permission to work on it. Please do not do any other backend API work at this time, and please focus on completing the work for #611. Once that is done you can begin working on #628 or another API issue once we have assigned the issue to you.

@to-sta and I will review this, but understand that further PRs like this where you haven't been assigned will be closed.

andrewtavis:
Do not send along further commits to this issue. We'll finalize it ourselves.

to-sta:
@andrewtavis 

<ins>Workflows:</ins>
the pr_ci_backend check fail do to a simple linting issue and the mypy check fails because we are overwritting the ViewSet methods with a subclass of the Superclass. 

Here is an example:
```python
backend/entities/views.py:48: note:      Superclass:
backend/entities/views.py:48: note:          def create(self, request: Request, *args: Any, **kwargs: Any) -> Response
backend/entities/views.py:48: note:      Subclass:
backend/entities/views.py:48: note:          def create(self, request: Request) -> Response
```
Since we do not use *args and **kwargs we are overwritting the Superclass. I would say we ignore that issue and add a top level comment again with:

```python
# mypy: disable-error-code="override"
```
<ins>API:</ins>
Since we need to have an `Organization` instance before we can create a `organization_application`, I would suggest to add creating the `organization_application` within the OrganizationViewSet's create method or overriding the Organization model save method. 

The workflow for creating a Organization would be:

`https://activist.org/de/organizations/create/ `-> this would use the `OrganizationViewSet `create method (POST) to create an Organization if the user is authenticated. If it is successfully created, we would also create a organization_application instance with the default status. Here we need to have the organization_application_status before we can create an organization_application. We can create a fixture for this, that we load in with `docker-compose.yaml`.

Should we add a unqiue constrant to the name field of Organization model?

Other then that the create method looks fine to me!

andrewtavis:
I'll check this out a bit later, @to-sta. Wanted to remove the frontend changes so we're fresh :) Thanks for the details you've sent along!

to-sta:
@andrewtavis I added the models for status as discussed. I added `unique=True` to the org. name. I was testing the endpoints with swagger ui and saw that i am able to create duplicates. 

I made some changes to the serializer, now the `deletion_date` will be excluded from responses. We can get back to this when we work on the destroy method. created_by is set on the created method by the authenticated user, therefore the field should be a read_only field. `status_updated` and `acceptance_date` that should be set in the backend and therefore are also read_only. 

```python
class OrganizationSerializer(serializers.ModelSerializer[Organization]):
    class Meta:
        model = Organization
        exclude = ["deletion_date"]
        extra_kwargs = {
            "created_by": {"read_only": True},
            "social_accounts": {"required": False},
            "status_updated": {"read_only": True},
            "acceptance_date": {"read_only": True},
        }
```

to-sta:
I additionally removed the logic from update, partial_update and destroy, since these methods still need to be properly implemented. Since these 3 methods only have pass instead of a return value the ci_backend workflow is failing.

to-sta:
Thinking about the destroy method of the org API. We have a `deletion_date` field on the organization model. What logic should I implement here:

Keep the org after deletion (any fields that should be removed?), set a `deletion_date` (and `status_updated`) as well as assign a new `status`?

to-sta (approved):

andrewtavis:
@to-sta, sorry for the broken backend run. I don't think that any of the changes that came through in this would be problematic, but maybe it's because there are parts of the code in here now that aren't accounted for in this PR as it's coming from other commits? Can you do a quick fix of the backend errors and I'll bring this in?

andrewtavis (approved):
Thank you for all the hard work here, @to-sta! (also for the quick fixes just now 😅) So happy to have this finalized 🚀

Issue #596, opened 2023-12-03
Focus state of activist logo in footer is broken


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

There's an issue with how the `LogoActivist` components are being rendered in [FooterFlex](https://github.com/activist-org/activist/blob/main/frontend/components/footer/flex/FooterFlex.vue) and [FooterFlexCol](https://github.com/activist-org/activist/blob/main/frontend/components/footer/flex/FooterFlexCol.vue). When trying to focus the logo the `focus-brand` styling that's inherent within the [LogoActivist component](https://github.com/activist-org/activist/blob/main/frontend/components/logo/LogoActivist.vue) isn't being displayed. This is being caused by `overflow-y-hidden` being applied to the parent `<div>`, which in turn was implemented to solve an issue where the logo was expanding beyond its bounds on mobile Safari in 0d7485e

brunopinhal:
I would like to work on this, could you assign to me?

krallowski:
Hi there :)
What about setting box shadow inside for this specific item with box-shadow: inset ?

andrewtavis:
Hey @krallowski 👋 If you wanted to give it a shot and we can see if this fixes it it'd be very appreciated! Let me know if you'd have interest and I can assign you :)

krallowski:
Hey :)
Yeah, why not, I will try, thanks!

andrewtavis:
Thanks @krallowski! Let me know if there's anything I can do to help 😊

andrewtavis:
Closed via #630 😊 Happy to send along some more issues that need to be worked on if that'd be of interest, @krallowski. Let me know :)

PR #630, opened 2023-12-26
Missing logo focus

krallowski:
### Contributor checklist

---

### Description

The class focus-inside was added to the div of th activist logo in th footer. It adds the focus to the div fo :focus-within to avoid that the visible focus of the logo inside is partly hidden

Also, please describe shortly how you tested that your change actually works.
--> The change was tsted in Firefox, not for mobile

### Related issue

- #596


andrewtavis:
Thanks for this, @krallowski! We also need the same custom class within `FooterFlexCol` so it works on mobile. Do you want to send that change along?

krallowski:
Thanks @andrewtavis for reviewing and the hint. Added and pushed again :)

andrewtavis (approved):
Thanks so much for this, @krallowski! Really appreciate the quick commit after the first review as well 😊 By all means let me know if you'd be interested in working on another issue!

Issue #605, opened 2023-12-09
Favicon isn't showing up on some browsers


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

I'm realizing that the favicon isn't showing up on some browsers. Specifically on Safari and on Brave we're not getting the favicon showing up, but instead it's just the default `A` or `Ac` depending on the browser.

Would be great if someone wanted to look into this! The file is [frontend/public/favicon.svg](https://github.com/activist-org/activist/blob/main/frontend/public/favicon.svg), and the changes may need to take place in [frontend/nuxt.config](https://github.com/activist-org/activist/blob/main/frontend/nuxt.config.ts#L191).

Happy to support someone on this! Can also look into it myself a bit later :)

Jonath-z:
I can look into it

Jonath-z:
But I can see the favicon
![Screenshot from 2023-12-09 21-31-29](https://github.com/activist-org/activist/assets/79799120/0ba3bc0a-bbcb-4d44-ae35-7aa3f0b24057)

andrewtavis:
Thanks @Jonath-z! Glad to have the support for this :) What browser are you using that you can see it? For Safari I'm not seeing it on a Mac, and I wasn't seeing it for Brave on an iPad.

Jonath-z:
@andrewtavis I can see in firefox and brave as well, I'm using Ubuntu linux 23.10

- Brave
![Screenshot from 2023-12-10 06-37-53](https://github.com/activist-org/activist/assets/79799120/663eb6eb-83df-44da-8d43-75bc49f5d882)

- Firefox
![Screenshot from 2023-12-10 06-37-24](https://github.com/activist-org/activist/assets/79799120/e4ea3013-2cc4-4d6e-8f58-fb0e832d9dae)

andrewtavis:
Hmmm 🤔 Let me check and maybe this isn't an issue we can deal with. Could be just a Brave for iPad and general Safari thing, although each had favicons from other sources.

andrewtavis:
[...]

PR #632, opened 2023-12-27
fix: favicon not showing in safari

Roshanjossey:
### Contributor checklist

---

### Description

add favicon as png so that safari uses it as favicon.
Tested on safari 16.4

#### Screenshot
image

### Related issue

- #605


andrewtavis (approved):
Thanks so much for this, @Roshanjossey! You've sent us so many contributors that it only makes sense that you get a contributors credit as well 😊 Really appreciate you checking in with us for this :)

Just checked via the deployment preview and it is working on Safari 🎉 One thing that I did notice was that it's dark mode only as of now, but I'll merge and see if I can figure out how to target light mode to a black icon.

Thanks again!

PR #636, opened 2023-12-28
Feature/add image table

rigojr:
### Contributor checklist

---

### Description

Adds the image table, alongside its serializer and some FK in different models.

### Related issue

- [#ISSUE_NUMBER](https://github.com/activist-org/activist/issues/620)


rigojr:
Howdy, @andrewtavis I have some dubs about this task
- I added the image table as an FK in Users, Organization, Group, and Event in one of them, the Event, I could add a validation in the serializer, but in the others, I'm not sure if I have to.
- There are `about_images` properties in Group and Organization, I could not add an array of FK.
- I added the `ImageSerializer`, However, I added nothing in the `validate` method, not sure what validation we should perform over there.

I don't know if you or @to-sta can point me in the right direction, let me know what you think, thank you.

PD: I suffered using Figma, 😭 I don't know if I have to add more FK can you let me know how I can easily check with model is pointing to `Image`?

andrewtavis:
Thanks, @rigojr! I'll check this more tomorrow, but I can definitely help with what should go where! Makes sense that the Figma was a bit hard to follow as this table in particular has lots of things connected at one node.

to-sta:
I think I missed the issue @andrewtavis 😅 even tho you tagged me. 

@rigojr @andrewtavis I think we should use an `ImageField` for `file_location` instead of `CharField`. The `ImageField` takes care of saving the file and then stores the path in the database. 

```python
# thinking about file_location, we could also be a bit more specific 
image_location = ImageField(uploaded_to="images/")
```

Validation is bigger topic here since we have to think about file extensions, file size and duplicate names. And we can handle this on the model or/and the serializer. I would suggest we make validation a seperate issue. 

@andrewtavis 
We can configure the location with the MEDIA_ROOT in settings.py later on (for production). I think by default this will be the root of the django project. That should the job for now.

>There are about_images properties in Group and Organization, I could not add an array of FK.

Yeah, an ArrayField can not be used with relationship to another model. 

@andrewtavis in case we do not need additional data like the `creation_date` for the image, I would say let's get rid of the table and use an `ArrayField(ImageField())` instead. That would simplify the models and the schema. In case you want to keep a seperate table we have to use a different relationship here, like a `ManyToManyField`.

rigojr:
Hey Guys, I just updated this PR with the `image_field` advice from @to-sta. I am not sure what to do about the `about_images`.

andrewtavis (approved):
Looking at it together with @to-sta for Code Brunch 🍞☕, @rigojr. Thanks for all of this! Looks great to us 😊 Note that we've made a new issue #670 to look into validation steps for images. Maybe you have some ideas for that!

Issue #613, opened 2023-12-23
Disable disappearing header on mobile


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This is quite a simple issue and great for someone who wants to do a first issue for activist. In a recent product sync we decided that we do not want the header to disappear on mobile as it does for desktop. The disappearing header can be seen on the landing page at [activist.org](https://activist.org), and the mobile header can be seen in application pages like [activist.org/home](https://activist.org/home) beneath the medium breakpoint.

As we have the page navigation dropdown for mobile, it no longer makes sense for the header to disappear. The change would need to happen within the [frontend/components/header/HeaderAppPage](https://github.com/activist-org/activist/blob/main/frontend/components/header/HeaderAppPage.vue) component, with the goal being to find where the header is being hidden beneath the medium breakpoint and removing it.

### Contribution

Happy to support someone who wants to work on this! Let me know if there are any questions. Would be great if we could get to this soon 😊

VTPrince:
Hey @andrewtavis  , I would like to get started on this issue, can i open a PR for this?

andrewtavis:
Sure thing, @VTPrince! Assigning you now, and feel free to send along a PR :)

andrewtavis:
Hey @VTPrince 👋 @mikemumford sent along a PR for this and given the simplicity of the issue and wanting to get things closed I chose to merge it in. You can see the PR in #638. Thanks for your offer to help, and hope that we can find a new issue to work on soon 😊

PR #638, opened 2023-12-29
removed fading from mobile header

mikemumford:
### Contributor checklist

---

### Related issue

- #613

### DIFF
--- a/frontend/components/header/HeaderMobile.vue
+++ b/frontend/components/header/HeaderMobile.vue
@@ -1,12 +1,7 @@
 <template>
   <header
     ref="header"
-    class="relative sticky top-0 z-10 w-full h-12 md:hidden duration-500 drop-shadow-md"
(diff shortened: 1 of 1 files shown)


andrewtavis:
Hey there @mikemumford :) Normally would be good to write into the issue to see if we can re-assign, but also this is easy enough so there's no need for the stress. Looking at it, we also don't need the variables defined in the script tag to check scroll and opacity. The style tag is also related to this hide-show functionality, so that can also be removed 😊 I'm happy to do these changes myself, but wanted to give you the option of sending along a further commit if you wanted to. 

Thanks for the contribution!

andrewtavis (approved):
e81af52 finalized the code that needs to be removed, @mikemumford :) Thanks so much for the contribution here! We'll be making more issues in the future, so please let us know if others would be of interest 😊

Issue #580, opened 2023-11-25
Create ModalCommandPalette component base


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to create the baseline structure of the `command/ctrl + k` command palette, which would be a modal over the screen. Designs for this component can be found [here on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=7096%3A124636&mode=design&t=LosB9NvimjJgGZTZ-1), and can further be seen below:

![Jump To Overlay](https://github.com/activist-org/activist/assets/24387426/726a6e17-6dab-4756-b677-df3f9ef64c56)

This component would have links to various pages, which could be done now (although some links don't have their routes). The section below would be populated by organizations that the user is involved in or often visits.

This component would be put in the [frontend/components/modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal) directory, with the logic for the modal being copied from other components in that directory. A final part of this issue would be triggering this modal when the user presses `command/ctrl + k` :)

### Contribution

Happy to support someone on this or work on it myself eventually! This is not MVP functionality, but we can get started on the structure now and add the links in later 😊

andrewtavis:
Note that #582 should be finished before this, but this issue can also be worked on at the same time :)

mattburnett-repo:
writing in here

andrewtavis:
Assigned, @mattburnett-repo :) I'll let you know when the base component is merged in and then we can get started 😊

andrewtavis:
Hey @mattburnett-repo 👋 #582 is all closed, so we have the base component for you to work with now 😊 Let me know if you have any questions!

andrewtavis:
Reopening for now, @mattburnett-repo :) Quick things:

- We now have the links as `NuxtLink` components so we can use local paths
- Matched the icons with others on the platform
- Moved the styling out from the `tailwind.css` file (see below)

Final things that are needed here would be:
- It would be good if the styling was applied to two `v-for` calls where the organizations are generated by one and the options above are generated by the other
  - For the options above we'd need to save them in the `<script>` block as an array of some sorts with all the needed parts
[...]

PR #646, opened 2024-01-06
Fix: #580 Create ModalCommandPalette

mattburnett-repo:
### Contributor checklist

---

### Description

This PR addresses Issue #580. It creates a baseline for the component as described in the original issue.

It is only tested on MacOS. 

There are some TODO tasks:
   │  │  └─ ModalCommandPalette.vue
   │  │     ├─ line 18: TODO : Input area is much smaller that tag width.
   │  │     ├─ line 37: TODO : Sort out router-link target routes (the to="..." things).
   │  │     └─ line 170: TODO : Maybe there is a better / more future-proof solution than userAgent?

### Related issue
- #580


andrewtavis:
Thanks @mattburnett-repo! Do you want to check the errors reported above, or should I get to them?

andrewtavis:
> Error:   19:13  error  The element inside `<transition>` is expected to have a `v-if` or `v-show` directive  vue/require-toggle-inside-transition

I'd be happy to do the i18n ones on review :)

mattburnett-repo:
I made the errors. I'll fix them. Should have another commit / PR in maybe an hour.

mattburnett-repo:
fixed the i18n errors in the ModalCommandPalette code.

the remaining i18n_check error involves code outside of the code for ModalCommandPalette. could you fix that one, please?

frontend/components/modal/ModalCommandPalette.vue
```diff
@@ -0,0 +1,196 @@
...
+
+const { slash, meta_k, ctrl_k } = useMagicKeys({
+  passive: false,
+  onEventFired(e) {
+    if (["slash", "meta_k", "ctrl_k"].includes(e.key) && e.type === "keydown")
+      e.preventDefault();
+  },
+});
+
+whenever(slash, () => {
```
andrewtavis, line 178: Quick thing that I'm realizing here @mattburnett-repo is that I didn't do a good job of explaining that slash should maintain its old functionality for this :)

Ultimately activist will have two kinds of search: the slash one that's from the search bar that searches within the current domain for events or organizations or anything if we're not within a sub domain, and then the command palette that should only be triggered by `meta_k` or `ctrl_k`.

Would it be possible to:
- Remove the slash functionality from this
- Uncomment those lines in the search bar to show the command palette shortcut indicators

I'll then go through and do some styling. After that we should be good!

frontend/components/modal/ModalCommandPalette.vue
```diff
@@ -0,0 +1,196 @@
...
+      </router-link>
+    </template>
+  </ModalBase>
+</template>
+
+<script setup>
+import { useMagicKeys, whenever } from "@vueuse/core";
+
+// TODO: Maybe there is a better / more future-proof solution than userAgent?
+const isMacOS = /Mac/.test(navigator.userAgent);
```
andrewtavis, line 168: You can use the device check module as is done [here](https://github.com/activist-org/activist/blob/cb57435f2fbb3e4262ca18e8b0404923d624526f/frontend/components/SearchBar.vue#L37C17-L37C17) 😊 These are also the lines we want to be uncommented :)

andrewtavis: This is as good of a solution as I could think of, and would be fine for now until we realize that it has major drawbacks.

frontend/components/modal/ModalCommandPalette.vue
mattburnett-repo: No worries @andrewtavis . We had earlier some discussion about overall CommandPalette functionality and I probably got confused during that discussion. 

I've implemented your suggestions and they will appear in the next PR.

frontend/components/modal/ModalCommandPalette.vue
mattburnett-repo: Got it. The $device.isMacOS (from the template example) doesn't work in a script block. It took me a minute, but I figured out to use useDevice().

andrewtavis: Awesome! Great to hear, @mattburnett-repo!

frontend/components/modal/ModalCommandPalette.vue
andrewtavis: Looking forward to checking out the finalized version later, @mattburnett-repo! Thanks for the quick reaction :) :)

andrewtavis:
Note to self, do not do Headwind formatting for the Tailwind classes while reviewing as it'll cause crazy merge conflicts 😅

andrewtavis (approved):
First version is done, @mattburnett-repo! This is really amazing stuff here 😊 There are still some minor things to do that I'll mention in the issue, but I think we're good to merge this in :)

Issue #617, opened 2023-12-24
Documentation: add directions for getting dev env up without Docker


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

From a [discussion on Matrix](https://matrix.to/#/!CRgLpGeOBNwxYCtqmK:matrix.org/$0MkXg1BqAZXzpiA7hEirZ9uEWae9H8bkgu4oZTg_nLY?via=matrix.org&via=acter.global&via=chat.0x7cd.xyz) :)

A common issue that people have been having is that Docker doesn't work as well on certain machines. Specifically Windows computers at certain levels need to run a Linux virtual machine in order to get Docker up and running, basically making the process to expensive for it to function.

Our suggestions to community members till now have been to split the development environment and just run the frontend via Yarn or the backend with Django. We should include directions for this in the [Development environment section of the contributing guide](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#development-environment-) and also direct users to this section in the note to contact the team on Matrix in the [Environment setup section of the readme](https://github.com/activist-org/activist?tab=readme-ov-file#environment-setup-). 

Re the original discussion, I don't think we need docker compose specific processes for the frontend or backend as those won't address Windows users, and I don't think we need two processes for getting the frontend (Yarn + docker compose) and backend (Django + docker compose) respectively. 

Happy to discuss and help someone implement!

rstfari:
Hey, seeing if I could help!

andrewtavis:
You'd be welcome to help, @rstfari! Let me know if you have any questions about what to put in there. Happy to discuss here first, or also send along a PR and we can discuss in there 😊

rstfari:
Hey, so before I update the documentation, I just wanted to make sure that I am writing a step-by-step guide for the split development environment. Also, I am not familiar with docker yet but wanted to know where I could learn the steps so I could write the documentation correctly.

andrewtavis:
CC @to-sta: can you check the steps for the backend and verify that this is what we'd be looking to add?

andrewtavis:
Note that someone can set their ports differently in the `.env` file, but let's maybe ignore that for this as if they change it they should know that they did that? Also this is related to #570 where these steps might change.

[...]

PR #647, opened 2024-01-07
added directions on split dev environment

rstfari:
### Contributor checklist

---

### Description

Add documentation on splitting front and back-end environments. ### Using Yarn or Python section added
to CONTRIBUTING.md at line 200. Link to this section added to README..md at line 192.

Made sure the link properly moves user to correct section.

### Related issue

- #617

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -229,7 +229,7 @@ DATABASE_HOST=db
 DATABASE_HOST=localhost
 ```
 
-Now we can run the server in another terminal with:
+Now you can open up a new terminal to apply database migrations and start the local server:
 
 > [!NOTE]
 > Make sure to install the dependencies first. Best practice is to create a virtual enviroment first and then install the dependencies. Our dependencies are inside the `requirements.txt` file and can be install via `pip install -r requirements-dev.txt`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -214,13 +214,35 @@ You can then visit http://localhost:3000/ to see the development frontend build 
 
 **Backend: Python**
 
+Our backend depends on a connection to a postgres DB, therefore we need to setup the database first. Here our best opition is to still use docker to create a postgres DB with the following command:
+
+```bash
+docker compose up db
+```
+
+In order to connect to the DB, we need to change the `DATABASE_HOST` environment variable inside the `.env` file first.
+
+```bash
+# Current
+DATABASE_HOST=db
+# Changed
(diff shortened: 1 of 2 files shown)


andrewtavis:
Thanks for this, @rstfari! @to-sta, do you want to send along some edits for how you'd want the backend to read? 😊

to-sta:
This is a bit more complicated, since we can not run the backend without a postgres database. We use a postgres specific field (`ArrayField`) in our models what prevents us from using a `sqllite` DB. 

That means we still need to create a postgres DB.

Generally I think there are 2 ways of reducing the loading time:

1. Only run the database and the backend
    ```
    docker compose up --build db backend
    ``` 
2. Use Docker for the Database and then run the local server. 

    Here we need to adjust the .env DATABASE_HOST variable first. 
    ```
    # Current
    DATABASE_HOST=db
    # Changed
    DATABASE_HOST=localhost
    ```
    Then we can start the server:
    ```bash
    # In the root activist directory:
    cd backend
    python manage.py runserver
    ```

I can add the second option to the docs. What you think @andrewtavis?

andrewtavis:
Is there any way to get it up without Docker, @to-sta? Remember that for Windows users even that might be too much 🤔 But then if memory serves me it seems to be finishing for the db and it's mostly the frontend that can't finish. If there's no way to get it without Docker then by all means send along the second method! We can always iterate as needed :)

to-sta:
In order to run the backend without docker we need to make the backend models database agnostic with relationsships instead of postgres specific fields. 

This would mean, we have to change the schema, all models that use the ArrayField and create a local.py settings file that uses a sqllite DB instead of the postgres DB.

Would be a bigger issue to work on. 

I will add option 2 for now and we can discuss the changes necessary to accomplish that in a dev sync?

to-sta:
Thinking about this again. We also need to make migrations and then migrate before we can run the server. I think this might be a bit too complicated. If we migrate we are also creating migration files that should be excluded inside the `.gitignore` file.

```bash
# In the root activist directory:
cd backend
python manage.py makemigrations
python manage.py migrate
python manage.py runserver
```

andrewtavis:
Option 2 sounds like the correct route to me then :) Thanks for the further explanation, @to-sta! Good to have it documented in here as well. Let's do option 2 and we can react to it if it's still problematic!

to-sta (approved):

andrewtavis (approved):
All looks great to me 😊 Thanks for keeping at this, @rstfari! And thanks @to-sta for finalizing the backend directions :)

PR #648, opened 2024-01-08
feat: A11y enhancement to cardTopicSelection

UnknownSean8:
### Contributor checklist

---

### Description

- Tabbing from input to topics
- Arrow key control between topics

### Related issue

- #527


frontend/components/form/checkbox/FormCheckbox.vue
```diff
@@ -2,7 +2,7 @@
   <div class="relative flex flex-row items-center justify-start">
     <input
       :id="uuid"
-      class="cursor-pointer mb-0 peer appearance-none w-[1.375rem] h-[1.375rem] border border-light-menu-selection rounded-sm bg-light-button dark:bg-dark-button dark:border-dark-menu-selection focus-brand"
+      class="connect cursor-pointer mb-0 peer appearance-none w-[1.375rem] h-[1.375rem] border border-light-menu-selection rounded-sm bg-light-button dark:bg-dark-button dark:border-dark-menu-selection focus-brand"
```
andrewtavis, line 5: Would it make sense to apply this to an element on the page itself rather than a component as if there are say social accounts then they'll be on in between?

frontend/components/card/CardTopicSelection.vue
```diff
@@ -88,6 +101,36 @@ const moreOptionsShown = ref(false);
 const inputFocus = ref(false);
 const emit = defineEmits(["update:modelValue"]);
 
+const inputTabbing = (e: KeyboardEvent) => {
+  const firstTopic: HTMLElement | null = document.querySelector(".topic");
+
+  firstTopic?.focus();
```
andrewtavis, line 107: Is it possible to get the first topic being focussed even if we're tabbing back from the element after CardTopicSelection? As of now we're getting the last element, and I think consistency would be good,

andrewtavis:
Random idea, pinging both @UnknownSean8 and @TeddyGavi: should we be doing a grid here somehow such that if I'm on the first `ShieldTopic` of the first line that I could press down and get to the first element on the second line? I'm asking as it could be more user friendly, but beyond this it would mean that we'd have the same navigation for mobile and desktop where the difference would be that for mobile we would have another column to choose from.

andrewtavis:
@UnknownSean8, big thing is it would be great if we could focus the next element after one is selected with return. As of now if you select one the focus is lost and you need to tab out of the component and back in in order to make another selection.

frontend/components/card/CardTopicSelection.vue
```diff
@@ -9,18 +9,22 @@
...
+        v-for="(t, index) of filteredTopics"
         @click="selectTopic(t)"
         @keydown.enter.prevent="selectTopic(t)"
         :key="t.value"
         :topic="t.label"
-        class="max-sm:w-full"
+        class="topic max-sm:w-full"
         :active="isActiveTopic(t.value)"
         :isSelector="true"
+        @keydown.tab.prevent="topicTabbing($event)"
```
andrewtavis, line 25: Would you be able to fix the errors for attribute errors, @UnknownSean8, or do you want me to work on this? :)

UnknownSean8: It's okay, I can fix it over the evening. Thanks! :D

andrewtavis: Thanks @UnknownSean8!

frontend/components/form/checkbox/FormCheckbox.vue
UnknownSean8: I understand what you mean, I would still need to research whether there's another way to implement this, as the current one doesn't seem to be a good practice. I'll go ahead and start with your direction on the element rather than the component and see how it goes.

andrewtavis:
Hey @UnknownSean8 👋 To me this is working perfectly for desktop as far as tabbing into it, tabbing back to it and using left and right to navigate 😊 Do you want me to merge this and we can do a separate PR for up and down, or do you want to continue to commit to here?

UnknownSean8:
Hi @andrewtavis, for the tabbing into the topics, I was thinking whether to always tab to the first topic or tab to the first unselected topic.

As for the merge, I don't mind finishing the up and down before merging. But the final decision is up to you.

Thanks!

andrewtavis:
Let's keep it as it is right now as far as into the first topic as in some cases we'll pre-select topics for people that they may want to deselect (say for an event for an organization that works on three topics and they want to deselect one of them for an event as it's not applicable). And I'd be happy to do a full review and discussion once it's all finished 😊 Feel free to send along commits to this one and we'll finalize and merge this in! 🚀

UnknownSean8:
IMO, I think the calculation for the topic filtering when going up and down can be more accurate. Also, I think I can merge the key down into one instead of separating it into multiple ones in the current commit. Let me know what you think, @andrewtavis.

Thanks!

andrewtavis:
Thanks @UnknownSean8! Stoked to check it out later :)

andrewtavis:
If you think you can merge the key down functionality, then it sounds good to me @UnknownSean8 :) Big thing is that I wasn't able to tab into the options on mobile to try the up and down arrow commands. If I clicked one of the options then I could tab between the topics, but then that's also not what we're after. More that it works the same as left and right where we tab to or back to the first.

UnknownSean8:
> If you think you can merge the key down functionality, then it sounds good to me @UnknownSean8 :) Big thing is that I wasn't able to tab into the options on mobile to try the up and down arrow commands. If I clicked one of the options then I could tab between the topics, but then that's also not what we're after. More that it works the same as left and right where we tab to or back to the first.

Oops, sorry, I forgot to include it for the mobile view. However, I encountered a problem where I wasn't able to tab into the topics because I couldn't detect if the application was in mobile or desktop view.

andrewtavis:
Just remembered something, @UnknownSean8: anther reason why the mobile version of this and its a11y is so important is that the mobile version is being used as a filter in the left sidebar :)

UnknownSean8:
Ahhh, okay thanks. I'll keep it in mind and look into it tonight. Thanks @andrewtavis

UnknownSean8:
It's fixed now and working as intended for both mobile and desktop views. I just realize I don't understand what you mean by being used as a filter in the left sidebar. 😵‍💫

andrewtavis:
Looking forward to the review, @UnknownSean8! What I mean is say we're searching for events and we have a list or a map of them displayed. The mobile version of this component will also serve as the filter in the left sidebar for allowing the user to select which topics to filter by :)

andrewtavis (approved):
Hey @UnknownSean8 👋😊 Checking this and it's all looking great to me! 🚀 Thanks so much for the hard work here. Let's check in in the issue and then we can close this up :)

UnknownSean8:
Hi @andrewtavis, thanks! Regarding your previous comment about using this as the side bar item, I just realize there's a problem on my implementation as the current "tabbing" implementation is hardcoded to the next component.

I'll need to fix it up, sorry for the inconvenience caused. :P

andrewtavis:
Hardcoded to the next component being it always goes to the checkbox, or will go to the next component in general? If it's the latter then all's fine :)

UnknownSean8:
Unfortunately, it's the former one. I'll see what I can do about it and check if it's possible to do the latter one.

Issue #634, opened 2023-12-28
Close search result menu tooltip when clicking the button


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

After recent commits the functionality of [MenuSearchResult](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuSearchResult.vue) is almost where it should be.  This functionality can be seen on the [home page](https://activist.org/home/). We have:

- The ability to open the tooltip by clicking the button
- The ability to close the tooltip by clicking outside of it (important for mobile)
- The ability to close the tooltip by moving the cursor away from the button and tooltip (important for desktop)

The issue is that the above conditions are conflicting and we actually can't close the tooltip by clicking the button again anymore (my assumption is that both the first and third options above fire and we just get it again).

Any support on this would be appreciated! Happy to help as well :)

arescrimson:
Hi! I can work on this if you want :))

andrewtavis:
Hey @arescrimson 👋 Thanks for your interest in activist and wanting to help! We'd be happy to get some support here! I'll assign you now, and please let me know if you have any questions!

arescrimson:
So this is actually more tricky than I thought itd be: 

Basically, your assumption is correct: when I disable the onClickOutside function, you can then click on the menu itself and it will toggle and and off, as intended. However, it looks like the onClickOutside function is actually firing when you press on the menu item itself, which then results in showToolTip being toggled to false when menu is clicked(unintended) but then being toggled to opposite of showToolTip(true) by the menu(intended), which then results in the tooltip not closing because showToolTip is still true.

![image](https://github.com/activist-org/activist/assets/66581240/ee80f751-c389-4954-bd2e-af3e4bbceb07)

For now, I'm still not too sure, because it looks like onClickOutside is a Vue function, and I'm still unclear on why it is firing when I am clicking on the menu component itself. Currently I'm working on potentially adding another ref that toggles when the menu component itself is called, and then blocks the onClickOutside function from firing that way, but other ways are also welcome. If anyone has any ideas, let me know :)))

andrewtavis:
[...]

PR #649, opened 2024-01-09
Fix: Menu tooltip close fix

arescrimson:
### Contributor checklist

---

### Description

This pull request fixed the MenuSearchResult tooltip functionality. Previously, it would not close on menu button click due to an `onClickOutside` event firing, therefore preventing `showTooltip` from being set to false in the menu button click. 

This was solved by reassigning  the `tooltip` reference to the `div` of the menu button and tooltip components. This will allow the `onClickOutside` function to correctly handle the visibility of these components appearing. 

I additionally added a `toggleToolTip` function for future readability/debugging purposes. Finally, I removed the `tooltip` references from the TooltipMenuSearchResult components themselves, as the `tooltip` reference is now handled by the parent div. 

I tested this by logging the value of the `showTooltip`, and it now correctly toggles between true/false on every click, instead of being constantly reset to true as it was before. This allows users to also click the menu buttons again to close the tooltip. 

Fixed showTooltip updating properly: 

image

### Related issue

- #634

### DIFF
--- a/frontend/components/menu/MenuSearchResult.vue
+++ b/frontend/components/menu/MenuSearchResult.vue
@@ -1,7 +1,7 @@
 <template>
-  <div @mouseleave="showTooltip = false" class="p-2">
+  <div @mouseleave="showTooltip = false" class="p-2" ref="tooltip">
     <button
-      @click="showTooltip = !showTooltip"
+      @click="toggleTooltip"
       @keydown.shift.tab="onShiftTab"
       class="relative flex items-center justify-center w-8 h-8 rounded-full style-cta md:w-6 md:h-6 elem-shadow-sm"
     >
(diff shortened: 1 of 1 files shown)


arescrimson:
Not really sure why the prettier is failing, but after going back and reformatting the document, it looks like prettier removed a trailing whitespace after the `const tooltip = ref()`, perhaps that's it? I could do a hotfix push right now to confirm, but would welcome any input as to why prettier is failing.

andrewtavis:
Looks ok now, @arescrimson :) Thanks for the hot fix!

andrewtavis (approved):
Really great, @arescrimson! I'll go ahead and do a minor edit to the ref so that we have something more representative that includes the button as well. Will write in the issue and close it once it's all done 😊

Issue #643, opened 2024-01-05
Add Vue and Vite to package.json to silence frontend warnings


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

A consistent question that we get from new developers in the community is where there are so many errors in the frontend when they start Docker. One such error is:

```
warning "nuxt-icon > @nuxt/devtools-kit@1.0.6" has unmet peer dependency "vite@*".
```

I believe this is happening as we don't technically have Vue or Vite in the [package.json](https://github.com/activist-org/activist/blob/main/frontend/package.json) file as they're being added by Nuxt. Adding these in would potentially silence the errors and improve the developer experience.

Thoughts on this and an implementation welcome are!

andrewtavis:
CC @ahmedy00 and @rigojr 👋 Would be nice to get either of your two's opinions on this!

arescrimson:
Hi! I saw this issue and was wondering about it; looks like the error messages new developers like me are receiving looks like this(with a few more lines of errors): 
![image](https://github.com/activist-org/activist/assets/66581240/b0d98038-dc6d-4450-97d0-4cbb60cb8e79)
With this, I noticed that activist is currently using nuxt version 3.8.2, so I looked at what versions of vite and vue is specified: 
![image](https://github.com/activist-org/activist/assets/66581240/1c7b3656-503f-4433-84f5-71fb71fd2464)
I then added these specific versions in to the package.json, and restarted the frontend. This time with the build, these errors were no longer visible: 
![image](https://github.com/activist-org/activist/assets/66581240/4349e4ec-8643-4382-923f-3afafa0878ce)
However, I'm still not too sure if this is the correct way to go about this, but I'm very welcome to more experienced opinions for any guidance/confirmation. The vue/vite versions seem to be correct, but I have not tested this extensively with the new specified packages yet.

arescrimson:
Investigating more, it looks like the Vue version is actually 3.4.5, and the Vite version is 5.0.11 for Activist Frontend usage. 

![image](https://github.com/activist-org/activist/assets/66581240/f542b1cb-5d24-4fe5-b5a5-6db8182c62b6)

Changing package.json values to these versions seems to have the same result; the errors are still not showing:

[...]

PR #653, opened 2024-01-15
Fix: Updated package.json inclusion of Vite and Vue

arescrimson:
Updating Vite and Vue versions here will prevent frontend errors. Vue and Vite versions confirmed through project Vite and Vue versions.

### Contributor checklist

---

### Description

This change is to resolve issue #643, where contributing developers, upon docker container startup, would see many error logs from unmet peer dependencies. This was because Vite/Vue versions were not specified in the `package.json` file. 

This was fixed by first checking Vite/Vue versions that Activist is currently using. 

image

Before adding these in, these errors would be logged. 

image

After, this fix removes the aforementioned errors from logging. 

image

This PR will also not include `yarn.lock` - that should update on its own. Let me know if anything needs to be changed, however ;DD

### Related issue

- #643


arescrimson:
Interesting; looks like TypeScript checks are failing on two files: `SidebarRight` and `MenuSearchResult`. @andrewtavis, any thoughts? We took a look at both quite recently, so was wondering what could have happened.

andrewtavis:
One thing that might be happening is that the types have been updated given the explicit inclusions of these dependencies. Not really sure though as I haven't touched these recently aside from maybe some localization key changes. Do you want to fix these, or should I? Totally fine either way! :)

arescrimson:
What I thought too; introducing new dependencies like this could have more impact outside of squashing these error logs, but I do find it interesting since we literally just worked on both of these files. Could look more into this, looks like some simple typescript changes so I'll see what I can do and test if it fixes it :)) Of course if you see what's wrong/ what needs be changed feel free to edit/let me know.

andrewtavis:
Thanks for looking into this, @arescrimson! The check should handle most of the changes, so I wouldn't worry too much about something outside of it. Ideally you would also see these as errors in the files now, and you should also see it locally by running `yarn nuxi prepare` and `yarn nuxi typecheck` in the `frontend` directory to prepare he TS types and check the frontend :)

arescrimson:
Cool! Tested out one final change and it looks like this one could work. Basically, TypeScript is expecting a object value here for `onClickOutside`, and we are providing it with `quickActionBtnAndMenu`. Since this is a reference to the `ref()` object, we get this type mismatch error. However, if we instead pass in `quickActionBtnAndMenu.value`, it looks like it fixes it, and will do so for the other file as well. Will confirm this in the morning!

andrewtavis:
Are we good for a final review here, @arescrimson?

arescrimson:
@andrewtavis Yep!

andrewtavis (approved):
Thanks for the research and the implementation, @arescrimson! Glad to have these errors figured out 😊 Just a few more and the dev experience will be where it needs to be :)

PR #654, opened 2024-01-16
Fix: Updated Rollup Dependency in package.json

arescrimson:
Updated rollup versions to be specified in package.json to suppress peer dependency errors.

### Contributor checklist

---

### Description

This change concerns a **Frontend** error raised during building, where this error shows. 

image

Because of this issue's relation with PR #653, we can include `rollup` in `package.json` to suppress this error. For reference, the error specifies that `rollup` version should be of `rollup@^1.20.0 || ^2.0.0 || ^3.0.0`. However, Activist is actually using `rollup` version `4.9.5`. 

image

However, I discovered that specifying version `4.9.5` did not resolve the issue. Therefore, I simply edited the `package.json` to specify the exact versions of `rollup` as in the error. By doing this, the errors no longer show. 

image

### Related issue

- #643


andrewtavis:
Quick question on this, @arescrimson: does it work if we do `"rollup": "^3.0.0"` so that we specify at least the latest full release?

arescrimson:
@andrewtavis Good thinking. Looks like it does; however note `"rollup": "^4.0.0" does not work. Will make that change.

andrewtavis (approved):
Thanks @arescrimson! Will send along the `yarn.lock` shortly and we'll be good to go :)

Issue #683, opened 2024-02-02
Creating a custom_settings file in the backend


kurtisolo:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

I usually create a `custom_settings.py` file in the backend directory of my projects, and then import everything in `settings.py` simply using `from backend.custom_settings import *`

In this way it's much easier to manage more granular settings. For example I'm working on issue #361 and it would be good to have in the `custom_settings.py` file different values for the caching.

Example: 

```
PURPOSE_1_CACHE_DURATION = 10 #minutes
PURPOSE_2_CACHE_DURATION = 30 #minutes
```

Then in the whole backend, when caching is necessary, we can always import and use the values from the settings. For example:

```
from django.core.cache import cache
from django.conf import settings

def my_view(request):
    data = cache.get('my_data')
    if not data:
        data = calculate_heavy_data()
        cache.set('my_data', data, timeout=settings.PURPOSE_1_CACHE_DURATION*60)  #600 seconds
    return data
```

It's very practical and scalable having more granular settings in a single file, for example we can also put there the number of the pagination elements, or the imports from the .env file.

I use this approach in big projects and saves me a lot of time once it's done. Also it helps a lot keeping the `settings.py` file more clean.

### Contribution

Happy to support someone on this issue or work on it myself (I can also put this together with issue #361 )

andrewtavis:
Hey @kurtisolo 👋 All sound good to me. Feel free to combine this with #361 :) I'll assign you now!

Issue #665, opened 2024-01-27
Rename `special-text` Tailwind classes to `distinct-text`


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In line with having a `content` background and a `distinct` background, it would be good if we have a normal `text` and then a `distinct-text` rather than `special-text`. This can be achieved via an all code search 😊

### Contribution

Happy to support someone with this or get to it myself!

andrewtavis:
Closed via 071a047 😊

Issue #427, opened 2023-10-08
Discussion of documentation of Vue components


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to discuss and potentially implement the first instances of documentation `<docs>` blocks within Vue single file components (SFCs). Information on custom blocks can be found [here](https://vuejs.org/api/sfc-spec.html#language-blocks), and info on docs blocks can be found [here](https://vue-loader.vuejs.org/guide/custom-blocks.html#example).

Would be great to get the community's feedback on this and if we'd like to put quick bits on usage in the `<docs>` blocks :)

### Contribution

Would be happy to discuss this with people! Implementing any needed settings changes and adding in the first few `<docs>` blocks could also be a great `good first issue` 😊

andrewtavis:
@ahmedy00, it'd be interesting to get your feedback on this as you've been working with Vue longer than most of us 😊 Have you used `<docs>` tags before?

ahmedy00:
@andrewtavis I haven't used the <docs> tag in any project before, so I don't have educated guess as to the benefits it can provide.
In my opinion it may seem beneficial for new contributors at first glance but it may cause code complexity, readability and compatibility problems in the long run.

andrewtavis:
Yes, I was thinking about also the need to maintain all of this 🤔 The whole point for Vue is that it has the structure of HTML as a basis and should be easy to understand given good naming criteria. We'll also need to do documentation eventually as well, which is where we'd send people 🤔 My one question is, would there be a way to auto document components? Have you heard of/used anything for that? Would be cool if we could generate something from the code and just make sure that we write proper comments where needed.

andrewtavis:
Thanks for the feedback!

andrewtavis:
Not sure if you'd have interest in #424, btw @ahmedy00. I know that you prefer `script-template-style`, but to me/the team as we're using Tailwind it really does make sense to have template first as we don't have style blocks. If you wanted to set up the linting for that it'd be a lot of help 😊

ahmedy00:
[...]

PR #686, opened 2024-02-04
added custom_settings.py

kurtisolo:
useful for having a single source of truth

### Contributor checklist

---

### Description

This change is intended to make it easier to modify specific settings without having to search for them throughout the entire project. It's always useful to have a single source of truth in such large projects.

### Related issue

- #683

### DIFF
--- a/backend/backend/custom_settings.py
+++ b/backend/backend/custom_settings.py
@@ -7,4 +7,3 @@ check the path: backend/backend/settings.py
 # Pagination settings
 PAGINATION_PAGE_SIZE = 20 
 PAGINATION_MAX_PAGE_SIZE = 100
-
--- a/backend/backend/custom_settings.py
+++ b/backend/backend/custom_settings.py
@@ -1,0 +1,10 @@
+'''
+This file contains granular settings specifically related to the Activists project.
+For more general settings strictly related to the functioning of Django and its components, 
+check the path: backend/backend/settings.py
+'''
+
+# Pagination settings
+PAGINATION_PAGE_SIZE = 20 
+PAGINATION_MAX_PAGE_SIZE = 100
+
--- a/backend/backend/paginator.py
+++ b/backend/backend/paginator.py
@@ -1,7 +1,8 @@
 from rest_framework import pagination
+from django.conf import settings
 
 
 class CustomPagination(pagination.PageNumberPagination):
-    page_size = 20
+    page_size = settings.PAGINATION_PAGE_SIZE
     page_size_query_param = "pageSize"
-    max_page_size = 100
+    max_page_size = settings.PAGINATION_MAX_PAGE_SIZE
--- a/backend/backend/paginator.py
+++ b/backend/backend/paginator.py
@@ -1,8 +1,9 @@
 from rest_framework import pagination
-from django.conf import settings
+
+from . import custom_settings
 
 
 class CustomPagination(pagination.PageNumberPagination):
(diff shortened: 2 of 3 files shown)


andrewtavis:
Thanks for the PR on this, @kurtisolo! Quick note: no stress on the frontend formatting checks. I'm realizing that one file needed to get formatted after the changes I made in #665. Do you want to look into the backend formatting, or should I take care of that? :)

kurtisolo:
@andrewtavis let me try to fix it =)

kurtisolo:
@andrewtavis two different solutions:

- using `from backend.custom_settings import *  # noqa: F403` to make ruff ignore this line

- directly importing `custom_settings.py` along the project instead of settings.py. For example instead of:
```
from django.conf import settings
from rest_framework import pagination

class CustomPagination(pagination.PageNumberPagination):
    page_size = settings.PAGINATION_PAGE_SIZE
    page_size_query_param = "pageSize"
    max_page_size = settings.PAGINATION_MAX_PAGE_SIZE
```
then we should use:

```
from . import custom_settings
from rest_framework import pagination

class CustomPagination(pagination.PageNumberPagination):
    page_size = custom_settings.PAGINATION_PAGE_SIZE
    page_size_query_param = "pageSize"
    max_page_size = custom_settings.PAGINATION_MAX_PAGE_SIZE
```

I think both solutions are ok, we just have to add a couple line for documentation. I'm also working on the Wiki for backend documentation, so everything will be documented there

andrewtavis:
I think it's ok to go the second route, @kurtisolo :) Makes it a bit more clear what the source of all this is. Thanks for the suggestions, and further for the work on the wiki! Really exciting to see what you're working on 😊

andrewtavis:
I'm thinking as well that maybe we could look into a typical documentation website for this such that it's updated from the code itself? If you thought that working out something from [Sphinx](https://www.sphinx-doc.org/en/master/) could help, then let me know. Maybe we could use that as a baseline and then also include component-wise frontend documentation from #427 🤔

kurtisolo:
@andrewtavis I've never used Sphinx, but it would definitely be a great opportunity to learn =) It would surely provide a solid and scalable foundation, so as to have comprehensive and high-quality documentation in the future.

andrewtavis:
I'd be happy to set up a baseline instance if that'd be of interest, @kurtisolo. We could deploy to [readthedocs](https://about.readthedocs.com/?ref=readthedocs.com) at first and then shift the hosting later on when we have a bit more capacity?

andrewtavis:
Heyo @kurtisolo: let me know what you think of [mkdocs-material](https://github.com/squidfunk/mkdocs-material) that's built on [mkdocs](https://www.mkdocs.org/) as a documentation solution. Not a Wiki, but serves the main purpose of being directly in the codebase. There's also Python autodoc functionality already from [mkdocstrings](https://mkdocstrings.github.io/usage/), with them also considering adding in TS support. The frontend docs would be mostly handled by the design system anyway, with us just needing to host and point people to the Histoire instance that has the examples.

backend/backend/custom_settings.py
```diff
@@ -0,0 +1,9 @@
+'''
+This file contains granular settings specifically related to the Activists project.
```
andrewtavis, line 2: Quick note on this, @kurtisolo: we always do `activist` in lower case :) No stress though! I can fix 😊

andrewtavis (approved):
Really great, @kurtisolo! Appreciate the attention to detail as always 😊

PR #693, opened 2024-02-10
clean secret key overwrite

to-sta:
### Contributor checklist

---

### Description

Just saw that we overwrite the secret_key settings in backend `settings.py`. There is no reason to keep that in here.

### Related issue

No related issue


andrewtavis (approved):
Thanks for the find, @to-sta!

Issue #703, opened 2024-02-15
Add missing lang attribute to html


lucianmurmurache:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

This is an accessibility (and seo) fix/improvement, currently the html element does not have a [lang attribute](https://www.w3.org/WAI/standards-guidelines/act/rules/b5c3f8/).

Nuxt seems to have a nice [plugin ](https://nuxt.com/docs/guide/directory-structure/plugins) feature where the attribute can simply be injected into each page by default, without having to add it manually to all <Head>...</Head> elements so I will opt for this approach.

andrewtavis:
Thanks for opening the issue, @lucianmurmurache! Will review the PR in the coming days :) :)

andrewtavis:
Will close this once I send along the period for the comment, @lucianmurmurache. Again no stress on that 😊

andrewtavis:
Sent it along c90b8a0 so we can close this, @lucianmurmurache :) Thanks again for all the work here!

PR #704, opened 2024-02-15
Add html lang attribute

lucianmurmurache:
### Contributor checklist

---

### Description
This addresses the missing lang attribute to the html element for all pages.
The attribute is injected via a plugin created under `/plugins/i18n-head.ts`

### Related issue

- #703


frontend/plugins/i18n-head.ts
```diff
@@ -0,0 +1,8 @@
+// Injects the lang attribute in all nuxt <Head></Head> components
```
andrewtavis, line 1: Minor nitpick that we normally do complete sentences for comments that are their own lines (a Python convention I just do these days for everything). No worries whatsoever on it though :) I can get this.

Issue #711, opened 2024-02-16
Sidebar right for mobile should close after option selection


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In conversations with @lucianmurmurache, SidebarRight is not closing when an option is selected. This can be seed from the mobile view on any page where the burger menu is used to open the sidebar, and then once a new localization, color mode or another option is selected the sidebar is still open. 

Suggestion for how to fix this would be:
> A ref and a listener for click outside should do the trick here

This is a good first issue for someone wanting to help, or a community member can also pick it up 😊

lucianmurmurache:
Happy to take care of it @andrewtavis 

There is another [issue ](https://github.com/activist-org/activist/issues/631)that could potentially reuse the same logic

lucianmurmurache:
@andrewtavis I am actually not 100% sure if the functionality would be sound from an UX perspective, do you know if there is any community member that specialises in UX to give me a little hint here? 

Considering that there are two menu items, which themselves are drop downs, I'm thinking that it might be less convenient if the user wants to change both language and theme, they would then have to open the menu twice. I can take care of the menu closure on click outside for now and leave this as a draft until we get some clarity 💡

andrewtavis:
I think clicking outside would be fine for now, @lucianmurmurache :) We don't have a specific UX person right now, but we can also chat about it in the next dev sync 😊

lucianmurmurache:
In that case, the PR is ready @andrewtavis 👍

andrewtavis:
a536fc6 and a minor change in 3e9aec5 do some final touches for this, @lucianmurmurache :) One thing I noticed post merge is that it was opening on any click outside, so even a random click on the page would open the sidebar 😅 We now have `closeMenuState` that just sets it to false that's called. There likely is doubtless a way of doing this, but I wanted to fix prod quickly (hotfix welcome as we're doubtless calling this TS function needlessly when the sidebar is closed). 

[...]

PR #712, opened 2024-02-16
fix: closes menu on click outside

lucianmurmurache:
### Contributor checklist

---

### Description
SidebarRight is not closing when an option is selected. This can be seed from the mobile view on any page where the burger menu is used to open the sidebar, and then once a new localization, color mode or another option is selected the sidebar is still open. 

❗ Should check UX flow if the desired behaviour above is correct. For now the fix only addresses click/press outside the menu

### Related issue

- #711

### DIFF
--- a/frontend/components/sidebar/right/SidebarRight.vue
+++ b/frontend/components/sidebar/right/SidebarRight.vue
@@ -29,20 +29,13 @@
 <script setup lang="ts">
 import { onClickOutside } from "@vueuse/core";
 
-const target = ref();
+const target = ref<HTMLElement | null>(null);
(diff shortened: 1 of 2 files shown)


andrewtavis:
What's your opinion on adding in an `X` to close the menu in the top right, @lucianmurmurache? All's working great, but in looking at it I feel that it's missing 🤔

andrewtavis:
I feel that having an actual close button would be important for the user experience :)

lucianmurmurache:
@andrewtavis great suggestion! Please take a look at the latest update, the sidebar now has a close button. Let me know if any changes are needed

Issue #669, opened 2024-01-27
Create route to organization vote via accepted icon


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now we can see the votes that an organization receives to get onto activist by setting the status of the organization to pending. We also need to add in a route to the organizations application, which would be within the `About` route, but would be visible if the user clicks the accepted icon. Designs for this page can be seen [in the designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2642%3A20194&mode=design&t=RdDHojCud6Kv6mvh-1). We can go in and edit the [about page for organizations](https://github.com/activist-org/activist/blob/main/frontend/pages/organizations/%5Bid%5D/about.vue) to then display this.

### Contribution

Happy to support someone who has interest in working on this 😊 This is a bit difficult for a first issue, but could be doable. The big thing is that we need to display the [CardOrgApplicationVote](https://github.com/activist-org/activist/blob/5d8dac27f4163a2cbd82033e4605ce93ae49695b/frontend/pages/organizations/%5Bid%5D/about.vue#L66) that's displayed on pending and also need to add in some of the [discussion components](https://github.com/activist-org/activist/tree/main/frontend/components/card/discussion).

andrewtavis:
CC @rigojr who has interest in the voting process :)

rigojr:
I could help with this, let's talk more about it to see the challenges over here.

andrewtavis:
You've got a lot on your plate right now, @rigojr, so let's not worry about this as much for now. Just wanted to make you aware of related work 😊 The scrollbar would definitely be the highest priority now that the images api is merged :)

UnknownSean8:
Hi, I could work on this if it's still available. Thanks.

andrewtavis:
Sounds good, @UnknownSean8! Happy to discuss this async, or we can also look into it in the sync 😊

UnknownSean8:
@andrewtavis, maybe we could discuss this async first as I am not so sure if I will be able to join the sync this weekend as something came up 😟. But I will still be able to work on it over the weekends 😃.

andrewtavis:
[...]

PR #713, opened 2024-02-19
Feat: route to organisation vote

UnknownSean8:
### Contributor checklist

---

### Description
Route to organization votes through the accepted icon. Just wondering if this is the feature that you're looking for?

For now, I am making changes to the IconOrganizationStatus, however, I noticed that there is another BtnRouteInternal that might be useful in this case. Is it better to use the current component now or make changes to the other one?

The hiding of the unneeded card from both statuses are not implemented yet.

Thanks.

### Related issue

- #669


andrewtavis:
Hey @UnknownSean8! 👋 Just checking, isn't it supposed to change the view when I click the green checkmark in a circle next to the organization name on the about page? Just tried the preview, and we now have the discussion elements on the organization about page as they should be when the application icon is pressed, but then nothing happens when I click the icon itself (I'd assume that the discussions would come to the top and the about card/images would go away).

Issue #722, opened 2024-02-26
Add reduce motion preference


lucianmurmurache:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### This is more of a missing accessibility feature but will report it as a bug
Users that have motion sickness and disable animations on their device, would expect not to see any animations on the platform. We can make use of the [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) feature which is widely supported, when applicable, we can use the [tailwind class](https://tailwindcss.com/docs/transition-property#prefers-reduced-motion) instead. 

### How to enable and test:
- Windows 10: Start > Settings > Ease of Access > Display > Toggle “Show Animations in Windows.”
- macOS: System Preferences > Accessibility > Display> Check “Reduce Motion.”
- iOS: Settings > Accessibility > Motion> Toggle “Reduce Motion.”
- Android: Settings > Accessibility > Toggle “Remove Animations.”

andrewtavis:
Awesome, @lucianmurmurache!! Thanks so much for this :) :)

andrewtavis:
065e7b8 adds some directions to the contributing guide for this, @lucianmurmurache 😊 Feel free to edit this section as you see fit! Thanks again for all of this consideration on your part :) :)

PR #723, opened 2024-02-26
style/add prefers reduced motion

lucianmurmurache:
### Contributor checklist

---

### Description
Users that have motion sickness and disable animations on their device, would expect not to see any animations on the platform. We can make use of the [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) feature which is widely supported, when applicable, we can use the [tailwind class](https://tailwindcss.com/docs/transition-property#prefers-reduced-motion) instead.
#### How to enable and test:
- Windows 10: Start > Settings > Ease of Access > Display > Toggle “Show Animations in Windows.”
- macOS: System Preferences > Accessibility > Display> Check “Reduce Motion.”
- iOS: Settings > Accessibility > Motion> Toggle “Reduce Motion.”
- Android: Settings > Accessibility > Toggle “Remove Animations.”

### Related issue

- #722


lucianmurmurache:
> Looks like we just need to format `tailwind.css` for the frontend test to pass, @lucianmurmurache 😊 Missing the blank last line in the file? Maybe something else, but that's what jumps out :)

For some reason prettier was not auto formatting on my machine, it should be good now. Thank you @andrewtavis

frontend/tailwind.config.ts
```diff
@@ -82,6 +82,9 @@ export default <Partial<Config>>{
           instagram: "rgba(129, 52, 175, 1)",
         },
       },
+      transitionProperty: {
```
lucianmurmurache, line 85: Just for a bit of clarity, I added this here as I noticed that `transition-width` was being used but it is not included by default. It now works as expected. 
Example: [PasswordStrength](https://github.com/activist-org/activist/blob/main/frontend/components/PasswordStrengthIndicator.vue#L7)

andrewtavis: Really great to have the attention to these sorts of things, @lucianmurmurache :) Thank you!

frontend/components/card/CardConnect.vue
```diff
@@ -5,7 +5,7 @@
         {{ $t("components._global.connect") }}
       </h3>
       <div
-        class="p-1 break-all rounded-lg cursor-pointer transition-all hover:text-light-highlight dark:transition-all dark:hover:text-dark-highlight"
+        class="p-1 break-all rounded-lg cursor-pointer transition-all hover:text-light-highlight motion-reduce:transition-none dark:hover:text-dark-highlight"
```
andrewtavis, line 8: Thanks for the fix here and the other place where we're really transitioning on dark mode, @lucianmurmurache! 😄

andrewtavis (approved):
All looks fantastic on my end, @lucianmurmurache! I'll add a quick note on this to the contribution guidelines and we can close out the new issue! 🚀

Issue #729, opened 2024-02-27
Add animation to sidebar icon


lucianmurmurache:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Currently we are replacing the icon based on the open/close state. A slightly better approach would be to rotate the icon on open/close instead, this will avoid the icon re-render while at the same time, add a little animation.

andrewtavis:
Thanks for the issue, @lucianmurmurache! 😊

andrewtavis:
30619ab send along some minor fixes to the transition, @lucianmurmurache 😊 I found that adding in a `pb-1` on rotation fixed the issue that the icon looked like it was moving a bit, and I slowed the duration down to 200ms. Also added in `motion-reduce:transition-none` to the one in the left sidebar header 🥳 Hope the changes make sense to you :) Thanks for the issue and the implementation!

PR #730, opened 2024-02-27
chore: add animation to sidebar icon on toggle

lucianmurmurache:
### Contributor checklist

---

### Description
Currently we are replacing the icon based on the open/close state for both left and right sidebars. Instead, this approach will avoid the icon re-render while at the same time, add a little animation. The left sidebar icon will be rotated -180deg while the right icon will be rotated 180deg to keep the animations consistent.

### Related issue

- #729


andrewtavis (approved):
Thanks for this, @lucianmurmurache! I'm going to play around with the speed a bit and maybe slow it down slightly, and there seems to be a bit of a difference between the vertical spacing on either end of the chevron in the icon that makes it seem to shift when it's clicked. Should be easy to fix! Appreciate the quick work here! 😊

PR #731, opened 2024-02-27
fix: close sidebar on route change

lucianmurmurache:
### Contributor checklist

---

### Description
This addresses the discussion/decision to close the sidebar when the language option is selected (route change). This approach I decided on, would entail the least amount of changes, simply watching for route changes and then triggering the close function. 

### Related issue

- #711


andrewtavis (approved):
I really like this, @lucianmurmurache 😊 I wasn't actually thinking of it in terms of i18n route changes, but I think this also makes sense as the user would want to check it out a bit and see how the texts look. Originally was thinking more routing from the information or create menus, but way better this way. Thank you!

Issue #726, opened 2024-02-27
Update the copyright year on the website footer dynamically


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now in the [footer components](https://github.com/activist-org/activist/tree/main/frontend/components/footer) we have 2023 as the copyright year, but it would be great if this was dynamically change using TS. The goal of this issue would be to change the current way the copyright is displayed such that the localizable text is still there, but broken into three parts where the first is `Copyright ©`, the second is the dynamic year, and the third is `activist`.

### Contribution

Happy to support someone who has interest in working on this or getting to it myself at some point!

lucianmurmurache:
@andrewtavis I can take this one ✋

andrewtavis:
Thanks @lucianmurmurache! Will review later tonight odds are :)

andrewtavis:
Closed via #732 🙌 Thank you for this, @lucianmurmurache! Writing it here as well, please let us know if there are some improvements to the i18n strings that you can see that would benefit from a variable being passed to them (considering though that word orders might change in some locales).

On to the next pr 😊

PR #732, opened 2024-02-27
fix: make footer copyright year dynamic

lucianmurmurache:
### Contributor checklist

---

### Description
Change the footer year based on current year.

### Related issue

- #726

### DIFF
--- a/frontend/components/footer/flex/FooterFlex.vue
+++ b/frontend/components/footer/flex/FooterFlex.vue
@@ -51,7 +51,9 @@
         {{ $t("components.footer._global.powered-by-netlify") }}
       </a>
       <div class="mt-2 text-light-text dark:text-dark-text">
-        {{ $t("components._global.copyright") }}
+        {{
+          $t("components._global.copyright", { year: new Date().getFullYear() })
+        }}
       </div>
     </div>
     <!-- Note: Content Sections Right -->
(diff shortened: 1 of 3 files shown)


frontend/components/footer/flex/FooterFlex.vue
```diff
@@ -51,7 +51,9 @@
         {{ $t("components.footer._global.powered-by-netlify") }}
       </a>
       <div class="mt-2 text-light-text dark:text-dark-text">
-        {{ $t("components._global.copyright") }}
+        {{
+          $t("components._global.copyright", { year: new Date().getFullYear() })
```
andrewtavis, line 55: I really like this, @lucianmurmurache! 👏 Super cool! Let us know if you come across other places where we can pass some variables to the i18n strings so clean them up, but also keep in mind that word order might need to change in other languages :)

andrewtavis (approved):
Thanks for the change and the really intuitive usage of the i18n string, @lucianmurmurache!

frontend/components/footer/flex/FooterFlex.vue
lucianmurmurache: @andrewtavis sure I can take a look and create a PR if I find anything. In terms of the word order, if you meant for the change I made here, that should not be an issue, when writing the translation, all we need to do is to ensure that `{ year }` gets placed in the right spot and it should work like a charm!

andrewtavis: All sounds great, @lucianmurmurache!

Issue #725, opened 2024-02-27
Home screen is defaulting to the system color theme and isn't following user set preferences


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Something that was noticed during the most recent dev sync is that the [home page](https://activist.org/en/home/) is defaulting to dark mode. How to reproduce this is:

- Go to the page and see that it's in dark mode
- Go to the landing page and switch the color mode to light mode
- Go to any other page and see that the color mode selection is maintained
- Go back to the home page and it will switch to dark mode

Would be great if someone wanted to support with this! Happy to support or get to this eventually 😊

lucianmurmurache:
@andrewtavis I can take a look at this issue ✋

andrewtavis:
Thanks for this, @lucianmurmurache! 🎉🎉

andrewtavis:
Closed via #733 🥳🥳🥳🥳 So happy to have this fixed as it was really annoying for a bit there 😊

PR #733, opened 2024-02-27
fix: replace assignment with comparison operation

lucianmurmurache:
### Contributor checklist

---

### Description
This addresses the theme change on re-routing to certain pages. The issue was related to the friendly-captcha component where instead of checking whether the theme is dark, it would assign it. 

### Related issue

- #725


andrewtavis:
Wonderful stuff, @lucianmurmurache! Will do the final review later :)

andrewtavis (approved):
Thank you, @lucianmurmurache! Great find and so awesome to have this annoying little bug fixed :) :)

PR #742, opened 2024-03-04
Feat/tooltip modal for application status

UnknownSean8:
### Contributor checklist

---

### Description
Hi @andrewtavis, I was wondering if this is how it's supposed to look like.

Some questions:
1. Is the tooltip message for approved and pending the same?
2. Is the modal content for approved and pending the same? Except that pending have an input for writing discussions.

Thanks!

### Related issue

- #669


andrewtavis:
Is generally looking ok, @UnknownSean8 :) Some comments on it:

- I'm assuming that there's a boolean for if it's accepted or not and that the `About` page doesn't change for accepted orgs, right?
- If the organization is still pending, the message on hover should be `This organization is still waiting for approval from other organizations to join activist.`
- If the organization has been accepted, the message on hover should have a bit more information about what clicking would do: `tech from below joined activist on {Date} based on the support of 60 organizations already on activist. Click here to see their application.`
- We only want to see the application when the icon is the accepted one, so no click action when the organization is pending
  - When there is a click action to see the modal, let's also add `cursor-pointer` to the class so we get a hand cursor
- We want to have the votes component also be in the application modal above the discussion
- Do you want to also do the work for the `Application` tab here?
  - We'd want it to be above `About` with the rest of the sidebar elements being disabled
  - Maybe this should be a separate issue though?

Thanks for all the work here! Let me know if you need more information or a quick call to discuss 😊

UnknownSean8:
> Is generally looking great, @UnknownSean8 :) Some comments on it:
> 
> * I'm assuming that there's a boolean for if it's accepted or not and that the `About` page doesn't change for accepted orgs, right?
> * If the organization is still pending, the message on hover should be `This organization is still waiting for approval from other organizations to join activist.`
> * If the organization has been accepted, the message on hover should have a bit more information about what clicking would do: `tech from below joined activist on {Date} based on the support of 60 organizations already on activist. Click here to see their application.`
> * We only want to see the application when the icon is the accepted one, so no click action when the organization is pending
>   
>   * When there is a click action to see the modal, let's also add `cursor-pointer` to the class so we get a hand cursor
> * We want to have the votes component also be in the application modal above the discussion
> * Do you want to also do the work for the `Application` tab here?
>   
>   * We'd want it to be above `About` with the rest of the sidebar elements being disabled
>   * Maybe this should be a separate issue though?
> 
> Thanks for all the work here! Let me know if you need more information or a quick call to discuss 😊

I don't mind working on the application as well, but from what I remember, we've removed the need for the application tab and changed it to the tooltip + modal, or are we doing both together?

But I might only be able to push probably by the end of this week, is that okay?

Thanks! 😊

andrewtavis:
100% ok that it comes later this week. We're keeping the `Application` tab only for pending organizations. So it will be the first tab for organizations who have applied but haven't been accepted, but will go away when the org is accepted.

andrewtavis:
Hey @UnknownSean8 👋 Looking at the [organization application on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2642%3A20983&mode=design&t=HLnUrmfuyXSJsHL7-1), I think we'd be fine with just not including the tab for applications. There's nowhere to navigate, and it makes it easier on us. Also is a bit more UX friendly in so far as we won't have people wondering where the `Application` tab has gone after they're accepted :)

Mobile would need a change though. Specifically, if the organization is pending, then we wouldn't want to display [MenuNavigationDropdown](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuNavigationDropdown.vue). Let me know how this sounds! From here we'd just need some minor changes for it all to be ready 😊

UnknownSean8:
Hi @andrewtavis, these are the updates. 😄 

I'm assuming that there's a boolean for if it's accepted or not and that the About page doesn't change for accepted orgs, right?
- The pending/accepted is using a value in the organization object.
- The items in the About page are shown based on the items in Figma.

Mobile would need a change though. Specifically, if the organization is pending, then we wouldn't want to display [MenuNavigationDropdown](https://github.com/activist-org/activist/blob/main/frontend/components/menu/MenuNavigationDropdown.vue).
- Since the current organization status is based on a value in an object in about.vue, it would be a problem to get the value to the MenuNavigationDropdown, at least for now since the API call and client state are not initialized yet. I might've missed something for this.

Done
- [x] If the organization is still pending, the message on hover should be This organization is still waiting for approval from other organizations to join activist.
- [x] If the organization has been accepted, the message on hover should have a bit more information about what clicking would do: tech from below joined activist on {Date} based on the support of 60 organizations already on activist. Click here to see their application.
- [x] We only want to see the application when the icon is the accepted one, so no click action when the organization is pending
- [x] When there is a click action to see the modal, let's also add cursor-pointer to the class so we get a hand cursor
- [x] We want to have the votes component also be in the application modal above the discussion

Thanks!

frontend/components/card/discussion/CardDiscussionText.vue
```diff
@@ -35,6 +36,6 @@ import type { DiscussionText } from "~/types/card-discussion-text";
 
 defineProps<{
   isPrivate?: boolean;
-  discussionText: DiscussionText;
+  discussionTexts: DiscussionText;
```
andrewtavis, line 39: I'm gonna revert this back as we'll be querying from the backend for this, @UnknownSean8 :) Also, the type here would need to be `DiscussionText[]`.

frontend/components/header/HeaderAppPage.vue
```diff
@@ -48,13 +52,15 @@
 </template>
 
 <script setup lang="ts">
+import type { DiscussionText } from "~/types/card-discussion-text";
```
andrewtavis, line 55: We're not using this, so it's not needed to import it, right, @UnknownSean8? Maybe something from an earlier way you were working on this?

frontend/pages/organizations/[id]/about.vue
```diff
@@ -47,30 +56,42 @@
...
         :donationPrompt="organization.donationPrompt"
-      /> -->
+      />
+
+      <h3 class="text-left responsive-h3 font-display">
+        {{ $t("_global.discussion") }}
+      </h3>
+      <hr />
+
+      <CardDiscussionText
```
Comment (andrewtavis) [inline frontend/pages/organizations/[id]/about.vue:78]: Rather than doing the for loop within the component, we'd instead want the loop to happen on the given page. We'll make a `Discussion` component at some point that has all of the components here that then does the for loop and has an input that can be disabled based on if the user is alloed to participate or not. Maybe this is something you'd want to work on next, @UnknownSean8? I can make an issue for this!

frontend/components/modal/ModalOrganizationStatus.vue
```diff
@@ -0,0 +1,76 @@
...
+        />
+      </div>
+    </template>
+    <template #modalDisplay>
+      <div class="px-2 pb-2 pt-1 lg:px-4 lg:pb-4 lg:pt-2 max-h-[80vh]">
+        <DialogTitle class="font-display">
+          <p class="responsive-h2 font-bold">
+            {{ organization.name }}'s Application
+          </p>
+        </DialogTitle>
```
andrewtavis, line 38: Below this we also need `Status: Accepted` :) Adding it now 🚀

andrewtavis (approved):
Went through and did some changes to improve things a bit, @UnknownSean8 😊 Firstly I went ahead and did the edits to add in the `Discussion` component as I thought it made sense to add it now. Along with this there are some minor styling changes coming through and adding in things like the status codes to the i18n files. Let me know if you have any questions on the changes! Thanks for all the hard work here! 🥳

Issue #737, opened 2024-03-01
Switching to uv python package manager


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

A new python package manager emerged called [uv](https://astral.sh/blog/uv) from the team behind [ruff](https://docs.astral.sh/ruff/). Main benefit of switching to uv are the performance improvements (this will also reduce ci backend workflow time).

Switching would involve:

- adjustments to the readme / contrib guide 
- ci backend workflow
- updated requirements files 

There are no compatibility issues for us. It is a drop in replacement for virtualenv, pip, pip-tools (pip-compile and pip-sync). 

### Contribution

I am happy to help :)

andrewtavis:
Let's discuss this on Tuesday, @to-sta! Thanks for looking into it :)

to-sta:
As discussed today with @andrewtavis we are going use uv in the ci backend workflow for now.

andrewtavis:
Closed via #745 as I figure we can open another issue for future developments for this, @to-sta 🥳 Thank you!

PR #745, opened 2024-03-05
uv python package manager for backend workflow

to-sta:
### Contributor checklist

---

### Description

I replaced pip with uv as the python package manager to install all the dep. in the backend workflow. We can validate the promised performance benefit by comparing the workflow times.

### Related issue

- #737

### DIFF
--- a/.github/workflows/pr_ci_backend.yaml
+++ b/.github/workflows/pr_ci_backend.yaml
@@ -20,8 +20,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


to-sta:
Comparision of the workflow time:

uv: 14 seconds
pip: 24 seconds (I checked the workflow time from the PR #742)

andrewtavis:
🔥🔥🔥🔥🔥

Nice @to-sta 😊 Thanks for looking into this :

andrewtavis (approved):
Thanks for looking into this, @to-sta! 😊

Issue #748, opened 2024-03-07
Replace motion-reduce tailwind classes with reduced-motion package


lucianmurmurache:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

A more simplified approach to handling `prefers-reduced-motion` user preference. Instead of always having to manually add `motion-reduce:transition-none` or `motion-safe:animate-wiggle`, we can use this [npm](https://www.npmjs.com/package/reduced-motion?activeTab=readme) (shameless plug) package that handles this if the user preference is detected.

andrewtavis:
47bac0f is the edit of the changelog for this, @lucianmurmurache 😊 Thanks for the consideration and implementation here!

PR #749, opened 2024-03-07
chore/add-reduce-motion-pkg

lucianmurmurache:
### Contributor checklist

---

### Description
Swap `motion-reduce:some-class` for `reduced-motion` package to override animations when users have animations disabled on their device.

### Related issue

- #748

### DIFF
--- a/frontend/assets/css/tailwind.css
+++ b/frontend/assets/css/tailwind.css
@@ -177,9 +177,3 @@
     @apply opacity-50;
   }
 }
-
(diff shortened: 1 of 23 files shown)


andrewtavis (approved):
Really great idea for a package, @lucianmurmurache! Happy to adopt it 😊 I'll send along a quick change to the contribution guide to remove the part where we ask them to account for reduced motion preferences :)

Issue #576, opened 2023-11-23
Add a11y checks to repository


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would look into adding an accessibility workflow to the frontend checks that we already have in [pr_ci_frontend.yaml](https://github.com/activist-org/activist/blob/main/.github/workflows/pr_ci_frontend.yaml). The general thought would be to go with [axe](https://github.com/dequelabs/axe-core) for this and also add [axe Accessibility Linter](https://marketplace.visualstudio.com/items?itemName=deque-systems.vscode-axe-linter) to the suggested VS Code extensions that we have in the readme and contributing guide. Specifically in looking into it axe has support for Vue, but other options such as [Lighthouse](https://github.com/GoogleChrome/lighthouse) could also be considered :) Lighthouse also offers SEO checks as well that might be something to add.

### Contribution

Happy to work with someone on this!

andrewtavis:
CC @to-sta and @therealhoppi (that [article you sent along](https://www.matuzo.at/blog/building-the-most-inaccessible-site-possible-with-a-perfect-lighthouse-score/) got me thinking that we should check this).

thiagogquinto:
Hi @andrewtavis, I would like to work on it! Could you assign me?

andrewtavis:
Hey @thiagogquinto 👋 You can definitely implement this! :) Do you want to help a bit with the research for it beforehand? It looks like there are no [Lighthouse](https://github.com/GoogleChrome/lighthouse) related VS Code extensions for a11y as all I get for a search for "Lighthouse" is a theme 🧑‍🎨😅

Specifailly we're looking at what the Vue support is as it seems like [axe](https://github.com/dequelabs/axe-core) has it. Could you check if there's a good way to implement Lighthouse checks for Vue with GitHub Actions, and maybe we can make another issue? That new issue can be for adding in anything that Lighthouse has that axe might not like SEO suggestion 😊

thiagogquinto:
Thanks, sounds good! I will make some research about it.

andrewtavis:
Thanks @thiagogquinto! Let us know if there's anything we can do to support :)

thiagogquinto:
[...]

Issue #338, opened 2023-09-03
Allow collapsed sidebar to tab to search bar and open when focussed


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now in the left sidebar that's present on [activist.org/home](https://activist.org/home) and other test pages it's not possible to tab to the search bar when the sidebar is collapsed. The ideal behavior would be that when tabbing the sidebar would open when the search bar is focussed and then would collapse again if the user tabs past it.

### Contribution

Happy to work on this myself or support someone who has interest!

andrewtavis:
@luuu-xu, tagging you here for your opinion on this as I noticed this when I was reviewing #333 :)

yetea:
I'd like to work on this issue

andrewtavis:
Thanks for the interest in helping here, @yetea! Please let me know if there's anything I can do to assist 😊

andrewtavis:
Hey @yetea 😊 Checking in to see if you need any assistance with this issue :) Please let me know if there's anything I can do to help out! 

Hope you're well :)

andrewtavis:
Unassigning for now @yetea, but please let me know if you still have interest in working on this :)

uyk23:
Hi! I'd like to take a try at this issue if possible :)

andrewtavis:
Thanks for offering to help, @uyk23! Please let me know if there's something I can do to support :) :)

uyk23:
Thank you!

andrewtavis:
Hey @uyk23 👋 Checking in to see if you need some assistance here :) Please let me know if there's anything I can do to help!

andrewtavis:
Ping @ahmedy00: this issue seems to be a bit tougher than I expected. Bringing you in for some support for @uyk23, and we can problem solve on it as well in case it's needed 😊

ahmedy00:
Hey @uyk23, I would be happy to pair with you on this

uyk23:
Hey @andrewtavis @ahmedy00! I think I would benefit from the help, thanks! :)
So far, I was able to figure out how to tab over to the search symbol and highlight it orange.

[...]

PR #755, opened 2024-03-10
chore: add eslint plugin vuejs accessibility

lucianmurmurache:
### Contributor checklist

---

### Description
Implement [eslint-plugin-vuejs-accessibility](https://www.npmjs.com/package/eslint-plugin-vuejs-accessibility).
Address issues flagged by the plugin.
Update docs to reflect the above updates.

### Related issue

- #576

### DIFF
--- a/frontend/components/SearchBar.vue
+++ b/frontend/components/SearchBar.vue
@@ -6,16 +6,23 @@
     <div class="flex items-center pl-1 space-x-2">
       <Icon class="flex-shrink-0 w-4 h-4 my-1" name="bi:search" size="1em" />
       <Transition name="search">
-        <input
+        <div
           v-if="sidebar.collapsed == false || sidebar.collapsedSwitch == false"
-          @focus="onFocus"
-          @blur="onFocusLost"
-          ref="input"
-          class="w-16 h-5 bg-transparent outline-none"
-          :class="{ 'focus:w-5/6': isInputFocused }"
-          type="text"
-          :placeholder="$t('_global.search')"
-        />
+        >
+          <label for="input-search" class="sr-only">{{
+            $t("_global.search")
+          }}</label>
+          <input
+            @focus="onFocus"
+            @blur="onFocusLost"
+            ref="input"
+            id="input-search"
+            class="w-16 h-5 bg-transparent outline-none"
+            :class="{ 'focus:w-5/6': isInputFocused }"
+            type="text"
+            :placeholder="$t('_global.search')"
+          />
+        </div>
       </Transition>
     </div>
     <Transition name="shortcuts">
@@ -66,11 +73,13 @@
       :name="expanded ? 'bi:x-lg' : 'bi:search'"
       size="1em"
     />
+    <label for="expanded-search-input">{{ $t("_global.search") }}</label>
     <input
       v-if="expanded"
+      id="expanded-search-input"
       class="bg-transparent focus:outline-none"
       type="text"
-      placeholder="Search"
+      :placeholder="$t('_global.search')"
     />
     <Icon v-if="expanded" class="absolute right-3" name="bi:filter" />
   </div>
--- a/frontend/components/card/discussion/CardDiscussionInput.vue
+++ b/frontend/components/card/discussion/CardDiscussionInput.vue
@@ -109,7 +109,9 @@
         <div class="flex items-center space-x-3">
           <Button
             @mouseenter="showTooltip = true"
+            @focus="showTooltip = true"
             @mouseleave="showTooltip = false"
+            @blur="showTooltip = false"
             @click="showTooltip = showTooltip == true ? false : true"
             class="relative flex items-center justify-center w-16 h-10 rounded-lg elem-shadow-sm focus-brand"
             :class="{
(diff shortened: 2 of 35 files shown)


andrewtavis:
Nice @lucianmurmurache! Is there a way that we can also add this into the `pr_frontend_check` workflow?

lucianmurmurache:
> Nice @lucianmurmurache! Is there a way that we can also add this into the `pr_ci_frontend` workflow?

@andrewtavis it's already part of the flow. For example this [run](https://github.com/activist-org/activist/actions/runs/8222891255) failed due to a11y checks (happy little accident 🎨🖌️) `eslint . --ext .js,.vue` runs all the listed checks in [.eslintrc.js](https://github.com/activist-org/activist/blob/bc488886456c7716eaf2a35717e2530d519fa6c1/frontend/.eslintrc.js), and I added the plugin to the [eslint module](https://github.com/activist-org/activist/pull/755/files#diff-093b2d5625f20b6d845fe73f1af12762fc64ba9d0b9d94b14e8aa287e2d2ed1dR4). Also your editor will scream errors at you for :accessibility: violations instead of warnings.

andrewtavis:
This is amazing 🤩 Thanks @lucianmurmurache! :) :)

lucianmurmurache:
@andrewtavis I can make a little change here and also solve #338 in the same PR if you agree to a 2in1 approach

andrewtavis:
Either or, @lucianmurmurache! Happy to get another PR, but feel free to send it along to this one as well :) Goal is to merge by tomorrow :)

frontend/components/SearchBar.vue
```diff
@@ -66,11 +73,13 @@
...
       size="1em"
     />
+    <label for="expanded-search-input">{{ $t("_global.search") }}</label>
     <input
       v-if="expanded"
+      id="expanded-search-input"
       class="bg-transparent focus:outline-none"
       type="text"
-      placeholder="Search"
+      :placeholder="$t('_global.search')"
```
andrewtavis, line 82: Nice catch, thanks! 😎

frontend/components/card/discussion/CardDiscussionInput.vue
```diff
@@ -109,7 +109,9 @@
         <div class="flex items-center space-x-3">
           <Button
             @mouseenter="showTooltip = true"
+            @focus="showTooltip = true"
```
andrewtavis, line 112: Nice, yes definitely needed :)

frontend/components/sidebar/left/SidebarLeft.vue
```diff
@@ -4,10 +4,20 @@
...
+    "
     @mouseleave="
       sidebar.collapsed = true;
       setContentScrollable();
     "
+    @blur="
+      sidebar.collapsed = true;
+      setContentScrollable();
+    "
+    role="menu"
```
andrewtavis, line 19: Minor note here: as we are switching from the activist logo to an icon when we switch to a collapsed sidebar we are losing focus on blur when we tab from the sidebar itself to the logo :) Do you want to send along a quick fix for this?

Note that I have a commit coming with some changes shortly 😊

frontend/components/menu/MenuSearchResult.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <div
     @mouseleave="showTooltip = false"
+    @blur="showTooltip = false"
```
andrewtavis, line 4: I'm gonna revert these changes as it's actually not the desired behavior :) This div is to add the `p-2` around the button or the tooltip if it's shown and allow for the mouse leave only, but shouldn't be focused itself. The focus should go directly to the button. If we need these in here for a11y reasons, then let's discuss it in a sync as this component is actually pretty fine tuned for usability 😊

frontend/components/sidebar/left/SidebarLeft.vue
andrewtavis: As this is sidebar related, maybe we can add this into #338 so we can merge this in, @lucianmurmurache :) :)

andrewtavis:
Hehe ok then, let me revert some of my changes 😅 Why aren't these showing up in ESLint though, @lucianmurmurache?

andrewtavis:
As of now I'm adding the following to `MenuSearchResult` as I'm really not thinking that there's an easy solution for this:

```html
  <!-- eslint-disable-next-line vuejs-accessibility/mouse-events-have-key-events, vuejs-accessibility/no-static-element-interactions -->
```

I'm actually not getting the `Anchors must have content and the content must be accessible by a screen reader` errors locally though 🤔 Seems like this could be fixed in a few ways. Let me know if you have suggestions on this, @lucianmurmurache!

andrewtavis:
054024a sends along the final fixes. We needed some aria labels for `<a>` tags. I did end up ignoring the ESLint errors for `MenuSearchResult` for now. If we want to work on that, then we should likely talk about it in a dev sync before so I can go over the functionality 😊

andrewtavis (approved):
This is a huge step for the frontend checks, @lucianmurmurache! Thanks so much for the work here. Really is heartening to have so much work going into a11y now 😊

Issue #644, opened 2024-01-05
Persistent highlight from `focus-brand` custom class


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Something that we've been noticing recently is that the [focus-brand](https://github.com/activist-org/activist/blob/fe68c9b895da943cde0034cda2ec311fcb8834ac/frontend/assets/css/tailwind.css#L46) class that determines how page elements look on focus is persistent - meaning that an item's focus indicator isn't disappearing (usually after the next interaction it goes away).

A simple way to reproduce this is:
- Go to the [help section of the documentation](https://activist.org/en/help)
- Click any option in `Resources` or `Organization` including `Help` again to reload the page
- See that the clicked option is still focused when the page is reloaded

Thoughts on this and a fix would be much appreciated!

Jayraj-R:
If this issue is still open, I would like to give it a try

andrewtavis:
Hey @Jayraj-R 👋 Checking in with you on this :) Is there anything I can do to support?

devaggarwal-1:
If this issue is still not solved, I would like to work on it.

andrewtavis:
Hey @devaggarwal-1 👋 I'd be fine with switching it over to you :) Thanks for checking in!

@Jayraj-R, please let us know if there's another issue you'd like to work on some time!

lucianmurmurache:
@andrewtavis I can take a look at this ✋

andrewtavis:
Thanks @lucianmurmurache!

andrewtavis:
Closed via #756 🥳 Thanks @lucianmurmurache! :) :)

PR #756, opened 2024-03-10
style: replace focus with focus-visible

lucianmurmurache:
### Contributor checklist

---

### Description
This is both a slight UX imprvement and a fix for the issue mentioned below. Focus styles are generally not required when the user knows where they are focusing on, meaning if they use a touch screen or a mouse, it's not necessary to show the focus ring on the element you just pressed/clicked (a few exceptions, such as inputs), it is however absolutely necessary when using a keyboard. `focus-visible` is a viable replacement. Now we have a slightly cleaner design for users that use the mouse, and we still show a focus for things like inputs and textareas. A more detailed explanation can be found on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible)

### Related issue

- #644

### DIFF
--- a/frontend/assets/css/tailwind.css
+++ b/frontend/assets/css/tailwind.css
@@ -48,7 +48,7 @@
   }
 
   .focus-brand {
-    @apply rounded-sm focus:ring-2 focus:ring-light-link-text dark:focus:ring-dark-link-text focus:outline-none focus:border-light-link-text dark:focus:border-dark-link-text;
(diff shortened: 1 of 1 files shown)


andrewtavis:
Thanks for the research and the fixes for this, @lucianmurmurache! Will bring this in soon 🎉

andrewtavis (approved):
Is working perfectly, @lucianmurmurache! Thanks so much for the support here 😊

Issue #744, opened 2024-03-04
Update models and visualizers according to figma schemas


thiagogquinto:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The [figma schemas](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377-33014&mode=design&t=PzjgvzFnDcuQIjbv-0) of some models have been updated with some new fields, this issue focuses on implementing the changes to the backend so the models and serializers be on the date.

### Contribution

Happy to work on this!

andrewtavis:
Thanks @thiagogquinto! :)

to-sta:
Hi @thiagogquinto, this might be helpful [graph_models](https://django-extensions.readthedocs.io/en/latest/graph_models.html#example-usage). This is part of the django extensions package and let's you create ER-Modell automatically based on the django-project. Haven't used it yet myself but thought this might be useful in the future.

thiagogquinto:
@andrewtavis, the current _support_ table [here](https://github.com/activist-org/activist/tree/main/backend/authentication), has the following code:

```python
class Support(models.Model):
    supporter_type = models.ForeignKey(
        "SupportEntityType", on_delete=models.CASCADE, related_name="supporter"
    )
    supporter_entity = models.ForeignKey(
        "entities.Organization", on_delete=models.CASCADE, related_name="supporter"
    )
    supported_type = models.ForeignKey(
        "SupportEntityType", on_delete=models.CASCADE, related_name="supported"
    )
    supported_entity = models.ForeignKey(
        "entities.Organization", on_delete=models.CASCADE, related_name="supported"
    )

    def __str__(self) -> str:
        return f"{self.id}"
```

But the figma schema show that the fields ```supporter_entity``` and ```supported_entity``` points to User's ID, by the name looks like that the current code is correct, but I want to be sure.

thiagogquinto:
@andrewtavis , some questions have come up, would you be able to have a meeting sometime?

andrewtavis:
Yes sure, but it'd likely be best if @to-sta is there as well. He and I are meeting tonight at 18:00 UTC. I would that work for you too?

thiagogquinto:
[...]

Issue #626, opened 2023-12-25
Add flag table and to backend via schemas


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add in the flag table to the backend within the [backend/content](https://github.com/activist-org/activist/tree/main/backend/content) directory. The schemas for the table can be found in the [designs on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=8123%3A8250&mode=design&t=hNXVijyV1rCwwTiC-1). Another thing that would need to happen is that the `total_flags` columns should be removed from the tables in which they're found as we need to have unique identifiers for who is flagging what to avoid repeats.

andrewtavis:
@to-sta, general idea here was that we follow the support table so there's a shared logic. Only difference is that users are the only entities who can flag, so we don't need to do any entity matching for "flagger".

andrewtavis:
Assigning this to myself, @to-sta. Let me make a PR for this and you can review :)

to-sta:
Yeah, sure thing. Happy to see you joining the backend 🥳

sh-ran:
Hey! @andrewtavis 

So I took a look at the Figma and just wanted to clarify what I understood from past conversations about this issue.

We are creating a flag table in order to maintain a int value denoting when a user flags a resource.

So I would have to write a model for `Flags` and `FlagsEntity` in `models.py`. And then add their own `Viewsets` and `serializers` for the model. If that is the case, I am guessing this will have its own API endpoint with all the CRUD operations as other viewsets?

andrewtavis:
I think we can similarly do an API endpoint here. Not sure particularly on if it's super necessary for pagination for flags, but wouldn't hurt 😊 Keeping it consistent with the other styles would also be good :)

sh-ran:
Hey! @andrewtavis 

Just wanted to clarify, are all the three fields foreign key instead of primary key?

![Image](https://github.com/user-attachments/assets/9f49568c-1c1f-4cc0-9d1e-fa5d6b47b74f)

andrewtavis:
Would be great if you could bring in main for the next PR, @sh-ran, but we can help take care of that once the PR is open too :)

PR #757, opened 2024-03-11
Update models and serializers

thiagogquinto:
### Contributor checklist

---

### Description

- The PR update the backend of the already existing models to be on date with the latest version of the schema on figma.
- Some fields that no longer exist have been mainteined because for the removal would be necessary add new tables.
- I changed a little of the factories
- The suport table was changed so the fields can be "linked" to different models.

### Related issue

- #744


thiagogquinto:
I added a check for the fields end_time and start_time of the Event Model, and there is an type error because of it. But I am trying to fix it.

thiagogquinto:
I have made new commits after the tenth of them, but these commits have not been brought to the Pull Request, is this my fault?

thiagogquinto:
Also during looking at the factories.py files I noticed that specially the fields *creation_date* and *deletion_date* are being configured by different ways depending of the file, It's just a suggestion, but i think fot he best we shoud standandize this.

andrewtavis:
Hey @thiagogquinto! @to-sta and I are taking a look at this now :) Some minor things:

- It'd be good if you'd install the [Ruff extension for VS Code](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) that will take care of the formatting and import ordering errors :)
- Could you go through the [Pull Requests section of the contributing guide](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#pull-requests-) and look into installing [pre-commit](https://pre-commit.com/) as it's described there

Want to stress that there's no issue on our end with the checks failing! All your work is really appreciated, and we're just thinking that these two steps would take your backend commits to a level where we'll be able to merge on green from the get go ✅🚀

backend/authentication/models.py
```diff
@@ -49,21 +61,23 @@ def __str__(self) -> str:
...
         "User", on_delete=models.SET_NULL, null=True
     )
     user_icon = models.ForeignKey("content.Image", on_delete=models.SET_NULL, null=True)
     social_accounts = ArrayField(
         models.CharField(max_length=255), null=True, blank=True
     )
-    total_flags = models.IntegerField(default=0)
     private = models.BooleanField(default=False)
     high_risk = models.BooleanField(default=False)
+    creation_date = models.DateTimeField(auto_now_add=True)
```
to-sta, line 78: There are few minor things here: 
* Creation and deletion date are added by the CreationDeletionMixin. 
* And we got rid of the total_flags.

andrewtavis: We're going to have a central `flags` table now, @thiagogquinto :) So with that we don't need `total_flags`, but rather will be referencing the `flags` table via uuids.

backend/content/models.py
```diff
@@ -25,27 +27,28 @@
...
 
 
 class Resource(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     name = models.CharField(max_length=255)
     description = models.TextField(max_length=500)
-    topics = ArrayField(models.CharField(max_length=255), default=list, blank=True)
-    category = models.CharField(max_length=255, blank=True)
+    topics = ArrayField(
+        models.CharField(max_length=255), default=list, blank=True, null=True
```
to-sta, line 43: CharField and TextField are never saved as NULL in django therefore null=True is not necessary.

In case you want to read a bit more about here is a good source https://stackoverflow.com/questions/8609192/what-is-the-difference-between-null-true-and-blank-true-in-django

backend/entities/models.py
```diff
@@ -29,46 +28,50 @@
 class Organization(CreationDeletionMixin):
```
to-sta, line 28: Since we do not use a `creation_date` field here anymore, we can get rid of the mixin:

```python
class Organization(models.Models):
```

backend/authentication/models.py
thiagogquinto: I removed the fields total_flags, but as there is an issue for the addition of the flag table (#626), I will not add on this PR, unless that's okay to add it.

to-sta:
@thiagogquinto I made a few minor changes and reverted the support table changes. I do like the content type approach, this might be useful for support, tag and the flag table 😃. But I think we should address the change in a seperate issue and dicuss this a bit.

to-sta (approved):

backend/content/models.py
```diff
@@ -25,23 +27,23 @@
...
-        return self.name
+        return self.question
 
 
 class Resource(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     name = models.CharField(max_length=255)
     description = models.TextField(max_length=500)
-    topics = ArrayField(models.CharField(max_length=255), default=list, blank=True)
+    topics = ArrayField(
```
andrewtavis, line 42: We'll likely need to move this to `resource_topics` at some point, but all of this is gonna be more clear once we start testing for it all :)

andrewtavis (approved):
Really great stuff, @thiagogquinto, and thanks for the work to clean things up, @to-sta!

Would be great to have you at a [dev sync](https://etherpad.wikimedia.org/p/activist-dev-sync) sometime @thiagogquinto! @to-sta were talking about it last night 😊 Really appreciate the continued contributions 😊 Next one is Saturday the 23rd, if it can work :)

Issue #684, opened 2024-02-02
Make cursor pointer on map marker hover


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In #680 we added a basic marker tooltip to the [MediaMap component](https://github.com/activist-org/activist/blob/main/frontend/components/media/MediaMap.vue). One thing that's missing is that the cursor doesn't turn into a pointer when we hover over the marker, so it's not clear that clicking will have any kind of effect.

Note that the events map can be found at [activist.org/events/123/about)](https://activist.org/events/123/about) :)

We use [MapLibre GL JS](https://github.com/maplibre/maplibre-gl-js) for our mapping service, so their documentation and code related to it or [MapBox](https://github.com/mapbox/mapbox-gl-js) that it was forked from would be where to look. The following is a great example, but is making a default element rather than using the default marker from MapLibre GL JS:

https://maplibre.org/maplibre-gl-js/docs/examples/set-popup/

If we want to make a custom marker such that we can change the cursor though, this would also be possible. The question is then where does the source image for this live :)

### Contribution

Happy to assist with this or work on it myself at some point!

arescrimson:
This sounds like an interesting issue here, @andrewtavis. Would be open to looking further into it; noticed that there are a lot of new issues regarding the map features, and this would be a great place to start learning how it works 😄 

In addition, could you let me know where this is located? I definitely remember seeing a map component somewhere(possibly events?), but I couldn't find it.

arescrimson:
Okay: on further investigation it just looks like the functionality for changing the cursor into a map interaction pointer(grab hand?) is already there - the problem is just actually getting it to change to an actual pointer once the marker is moused over. Interesting issue! Let me know if there's anything I can do to help 😁

arescrimson:
Another thing, and let me know if I'm looking in the wrong place; in the actual deployed website, the map looks like this.

![image](https://github.com/activist-org/activist/assets/66581240/a1c00fb4-53b9-4550-bc29-4a8eb4407099)

With the development build, the map and marker look different, and it looks like the issue is resolved because the cursor does become a pointer on hover. 

[...]

PR #759, opened 2024-03-13
feat: show pointer cursor on map marker hover

nkoehring:
### Contributor checklist

---

### Description

Adds the `cursor-pointer` class to the marker that is being added to the map.

### Related issue

- #684 

### Problems

Although the marker now shows a pointer, and clicking the marker shows the expected popup, it also starts the routing (nice feature, btw!). For me that seems odd and probably unintentional.

### DIFF
--- a/frontend/components/media/MediaMap.vue
+++ b/frontend/components/media/MediaMap.vue
@@ -121,10 +121,11 @@ onMounted(() => {
           </div>`
         );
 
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Thanks for the changes here, @nkoehring! Way better/simpler than what I was considering on all of this 😊 Appreciate the help!

PR #760, opened 2024-03-13
Events API #628

josueemartinezz:
### Contributor checklist

---

### Description

- Added (list, create, retrieve, update, partial_update, destroy) HTTP methods for Events. Ensured event creators are admin associated within their organization. 
- Modified serializers to validate non-trivial input. Trivial validation checks are removed as they're ensured by the model setup.
- Added missing fields for the Event model (according to Figma designs)

TESTING
- Extensive testing was done on the backend to ensure the model catches bad user input to eliminate validation checks done in serializers.
- Verified that upon organization creation, an organization admin member is allowed to create an event whereas if is_admin is false, HTTP responds appropriately.  

NOTE: 
- breaks might occur as during the testing stage I would receive independent errors coming from authentication and organization models. However, @andrewtavis confirmed to leave these model definitions as they are after manipulating them.

### Related issue

- #628


andrewtavis:
Thanks for the hard work here, @josueemartinezz! As an initial check on all of this, would you want to go through the [backend check](https://github.com/activist-org/activist/actions/runs/8270629212/job/22681835056?pr=760) and send along fixes for all of that? @to-sta and I can then take a look afterwards :)

andrewtavis:
Two minor things that I just shared with another contributor that would be helpful for you, @josueemartinezz:

- It'd be good if you'd install the [Ruff extension for VS Code](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) that will take care of the formatting and import ordering errors :)
- Could you go through the [Pull Requests section of the contributing guide](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#pull-requests-) and look into installing [pre-commit](https://pre-commit.com/) as it's described there

Want to stress that there's no issue on our end with the checks failing! All this is very appreciated, and doing the steps above will just take it all to the next level 🚀

backend/entities/views.py
```diff
@@ -43,6 +43,11 @@ class OrganizationViewSet(viewsets.ModelViewSet[Organization]):
     serializer_class = OrganizationSerializer
     pagination_class = CustomPagination
     throttle_classes = [AnonRateThrottle, UserRateThrottle]
+    
+    def list(self, request: Request, *args: str, **kwagrs: int) -> Response:
+        queryset = Organization.objects.all()
```
to-sta, line 48: @josueemartinezz 

The variable `queryset` is unused and redundant 😅.
```python
queryset = Event.objects.all()
serializer = self.get_serializer(self.queryset, many=True)
```

backend/entities/views.py
```diff
@@ -70,7 +75,7 @@ def destroy(self, request: Request, *args: str, **kwargs: int) -> Response:
         instance = get_object_or_404(Organization, pk=kwargs["pk"])
         instance.delete()
         data = {"message": f'Organization {kwargs["pk"]} has been deleted successfully'}
-        return Response(data, status=status.HTTP_204_NO_CONTENT)
+        return Response(status=status.HTTP_202_ACCEPTED)
```
to-sta, line 78: Minor thing here. Accepted is good status for this but we generally use the `status.HTTP_201_CREATED` in case we create an instance.

backend/events/models.py
```diff
@@ -26,11 +26,14 @@
 class Event(CreationDeletionMixin):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     name = models.CharField(max_length=255)
-    tagline = models.CharField(max_length=255, blank=True)
+    tagline = models.CharField(max_length=255, blank=True, null=True)
```
to-sta, line 29: In django CharField and TextField are never saved as NULL values, therefore there is no need for null=True.

backend/events/models.py
```diff
@@ -41,6 +44,8 @@ class Event(CreationDeletionMixin):
     event_icon = models.OneToOneField(
         "content.Image", on_delete=models.CASCADE, null=True, blank=True
     )
+    creation_date = models.DateTimeField(auto_now_add=True)
```
to-sta, line 47: `creation_date` and `deletion_date` are added by the mixin:

```python
class Event(CreationDeletionMixin):
     ...
```

backend/entities/views.py
```diff
@@ -70,7 +75,7 @@ def destroy(self, request: Request, *args: str, **kwargs: int) -> Response:
         instance = get_object_or_404(Organization, pk=kwargs["pk"])
         instance.delete()
         data = {"message": f'Organization {kwargs["pk"]} has been deleted successfully'}
```
to-sta, line 77: The variable data is created but it is not used. You have to add it to the response.

to-sta:
@josueemartinezz Thanks for contribution 😃. I added a few comments. How did you test the API?

josueemartinezz:
> @josueemartinezz Thanks for contribution 😃. I added a few comments. How did you test the API?
Hi! Thank you so much for your review. I'll get these fixed right away. I tested the API by using Swagger-UI for all the HTTP methods, ensuring bad user input passed the Event model's field validation requirements. If it did, like a user's end_time precedes start_time, I raised exceptions in serializers.py. I tested for foreign key existence, but this again was accounted for by the key checks done by the field type. Please let me know of any testing strategies tips apart from just playing around with inputs on Swagger-UI.

andrewtavis:
Heyo @josueemartinezz! We have the other priority backend PRs merged and will get to this when we can! Sorry for the wait! Map for events has been taking some major strides, so the feature set is definitely there for when this is brought in 😊

PR #762, opened 2024-03-17
fix: make left sidebar accessible by keyboard

lucianmurmurache:
### Contributor checklist

---

### Description
Make left sidebar accessible via keyboard 

### Related issue

- #338

### DIFF
--- a/frontend/components/dropdown/DropdownBase.vue
+++ b/frontend/components/dropdown/DropdownBase.vue
@@ -2,6 +2,7 @@
   <Menu as="div" class="relative inline-block text-left">
     <div>
       <MenuButton
+        @focus="expandOnFocus"
         v-slot="{ open }"
         class="inline-flex w-full px-4 py-2 font-semibold select-none rounded-md style-btn"
         :class="{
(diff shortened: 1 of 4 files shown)


lucianmurmurache:
@andrewtavis there are still a few small things to adjust here but overall this would be the fix for #338 + allowing users to navigate other bits of the sidebar, not just the searchbar. It also focuses on the logo. Just one thing I need to try and fix which is that sometimes the sidebar does not expand when focusing on the 1st element (activist logo). I am not sure if it makes sense to open a new issue, so I leave that decision for you to make 😁

andrewtavis:
Let's include it all in this PR, @lucianmurmurache! Thanks for the efforts here so far 😊 Can't wait for the sidebar to be even better :)

andrewtavis:
We ready to go here, @lucianmurmurache, or is there still a bit more to come? :)

lucianmurmurache:
@andrewtavis I was so sure I set this to draft. I am still trying to fix an edge case here, if you focus on the menu, it now expands and you can navigate without an issue, the problem is that if you focus in reverse order, it only expands on the 2nd element from down up, I did not yet work out the reason. I will only have more time on Thursday to look into it unfortunately

andrewtavis:
All good, @lucianmurmurache! Let me know when it's ready :)

andrewtavis (approved):
Working perfectly, @lucianmurmurache! Thanks for the attention to detail with even all directions of navigation 😊 Hope to see you at the sync/Code Night 🌙 on Saturday!

Issue #553, opened 2023-11-18
Add modal to MediaImageCarousel 


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We already have the component for displaying images on organization pages in [MediaImageCarousel.vue](https://github.com/activist-org/activist/blob/main/frontend/components/media/MediaImageCarousel.vue), but one thing that's needed is a functioning expand to modal button in the top left corner that opens the issues in a similar way that components in the [modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal) directory function. This would open the swiper over the screen with a slight blur as with other modals. Cleaning up the component so it matches [the designs](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2642%3A20983&mode=design&t=hbrmPFJFE6NT0uV9-1) a bit more could also be a great contribution.

### Contribution

This issue is a bit more difficult, but would be a `good first issue` for a more experienced developer :) Happy to support or get to it myself eventually! 😊

nicholas-vecchio:
Hello Andrew! I would like to be assigned this issue.

If I understand correctly, you want a modal to be added so that the media image carousel can be expanded and come to the fore-front while blurring the rest of the page content? Thank you.

andrewtavis:
Hey @nicholas-vecchio! Thanks for your interest in this one. You described it exactly as it should function, and feel free to reference the [modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal) directory where our other uses of [Headless UI modal](https://headlessui.com/vue/dialog) are :)

Let me know if there are further questions or if I can help somehow!

andrewtavis:
Hey @nicholas-vecchio, just btw, there's been some progress on this and we now have a BaseModal component from #582 that could help with this. Let me know if you have any questions!

nicholas-vecchio:
Thank you for the update @andrewtavis! Apologies for taking so long to progress on this issue.

andrewtavis:
No stress whatsoever, @nicholas-vecchio! Let me know if you need anything from my end :)

andrewtavis:
[...]

Issue #657, opened 2024-01-21
Remove strong dependency between modals and their sources


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

As of now [ModalBase](https://github.com/activist-org/activist/blob/main/frontend/components/modal/ModalBase.vue) is creating a strong dependency between the modal and the normal that's triggering it. We've discussed this most recently in the last developer sync as well as in conversations afterwards. It would be good if this were removed and there was a bit more independence with the modal being triggered by a prop that's then more customizable.

@rigojr brought this up and has expressed interest in working on it :)

andrewtavis:
Hey @rigojr 👋 Went ahead and made the issue here so we don't forget it 😊 Let me know if you still have interest in working on it!

rigojr:
Hi @andrewtavis I would love to make this refactor.

Let me know when I could start, I understand that someone is doing some work over there, so I could start after that, let me know what you think.

andrewtavis:
Thanks for the willingness to help, @rigojr! Let's wait for a PR for the share modal component to get merged and then we can start :)

andrewtavis:
Hey @rigojr 😊 Let me know if a call on this would be helpful at some point! We also now have a modal for displaying the organization application, and the share modal is also live that would ideally link to the QR code modal, but this is broken because of this...

Hope you're well! :)

andrewtavis:
Hey @rigojr 👋 Checking in with you on this, as this issue's now much more important given that we're using modals in a lot of other places on the platform. Would you have time to take a look at this, or we could also do a call where we discuss the work to be done so I can implement?

andrewtavis:
Changes to fix this are coming along shortly :) Noting some things about the changes here 😇

The following code is within `ModalBase` and all child components in the `modal` directory:

```ts
const props = defineProps<{
  isOpen: boolean;
}>();

const modalIsOpen = computed(() => props.isOpen);
const modalShouldClose = ref(false);

const emit = defineEmits(["closeModal"]);
const closeModal = () => {
  modalShouldClose.value = true;
  emit("closeModal");
  modalShouldClose.value = false;
};
```

Modal components that use `ModalBase` for their styling include the following:

```vue
<ModalBase
[...]

PR #764, opened 2024-03-19
Issue 553 add modal

nicholas-vecchio:
### Contributor checklist

---

### Description

Added a component, ModalMediaImageCarousel.vue, which is a component meant to house the MediaImageCarousel, so that it may be used as normal, and within a modal.

Two buttons are added in ModalMediaImageCarousel, one with a fullscreen symbol which opens a larger modal of the swiper, and a add button which is simply a placeholder, to match the designs.

MediaImageCarousel is also changed to take a prop called 'fullscreen' which simply changes the size of the images depending on if the component is being used as normal, or in a modal.

### Related issue

- #553


andrewtavis:
Thanks for this, @nicholas-vecchio! Will look to review soon :) 

Thanks further for the work to fix the PR check errors 🙌 Great to have another contribution from you 😊

andrewtavis:
Hey @nicholas-vecchio 👋 One thing that I'm wondering if you'd like to implement is arrow key navigation between the pictures in the modal? It's looking really really great – so much so that I tried to use arrow keys to navigate the modal images and then realized we hadn't planned on it 😅 Could also make a separate issue for that if you'd prefer!

nicholas-vecchio:
@andrewtavis I'll look into it and let you know how I go :)

andrewtavis:
Thanks, @nicholas-vecchio! There are doubtless many ways of doing this, but check how we're doing it in the [SearchBar](https://github.com/activist-org/activist/blob/9e25355d075ad89309f4521142beab55bb5f0a2e/frontend/components/SearchBar.vue#L105) component with [useMagicKeys](https://vueuse.org/core/useMagicKeys/) :)

andrewtavis:
Hey @nicholas-vecchio 👋 FYI the merge conflicts are coming in because we're using a new plugin to order the Tailwind classes rather than an extension. Happy to help fix them if you'd need the hand with it!

andrewtavis:
Went ahead and fixed the merge conflicts for you, @nicholas-vecchio 😊 Maybe some new frontend errors from some new checks will come through, but we'll be happy to help with those!

andrewtavis:
I'll fix the errors here. Big thing on this @nicholas-vecchio is that we have a specific style for adding Tailwind classes:

```
<div
  class="NON_CONDITIONAL_CLASS"
  :class="{CONDITIONAL_CLASSES_WITH_CONDITIONS_FULLY_DEFINED}"
>
  Stuff
</div>
```

nicholas-vecchio:
Thanks for all your help :) I'll make sure to give the style guide another read.

Having issues with the final check, I believe the problem is with how the swiper is taking paramaters and prettier not recognizing it.

`
SyntaxError: Unexpected token (4:5)
  2 |   <swiper-container
  3 |     class="swiper card-style h-full w-full overflow-clip"
> 4 |     :slidesPerView="1"
    |     ^
  5 |     :spaceBetween="0"
  6 |     :loop="true"
  7 |     :pagination="{ clickable: true }"
  `

nicholas-vecchio:
Turns out I was just bein' a little silly and not running eslint and prettier from the right directory. Managed to get it working. Thanks for all the help :)

andrewtavis:
Thank you, @nicholas-vecchio! Will try to bring this in in the coming days :)

andrewtavis:
One thing that I'm removing from here @nicholas-vecchio is the plus in the bottom right corner, but that's because it's being implemented in #768 :)

frontend/components/media/MediaImageCarousel.vue
```diff
@@ -5,13 +5,22 @@
...
     >
-      <img class="h-[17.5rem] object-cover object-center" :src="img" alt="" />
+      <img
+        class="object-cover object-center"
+        :class="{
+          'h-5/6 w-5/6': props.fullscreen,
+          'h-[17.5rem]': !props.fullscreen,
+        }"
+        :src="img"
+        :alt="$t('components.media-image-carousal.img-alt-text')"
```
andrewtavis, line 22: Minor addition of an i18n key for this 😊

frontend/components/media/MediaImageCarousel.vue
```diff
@@ -20,6 +29,10 @@
 import { register } from "swiper/element/bundle";
 register();
 
+const props = defineProps({
```
andrewtavis, line 32: In Nuxt we don't need to import `defineProps` :)

frontend/pages/organizations/[id]/about.vue
```diff
@@ -53,7 +53,7 @@
           :organization="organization"
         />
         <div class="h-full w-full">
-          <MediaImageCarousel :class="{ 'lg:hidden': textExpanded }" />
+          <ModalMediaImageCarousel :class="{ 'lg:hidden': textExpanded }" />
```
Comment (andrewtavis) [inline frontend/pages/organizations/[id]/about.vue:56]: We need to make sure to include this so that the image carousel disappears if there's enough text in the About section to have a `View full text` button and then it's clicked :)

andrewtavis (approved):
Really really great stuff, @nicholas-vecchio! One thing that's needed here is that we need the modal to be triggered on return key press when it's focussed, but let's wait on this until after #657 has been finished. Great to work with you again!

Issue #736, opened 2024-02-29
signup API endpoint


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We need endpoints for accounts related actions. This issues is about the signup / registration endpoint. 

- signup (POST, DELETE) 

The POST and DELETE methods should be used to create and delete an account.

The API endpoints should be created via `APIView` rather than a `ViewSet`.

### Contribution

I am happy to help someone to work on this or work on this myself.

jaymeklein:
Hey @to-sta, i would love to help in this issue and this project!
You can assign me to this issue if you want to, and if there's no one else with this in hands.
If so, could you give me some directions on the already built API? Where should i create this `signup (POST, DELETE)` endpoint? i couldn't find anything in the [CONTRIBUTING.md](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md) file about this specific topic.

I already have some experience with APIs (mostly with [Flask](https://flask.palletsprojects.com/en/3.0.x/)), but i can learn Django in no time.

Best regards.

andrewtavis:
Assigning you here now, @jaymeklein 😊 Thanks for the interest in helping out! @to-sta is busy today, but he'll get back to you soon on directions for the API endpoint! Looking forward to working with you further :) :)

jaymeklein:
Hey @andrewtavis , thanks for assigning me this issue.
I'm gonna wait until he is available to make everything here as clear as possible.

Best regards.

to-sta:
@jaymeklein thanks for taking on this issue 😃. 

All account related API's are inside the directory `backend/authentication`. We use the [django restframework](https://www.django-rest-framework.org/) to create API's. 

You can find the User model [here](https://github.com/activist-org/activist/blob/main/backend/authentication/models.py#L49). 

We want to use the APIView class for the implementation of the post and delete method. That should look like:
```python
class SignupView(APIView):
    def post(self, request):
         pass
     
    def delete(self, request, pk):
         pass
```

I would suggest to create a seperate serializer for the API as well (instead of using UserSerializer):

```python
class SignupSerializer(serializers.ModelSerializer[User])
     ...
```

[...]

PR #769, opened 2024-03-21
feat(authentication): add signup endpoint

jaymeklein:
### Contributor checklist

---

### Description
- Implement SignupSerializer in serializers.py
- Implement SignupView in views.py
- Add POST and DELETE methods for /signup endpoint
- Include 'signup' in urlpatterns in urls.py

### Testing

`POST` method to create the user:
```HTTP
POST /v1/auth/signup HTTP/1.1
Host: localhost:8000
Content-Type: application/x-www-form-urlencoded
Cookie: csrftoken=pUvfgN5WhUs09f53I7hYkBYJH1pGpQRK
Content-Length: 75

user_name=dummy&password=123456ABCdef%40&password_confirmed=123456ABCdef%40
```
The result must be just the `user_name` argument, in case of valid fields:
```json
{
    "user_name": "dummy"
}
```

`DELETE` method to remove the user:
```HTTP
DELETE /v1/auth/signup HTTP/1.1
Host: localhost:8000
Content-Type: application/x-www-form-urlencoded
Cookie: csrftoken=pUvfgN5WhUs09f53I7hYkBYJH1pGpQRK
Content-Length: 39

id=597f74ba-710e-42f8-9af1-876ccaea2b6d
```

This is the result in case of an existent user:
```JSON
{
    "message": "User deleted successfully"
}
```

And this error in case of an error:
```JSON
{
    "error": "User not found"
}
```

### Related issue
- #736


to-sta:
@jaymeklein Made a few more changes to this. 

A bigger thing there was that the model for the `User` was not setup properly. Now we are properly extending the built-in `User` with the `UserModel`. Previously for example the pw was not hashed. 

I seperated the deleted method in order to simplify it a bit. A user has to be logged in to be able to deleted his account, now the permissions class makes `IsAuthenticated` makes sure of that. 

We should test again and the mypy errors are not fixed yet. I will check that out tmw. Btw in case you have any questions just ask 😃

andrewtavis:
Checking in with the both of you, @to-sta, is the expectation that @jaymeklein fixes the mypy errors? Really appreciate all the work here!

to-sta (approved):

andrewtavis (approved):
Regardless of failed backend test, we're going to merge this as on our end it's not looking like the given lines are problematic. @to-sta will look into the errors if they persist on merge :)

Thanks, @jaymeklein! Really appreciate you getting to this! 🎉

Issue #770, opened 2024-03-21
Add TailwindCSS prettier plugin and reorder classes


lucianmurmurache:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

A great way to ensure that tailwind behaves exactly the way we intend it to, is to add this [plugin](https://www.npmjs.com/package/prettier-plugin-tailwindcss) which follows the recommended [class sorting](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted).

The PR will be a bit annoying to review as a lot of files will change, however I argue that it would be a worthwhile investment into ensuring that the styles are always applied correctly and that there are no unexpected overrides.

lucianmurmurache:
@andrewtavis if you think headwind would be enough here, we can also keep it as is. I would be in favour of adding the plugin instead as it enforces the rules pretty much out of the box and not relying on an extension that someone can omit to add. If you agree here, I can have it ready by the end of today.

andrewtavis:
Sounds perfect, @lucianmurmurache! I'll get it merged by the sync :)

andrewtavis:
Closed via #772 🥳 Thank you, @lucianmurmurache!

PR #772, opened 2024-03-22
chore: add prettier plugin tailwindcss

lucianmurmurache:
### Contributor checklist

---

### Description
Add [prettier-plugin-tailwindcss](https://www.npmjs.com/package/prettier-plugin-tailwindcss)
Formal all files to avoid complications in future PRs

### Related issue

- #770

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,7 +4,6 @@
     "bradlc.vscode-tailwindcss",
     "charliermarsh.ruff",
     "esbenp.prettier-vscode",
-    "heybourn.headwind",
     "streetsidesoftware.code-spell-checker",
     "Vue.volar"
   ]
--- a/frontend/components/card/CardTextEntry.vue
+++ b/frontend/components/card/CardTextEntry.vue
@@ -1,7 +1,7 @@
 <template>
-  <div class="px-5 py-5 card-style">
+  <div class="card-style px-5 py-5">
     <div class="flex flex-col gap-3">
-      <label for="feedback" class="text-left responsive-h3 font-display">
+      <label for="feedback" class="responsive-h3 font-display text-left">
         {{ header }}
       </label>
       <p v-if="description">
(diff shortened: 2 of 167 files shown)


.vscode/extensions.json
```diff
@@ -4,7 +4,6 @@
     "bradlc.vscode-tailwindcss",
     "charliermarsh.ruff",
     "esbenp.prettier-vscode",
-    "heybourn.headwind",
```
andrewtavis, line 7: Nice, thanks for this!

andrewtavis: We need to also remove this from the readme and contributing guide as well, but I'll get those 😊 There's a dropdown with suggested extensions in each :)

frontend/components/card/CardTextEntry.vue
```diff
@@ -20,7 +20,7 @@
         <!-- Desktop -->
         <textarea
           id="feedback"
-          class="hidden w-full px-3 py-2 overflow-y-scroll bg-transparent border border-white resize-none rounded-md md:block dark:border-white md:resize-y md:overflow-hidden"
+          class="hidden w-full resize-none overflow-y-scroll rounded-md border border-white bg-transparent px-3 py-2 md:block md:resize-y md:overflow-hidden dark:border-white"
```
andrewtavis, line 23: Just a note to myself that we have `border-white` for each here and that this should be a custom color of some kind. I'll get to this with the text component work.

.vscode/extensions.json
lucianmurmurache: @andrewtavis I can make a quick update for the readme and contributing and update this PR. I forgot to check for any headwind mentions.

andrewtavis: Sure thing, thanks @lucianmurmurache! I'm halfway through 🥴😅

lucianmurmurache: ✔️

andrewtavis (approved):
Holy smokes was that a review 🤯 In checking through it all, I saw no cases where a class element was omitted, changed or added.

Thanks so much for this, @lucianmurmurache! Headwind was at least a good bandaid for the interim such that this review was a lot easier, but so happy to have a solution that will enforce the styles 😊 Would be great if you'd explain this a bit in the sync!

PR #773, opened 2024-03-23
fix: prettier plugin tailwindcss

lucianmurmurache:
### Contributor checklist

---

### Description
This is a fix to a bug I created 😅 
Now the prettier tailwind plugin will take into consideration the tailwind config content and I moved the prettier dependencies to dev (I initially used `-D` instead of `--dev`, muscle memory from npm )

### Related issue

- #770


andrewtavis (approved):
Based on spot checking and the review in #772 where no changes were seen, I'm confident that we're fine in this one as well 😊 Thanks @lucianmurmurache!

Issue #781, opened 2024-03-23
Prettier Tailwind sorting breaks pre-commit


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

An issue that has come up since #772 is that [pre-commit](https://pre-commit.com/) is failing for some contributors. From what we understand, some unix contributors (Mac/Linux) cannot run pre-commit with the Prettier as we're getting an error for the [prettier-plugin-tailwindcss](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) plugin. It's expected that there could be issues for Windows users as well. The current error is:

```bash
prettier.................................................................Failed
- hook id: prettier
- exit code: 1

[error] Cannot find package 'prettier-plugin-tailwindcss' imported from /PATH_TO_PROJECT/activist/noop.js
```

As of now the Prettier run within [.pre-commit-config.yaml](https://github.com/activist-org/activist/blob/main/.pre-commit-config.yaml) has been commented out so that people can continue to contribute.

andrewtavis:
CC @wkyoshida, @to-sta :)

lucianmurmurache:
Current output during commit in #782
```bash
ruff.................................................(no files to check)Skipped
ruff-format..........................................(no files to check)Skipped
prettier.............................................(no files to check)Skipped
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...........................................(no files to check)Skipped
check for added large files..............................................Passed
```
The **soLucian** seems to have been converting the pretier config file into `js` as well as using `module.exports = config;` instead of `export default config` I am not sure exactly why it caused any errors with the plugin declaration in `.prettierrc` or `export default config` but at this point I am not sure I want to know. Hopefully it works on your side @andrewtavis & @wkyoshida

andrewtavis:
394cc2d ultimately closes this as [pre-commit/mirrors-prettier](https://github.com/pre-commit/mirrors-prettier) that we were using for this was archived as changes in Prettier broke it beyond maintenance capacity. In looking a bit into this, if we do want Prettier running in pre-commit hooks, then the current method is apparently to use Husky for this, but this'd be out of scope. I think we're fine with the current setup of checking it in CI for now 😊

PR #782, opened 2024-03-24
fix: pre commit prettier plugin tailwindcss

lucianmurmurache:
### Contributor checklist

---

### Description
While searching for similar issues, one suggestion was to change the prettier config file to js/ts which seems to have fixed the error.

### Related issue

- #781


lucianmurmurache:
Currently the only way I found to push prettier to `prettier` `prettier-config-tailwindcss` is to move the config to a js/ts file. This seems to brig some issues to [pr_ci_frontend](https://github.com/activist-org/activist/actions/runs/8409986990/job/23027897431) which I cannot seem to shake off. @to-sta @wkyoshida @andrewtavis any input/suggestions?

frontend/package.json
```diff
@@ -53,10 +55,8 @@
     "pinia": "^2.1.6",
     "qrcode": "^1.5.3",
     "qrcode.vue": "^3.4.1",
-    "reduced-motion": "^1.0.2",
+    "reduced-motion": "^1.0.4",
```
andrewtavis, line 58: Nice 😉

lucianmurmurache: I noticed a warning between the time I uninstalled tailwindcss and installed it as a dev dep. that reduced-motion needed that dependency, I initially wanted to add tailwind support and added it in peerDependencies but later realised that it didn't need anything to work with tailwind. This was the reason for the update there

lucianmurmurache:
Prior to testing, please make sure to run `pre-commit clean` to start fresh 🧹

andrewtavis:
Switching it back to `ts` as the `js` config wasn't working for me. Same errors as before, and I did run `pre-commit clean` :) Played around with a couple things, but `pre-commit run --all-files` does run successfully locally.

andrewtavis:
Seems like the same error as before it was switched to a JS file, but this seems doable somehow??

@wkyoshida and @to-sta: do either of you have suggestions? And the file changes seem to be adding in a bunch of commas to things. I figure we can remove this as something that's changed?

lucianmurmurache:
@andrewtavis I am using `pre-commit 3.6.2` and did not have any issues 🤔 
At this point it's probably good to question whether all this effort is worth it for the plugin and perhaps we can consider reverting to `headwind`.

andrewtavis:
Let me check my pre-commit version later, @lucianmurmurache. Big thing for Headwind is whether we can automate it, as as of now I need to go through and run it on certain files. Even the current setup where the check is run on PRs and we can expect an improvement from an actively maintained library would be preferable to the manual runs :)

lucianmurmurache:
@andrewtavis since we rely on manual installation of the extension, I am not aware of any way to enforce headwind. 

Another thought would perhaps be to rethink the usage of pre-commit on frontend. If we set the prettier (and prettier-plugin-tailwindcss) rules as well as checks that occur during the workflow, do we need another step? One down-side I see is that if we all have different pre-commit versions (assumption here, I might be wrong), it will not be easy to manage (I pulled your recent changes to this branch and if I run `pre-commit run --all-files` they change again 🤷‍♂️ ). On the other hand, we can easily dictate the versioning of dependencies and prettier rules.

It's likely that I have the wrong setup, I don't know how we can sync into the same rules via pre-commit, it uses the same pre-commit-config file that targets the same prettier rules, yet the result is not the same.

to-sta:
@andrewtavis What about writing and commiting the changes within the `pr_ci_frontend` workflow? The could keep the new solution 😃 but have to remove prettier from pre-commit 😢. 

@lucianmurmurache devcontainers are a way of enforcing extensions but I would prefer staying IDE agnostic.

andrewtavis:
Ok, well, that was worth a minor try :)

I really do not want to move back to Headwind as even with this solutions shortcomings, it's still enforcing the rules on PRs in a way that the old system never could. Issue is that with the TS config I'm for some reason getting that the `es5` trailing comma rule isn't recognized? JS config it's giving the old errors for me at least.

Let's leave it as is. Within the `pr_ci_frontend` workflow we get errors, and we as maintainers know that we need to go back and fix them. Importantly at least for me, autosave is formatting it correctly with the current setup on main, whereas for JS and TS based config files I actually lose code formatting and class ordering 🤷‍♂️

Let's leave the issue open and we can pick it up at another time! It'll get there eventually! 😊

Issue #671, opened 2024-01-28
Implement Groups API


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In addition to the organizations API for #627, we also need a similar API for groups that are sub-organizations. Many of the model elements will end up being the same, but there are some minor differences. The schemas for groups can be seen [on Figma here](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=5377%3A96660&mode=design&t=GLVfWMR4HNeoHP0L-1).

### Contribution

Happy to support on this!

andrewtavis:
CC @to-sta :)

kurtisolo:
Unfortunately, as things stand, I don't have enough free time to delve into the intricate world of the project's APIs. However, I can suggest an approach that could simplify the work as the various `ViewSets` are developed. Certainly, as development continues, the files will start to become huge again and at some point difficult to navigate. An approach that I use in my platform, which makes my work much easier, is to divide the apps into sub-folders.
Practical example:
```
entities/
  |
  +--groups/
    |
    +--__init__.py
    +--views.py
    +--serializers.py
    +--whatever_else.py
  |
  +--organizations/
    |
    +--__init__.py
    +--views.py
    +--serializers.py
    +--whatever_else.py
  |
  +--__init__.py
  +--admin.py
  +--models.py
  +--urls.py
  +--etc.py
```
This way, navigating the backend would be much simpler, immediate, and intuitive. This would not entail migrations or other structural problems; it's simply a way to better organize the code. This is a long-term approach: surely at some point the various `views.py` files will become huge, so it might be wise to nip the problem in the bud.

My two cents =)

andrewtavis:
Yes splitting it up where we can makes sense, @kurtisolo! Thanks for the suggestion :)

andrewtavis:
Good chance that myself or @to-sta will work on this, so we'll keep this in mind 😊

thiagogquinto:
Hey @andrewtavis, could I try to work on this one?

andrewtavis:
Hey @thiagogquinto 👋 @to-sta and I checked in and it'd be great to have you work on this one. Let us know if there's anything you need!

[...]

PR #783, opened 2024-03-24
Group api

thiagogquinto:
### Contributor checklist

---

### Description

- Added the HTTP methods for the Group model
- The validations for the Group were removed
- The model OrganizationApplicationStatus was removed

### Related issue

- #671


backend/entities/views.py
```diff
@@ -1,250 +1,250 @@
...
+        
+        if request.user != group.created_by:
+            return Response({"error": "You are not authorized to update this group"}, status.HTTP_401_UNAUTHORIZED)
+        
+        serializer = self.get_serializer(group, data=request.data, partial=True)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+    
+    def destroy(self, request: Request, *args: str, **kwargs: int) -> Response:
```
to-sta, line 199: @thiagogquinto 
When we are deleting the group, we want to make sure that only the creator is allowed to do that. 

@andrewtavis 
At the moment, when we delete the group, we set the deletion_date. But there should be more to this, like:

1. Delete the group and remove the `deletion_date` field
2. Set fields to blank/NULL on deletion

to-sta (changes requested):

backend/entities/views.py
andrewtavis: I don't think we need to worry as much about keeping minor data for groups as we do for for orgs. I think the first option would be fine? Saves on storage as well.

andrewtavis: But would you suggest either, @to-sta?

andrewtavis:
Thanks for the review, @to-sta! @thiagogquinto, let us know if you need any help with the refactoring or if there are any questions!

backend/entities/views.py
to-sta: I guess it's not necessary to set values to blank/NULL. I would suggest to delete the entry and remove the deletion_date field.

to-sta (approved):

andrewtavis (approved):
Thanks for the fixes here, @to-sta, and for the continued work, @thiagogquinto! Hope to see you at the dev sync on Saturday! 😊

Issue #556, opened 2023-11-18
Add selection state to tags in GridFilterTags


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We have the component [GridFilterTags](https://github.com/activist-org/activist/blob/main/frontend/components/grid/GridFilterTags.vue) that provides tags for filtering search and page results. We as of now don't have styling added for when one of the options is selected. Potentially look into refactoring this component with a Headless UI component like [radio group](https://headlessui.com/vue/radio-group) would be good to achieve this, but we'd also need to make sure that we allow for selecting more than one option and also deselecting an option.

### Contribution

Happy to support with this or get to it myself eventually! This is a great `good first issue` for someone wanting to contribute to activist for the first time, and is of course open for those already in the contributor base. 

Let's first discuss the implementation and then we can go from there! 🚀

keythroy:
I can try to help on this one

andrewtavis:
Thanks @keythroy! Just assigned you :) Let's be sure to check whether [radio group](https://headlessui.com/vue/radio-group) can allow for selecting more than one option and deselecting all of them before we get too deep into this.

Looking forward to this!

andrewtavis:
Hey @keythroy 👋 Hope all's well! Let us know if there's anything we can do to support with this! 😊

UnknownSean8:
Hi @andrewtavis, I can work on this, thanks! :P

andrewtavis:
Thanks @UnknownSean8! Let me know if there's anything I can do to support :)

UnknownSean8:
Hi @andrewtavis, sorry for the delay! I will push the changes by this Sunday at the latest. :P

andrewtavis:
No stress, @UnknownSean8! Looking forward to them :)

UnknownSean8:
Hi @andrewtavis, I was wondering if it's better to use the previous CardTopicSelection component to implement this so we have a more consistent component. I saw the design in Figma it's rather similar, I can modify the current component to be more customisable.

Also, I can use the Switch component in HeadlessUI to implement the GridFilterTags, it follows the check box accessibility except the arrow keys navigation. But ofc, I can also implement the arrow keys navigation here.

[...]

PR #785, opened 2024-03-29
feat: grid filter tags with a11y

UnknownSean8:
### Contributor checklist

---

### Description

Grid Tags that are similar to topics with a11y.

### Related issue

- #556

### DIFF
--- a/frontend/components/grid/GridFilterTags.vue
+++ b/frontend/components/grid/GridFilterTags.vue
@@ -6,10 +6,9 @@
           <div
             class="style-cta elem-shadow-sm flex h-max w-max cursor-pointer items-center justify-between space-x-2 border rounded-lg py-1 px-4 sm:py-0 sm:px-2"
             :class="{
-              'style-cta': checked,
-              'style-cta-secondary': !checked,
-            }"
-          >
+    'style-cta': checked,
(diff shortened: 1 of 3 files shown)


UnknownSean8:
I'm having problems formatting using prettier due to the prettier tailwind plugin. 😦

andrewtavis (approved):
My commit adds in the needed formatting and makes some minor changes, @UnknownSean8 😊 Big thing was adding in selection on return for the switches :)

Thanks so much for this! Great that it's all looking so consistent now!

UnknownSean8:
Thanks for helping to fix it! :)

I'll try to install pre-commit and see how it goes.

andrewtavis:
pre-cmmit wouldn't have helped with this, sadly. As of now it's turned off for Prettier until we get some bugs with the Tailwind ordering figured out, but should be there soon :) Maybe check that you have the [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extension installed and set it to autoformat on save? It should follow our eslint rules such that things will be corrected as you go. I literally did nothing to fix the formatting except save the file when I brought it down. You know that it's all working when you can screw up the formatting and the Tailwind ordering and it'll get fixed on save ✨

Issue #673, opened 2024-01-28
Add MapLibre custom map option to clear directions


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The new map component can be seen at the [about page for events](https://activist.org/en/events/123/about). We're using [MapLibre GL JS](https://github.com/maplibre/maplibre-gl-js) for it, which is a fork of MabBox (some examples can be used). The user is able to click on the map to get directions, and the points and lines can also be dragged to change the route. One problem is that if the user clicks too much then the map gets messer, and there's no way to remove points. It would be great if we created a custom map option in [the MediaMap component](https://github.com/activist-org/activist/blob/main/frontend/components/media/MediaMap.vue) that would then call the following function:

```ts
function clearDirections() {
  directions?.clear();
}
```

### Contribution

Happy to discuss and help with implementation!

andrewtavis:
This option should appear clearly in the bottom left corner of the map as soon as someone starts making a route :)

andrewtavis:
01b82e3 adds the needed code for a button that says `Click to clear directions` on mobile screens, but it's been commented out as the button isn't clickable. Would be great if someone wanted to add this event listener functionality to this!

andrewtavis:
@GrandSuccess87, here's the next issue we were talking about for the map :)

nicholas-vecchio:
I'm happy to work on this @andrewtavis :)

So if I understand correctly, all that's left is to allow the clear all directions functionality to be clickable/touchable?

andrewtavis:
Exactly, @nicholas-vecchio :) We need to be able to click it, and from there the code to show the clear directions button on mobile should be uncommented. Thanks for picking this up!

andrewtavis:
Just checked on mobile, and all's good, @nicholas-vecchio! So amazing! Thanks again so much 😊😊

If you're interested in continuing to work on the map, the work you did here just unblocked #718 🥳 I'll write a bit more into that one now as I've just finished the designs for it 🧑‍🎨

PR #791, opened 2024-04-07
#673/Added events to MediaMap for clearDirectionsControl.

nicholas-vecchio:
### Contributor checklist

---

### Description

Uncommented mobile clearDirectionsControl div and allowed it to be conditionally rendered within addControl.

Added events within addControl to allow touch/click (depending on platform).

Added class to the created Div as 'maplibregl-ctrl' is required to allow pointer events, as by default all pointer events on the map are disabled, also removed margin as it was making clicks trigger on margin. The 'maplibregl-ctrl' class handles adding a margin without this issue.

### Related issue

- #673

### DIFF
--- a/frontend/components/media/MediaMap.vue
+++ b/frontend/components/media/MediaMap.vue
@@ -187,6 +187,7 @@ onMounted(() => {
                 if (window.innerWidth < 768) {
                   div.innerHTML = clearDirectionsControl;
                   div.addEventListener("touchend", () => directions.clear());
+                  div.addEventListener("click", () => directions.clear()); // for small desktops or tiling
                 } else {
                   div.innerHTML = clearDirectionsHotkeyControl;
                   div.addEventListener("click", () => directions.clear());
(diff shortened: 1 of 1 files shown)


frontend/components/media/MediaMap.vue
```diff
@@ -153,56 +153,55 @@ onMounted(() => {
...
-                  // if (window.innerWidth < 768) {
-                  //   div.innerHTML = clearDirectionsControl;
-                  // } else {
-                  //   div.innerHTML = clearDirectionsHotkeyControl;
-                  // }
+          map.addControl(
+            {
+              onAdd: function () {
+                const div = document.createElement("div");
+                div.className = "maplibregl-ctrl";
```
andrewtavis, line 186: Thanks for all the research in getting to this, @nicholas-vecchio!

andrewtavis (approved):
Thanks so much for this, @nicholas-vecchio! Via my commit I added a click listener for the mobile view as well, just in case someone's using the mobile view via tiling on desktop :)

This basically has us at production ready for the map, except for one other issue that's already being looked into! 😊

Issue #699, opened 2024-02-10
Add Facebook to Modal Share Page Component


GrandSuccess87:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue will focus on adding support to share activist pages (event/organization/group) on Facebook via the modal share page component.  

This component lives in [frontend/components/modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal). 

Icons would ideally come from one icon source on [Icones](https://icones.js.org/) such as [CoreUI Brands](https://icones.js.org/collection/cib).  

You may also find the icons you need on [Bootstrap](https://www.figma.com/file/p0nFlH6k1C9rTBjYjjiAUV/Bootstrap-Icons-(Community)?type=design&node-id=0-1&mode=design&t=vUxUgAs8hBzCRHYc-0).

### Contribution

Happy to pair up with someone on this or tackle it myself :))

GrandSuccess87:
Note: [Modal Share Page Component #581](https://github.com/activist-org/activist/issues/581) is in progress

andrewtavis:
I see you took my advice to do them all at once, @GrandSuccess87 😅😊 Thank you for all of these and spearheading such an important feature!

GrandSuccess87:
Of course! Hahaha! It was the better option indeed 😆

thiagogquinto:
Hey @andrewtavis, I was thinking in trying to work on the issue, and because of that I took a look on the code and noticed that there is already a code for the facebook, and to work I only had to import the ```SFacebbok```, and looks like it's working, so could I make a PR for it or is there something that we can't use in the code mentioned?

andrewtavis:
I'll let @GrandSuccess87 give some feedback on what the status is here, @thiagogquinto :) Would be great if you could help out here though!

GrandSuccess87:
Hii @thiagogquinto, although there's pre-existing code, you may find you run into an issue with one of the expected properties when you test it. It may just be a small one that's easily fixable.

andrewtavis:
PR is merged, but let's not close this quite yet. Currently when I try to use it I get the following:

[...]

PR #798, opened 2024-04-12
Add Facebook in Modal Share

thiagogquinto:
### Contributor checklist

---

### Related issue

- #699


andrewtavis (approved):
It actually doesn't link into Facebook because of the following:

Screenshot_2024-04-14 18 21 01

But I think this is great 😊 Merging this in for now, and we can play around with the options for it so it works before we close the issue :)

andrewtavis:
Thanks for the help as always, @thiagogquinto!

PR #824, opened 2024-04-20
Signup API tests

to-sta:
### Contributor checklist

---

### Description

I added the tests for the Signup endpoint. During the testing I fixed a few issues in the model and serializer (e.g. Passwords were not hashed). 

For backend tests in general I would recommend to use the docstring for documenting the scenarios that are tested.

I ran pytest to see if it works correctly. 

I also saw that we have failing throttling tests, that we should look into.

### Related issue

- #432 #736

### DIFF
--- a/backend/authentication/tests.py
+++ b/backend/authentication/tests.py
@@ -7,6 +7,10 @@ from .factories import (
     UserTaskFactory,
     UserTopicFactory,
 )
+from .models import UserModel
+
+from django.test import Client
+from faker import Faker
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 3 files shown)


backend/authentication/tests.py
```diff
@@ -24,3 +28,106 @@ def test_str_methods() -> None:
...
+            "username": username,
+            "password": strong_password,
+            "password_confirmed": strong_password,
+            "email": email,
+        },
+    )
+
+    assert response.status_code == 201
+    assert UserModel.objects.filter(username=username).exists()
+    assert UserModel.objects.get(username=username).password != strong_password
```
andrewtavis, line 106: Checking in here as I'm assuming the test passed, but then to me it would make sense that this is `==`?

backend/authentication/tests.py
```diff
@@ -24,3 +28,106 @@ def test_str_methods() -> None:
...
+    """
+    # Setup
+    fake = Faker()
+    username = fake.name()
+    second_username = fake.name()
+    email = fake.email()
+    strong_password = fake.password(
+        length=12, special_chars=True, digits=True, upper_case=True
+    )
+    wrong_password_confirmed = fake.password(
```
andrewtavis, line 58: Want to check really quick on Faker here as the percentage chance that the passwords generated would be the same when generated at random is ~0, but is there also something in Faker that within a given instance it's not going to return the same thing twice? Just wondering.

andrewtavis:
Happy to fix the merge conflict, @to-sta! Looks like a keep both that just wasn't managed properly :)

backend/authentication/tests.py
to-sta: @andrewtavis Thanks for review 😃. 

Since the `exists()` method will return `True` or `False`, it is not necessary to write `==`.

andrewtavis: What I mean is for `assert UserModel.objects.get(username=username).password != strong_password`. To me we should be asserting `==`?

backend/authentication/tests.py
to-sta: Every time you invoke the method it will generate a random password. It is very unlikely to get the same password for our cases, but why would you not want the chance of the same password? Users also may also have the same password without knowing it, so do our tests 😄

andrewtavis: Just mean that for that for the tests in this case we'd want this to 100% not be `strong_password`, and it's 99.99999%, so all good, but was wondering if there was something that meant that on instantiation we'd definitely not generate the same thing twice. Thanks for the explanation!

backend/authentication/tests.py
to-sta: Ah ok, I see. Maybe I should have placed a comment above there 😅.

There I am checking if the password is hashed. When I retrieve the password from the DB it should be hashed and therefore not the same as the raw password anymore.

andrewtavis (approved):
Thanks for all this and the explanations, @to-sta! I'll add in the comment as discussed 😊

Issue #829, opened 2024-04-21
Implement base Playwright testing for the frontend


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In #775 we were implementing a baseline Cypress integration with the project, but the decisions was ultimately made to switch over to [Playwright](https://playwright.dev/). The goal of this issue would be to remove all files and dependencies brought in by #777, and then implement a base route test with Playwright. Once this issue is closed, we'd then create individual issues for first making tests to check that needed elements are appearing on pages in the needed forms, and from there individual tests for interactions on the platform being possible.

The work for this issue would also include adding the run of the tests to [pr_ci_frontend](https://github.com/activist-org/activist/blob/main/.github/workflows/pr_ci_frontend.yaml) 😊

Work for this generally falls within #432 as an epic :)

### Contribution

CC @aasimsyed :) Happy to discuss further and help as I can!

andrewtavis:
@aasimsyed, feel free to write in here and I'll assign 😊

aasimsyed:
ok. thanks, @andrewtavis

andrewtavis:
Closed via #838 😊 Thanks so much for the immense effort into making this happen, @aasimsyed! Let me know if any of the following issues would be of interest:

- #724
- #871 
- #872

We can then schedule a call for after the 13th when I'm back to discuss making further testing issues 🚀

PR #838, opened 2024-04-24
Chore/implement playwright axe and axe core

aasimsyed:
### Contributor checklist

---

### Description

This PR removes Cypress and adds Playwright in its place as well as updates the GitHub Actions workfile `pr_ci_frontend.yaml` for CI testing. 
* This assumes that secrets are set up in the project under the `dev` environment for it to work.

#### Changes in this PR:
- implemented Playwright with basic tests for Landing Page 
 -- uses a Page Object Model design pattern
 -- uses scripts to run set in `frontend/package.json`
 -- to run against local (assuming server is running): ```yarn test:local```
 -- to run against prod: ```yarn test:prod```
 -- to run either using Playwright UI, append `--ui` on either command
- updated `pr_ci_frontend.yaml` to include server start/stop and run playwright tests
- updated `docker-compose.yml` to include health check
- tested Playwright tests using `yarn test:local`, `yarn test:prod`, and `yarn playwright test` and 
- tested GitHub Actions for frontend on branch

### Related issue

- #[829](https://github.com/activist-org/activist/issues/829)


aasimsyed:
for this PR to pass, the `pr_ci_frontend.yaml` and the project's GitHub secrets/environment will need to match.

frontend/package.json
```diff
@@ -43,7 +43,8 @@
     "vue-tsc": "^2.0.11"
   },
   "dependencies": {
-    "@maplibre/maplibre-gl-directions": "^0.7.0",
+    "@axe-core/playwright": "^4.9.0",
+    "@maplibre/maplibre-gl-directions": "^0.6.1",
```
andrewtavis, line 47: Note that the MapLibre downrade's happening here, which should be avoided 🤔

andrewtavis: I can handle this :)

frontend/tests/specs/landing-page.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { expect, test, LandingPage } from "../fixtures/page-fixtures";
+import AxeBuilder from "@axe-core/playwright";
+
+test.describe("Landing Page", () => {
+  // Initialize page before each test, wait for the landing splash to be visible
```
andrewtavis, line 5: Minor nit on this, and no stress, but we try to do complete sentences for comments if they're on their own line - so capped as you have it and with a period. If it's inline, then no cap and no period :) I'll get these 😊

frontend/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [main, master]
```
andrewtavis, line 4: Let's definitely remove references to `master` here, but why are we doing a separate YAML file for this? Just checking as there's lots of Playwright stuff in the frontend CI :)

aasimsyed: this is the example "boilerplate" file... not used in practice I believe.

frontend/package.json
aasimsyed: I think this was from fixing a merge conflict mistake. My bad.

frontend/.github/workflows/playwright.yml
aasimsyed: it gets created on `yarn create playwright`

andrewtavis:
Hey @aasimsyed 👋 Discussing all of this now with @to-sta and @wkyoshida :)

- @to-sta was saying that as far as secrets are concerned, we don't need to worry about this as we can just use random strings
- @wkyoshida was saying that it'd be best to be running this off of a staging backend at some point, but we all agree that for now this setup is fine so long as we're only running it from time to time
  - Reason for this is a worry about the GitHub Actions allotment
- We can switch this over to something that we trigger explicitly
  - Making a different workflow like `e2e.yaml` would likely be better (I had said we should include it in the frontend PR workflow, but given the timing it's taking we should make a single one)
  - Running this automatically within the release to deployment workflow would make sense, but we can make another issue for that as we start to develop that workflow

Let me know if you'd like to connect on all of this in the future. After Wednesday would be ideal, and I'm off Thursday and Friday 😊

aasimsyed:
@andrewtavis please take a new look at this. I have addressed your comments and restored the `pr_ci_frontend.yaml` from `upstream/main`, creating a separate `pr_ci_playwright_e2e.yaml` for Playwright E2E that is not automatically triggered (while it's a WIP). I made the Playwright one with the below trigger, which I think adds a button to manually run on GitHub if it's on the default branch: 
```on:
  workflow_dispatch:```
  
  I also added a couple of ids on `LandingSplash.vue` for easier locators, if that is OK.

andrewtavis:
Thanks, @aasimsyed!

andrewtavis (approved):
All's looking great, @aasimsyed! And thanks so much for handling the merge conflicts yourself 😊 Note on this, @to-sta and @wkyoshida:

- We discussed on Thursday and we're thinking that although expansive, this is the best way to do things at this point as we don't have access to a staging environment to run the tests on
- We need to trigger this workflow as it's set with `workflow_dispatch`
  - I've already added a check for the maintainer workflow that this and the new [OWASP Zap workflow](https://github.com/activist-org/activist/blob/main/.github/workflows/owasp_zap_full_scan.yaml) should be ran on PRs :)
- Especially for the situation where we're running on Netlify, it's a bit difficult to link the deployment previews
- Once we do switch away from Netlify, a more expansive system like this will be necessary, so it's best to keep it for now rather than rewrite it later

PR #852, opened 2024-04-29
updated usermodel integration for the admin panel

to-sta:
### Contributor checklist

---

### Description
In our current setup the user creation thru the admin panel is not possible since the custom user model is not used properly. With this PR it is possible to create a user in the admin panel and change settings.

### Related issue

- No related issue


andrewtavis (approved):
Thanks for all of this, @to-sta! 😊 I'll check the other one now too :)

Issue #842, opened 2024-04-25
Change signed out sidebar cta to "Join activist" and style


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This is a great `good first issue` for someone wanting to learn a bit more about activist development. On pages like the [home page](https://activist.org/en/home/) we have a `Join` option in the bottom left of the page. It'd be great if this were changed to "Join activist", but specifically Tailwind styling needs to be added such that there's no line break/glitch when the sidebar expands and retracts. 

Specifically:
- If we change the text by making a new key in [en-US.json](https://github.com/activist-org/activist/blob/main/frontend/i18n/en-US.json), then there's now two words instead of just one
- Then as the sidebar expands and retracts, for a moment the text will be on two lines
- The height of the dropdown thus increases for a moment, which is pretty unappealing glitch

Would be great to have it be "Join activist" so it's a bit more descriptive!

### Contribution

Happy to support someone with interest in working on this 😊

amilliiee:
okay! now that the world isn't ending I can turn my attention to this :)

andrewtavis:
Great to have you, @amilliiee! Some further details here:

- We'll want to make a new key that's in the `en-US.json` file that's `"Join activist"`. As of now we're using `components._global.join`.
- The i18n key is used in [components/dropdown/DropdownUserOptions.vue](https://github.com/activist-org/activist/blob/main/frontend/components/dropdown/DropdownUserOptions.vue), which is where the new key will go
  - We can thus name the new key `components.dropdown-user-options.join-activist`
- Once we apply that new key, you'll see that there's a glitch with the text breaking to two lines and thus the button being higher
  - We want to avoid this by changing the CSS on this component or parents of it such that the text won't break onto a new line

Let me know if further information would be helpful! :)

andrewtavis:
[...]

PR #854, opened 2024-04-30
Sidebar "join" button text change

amilliiee:
### Contributor checklist

---

### Description

This PR updates the "Join" button on the home page to "Join activist" and fixes the button resizing issue that appears with the change.

The updates were:
- added new key in en-US.json
- updated DropdownUserOptions to use new key
- included whitespace-nowrap in DropdownBase MenuButton

### Related issue

- #842

### DIFF
--- a/frontend/components/dropdown/DropdownBase.vue
+++ b/frontend/components/dropdown/DropdownBase.vue
@@ -4,7 +4,7 @@
       <MenuButton
         @focus="expandOnFocus"
         v-slot="{ open }"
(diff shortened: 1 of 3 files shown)


andrewtavis (approved):
Very minor changes just came through, @amilliiee :) We needed a dot instead of a dash to separate the component from its usage in the i18n key, and the build error was because the Tailwind classes weren't ordered correctly. The [contributing docs](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md) should have some good info on how to get your env set up with the proper extensions and pre-commit where these things will be fixed for you (or at least the Tailwind bit).

Appreciate the first contribution! Really great working with you 😊

Issue #694, opened 2024-02-10
Add Signal to Modal Share Page Component


GrandSuccess87:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue will focus on adding support to share activist pages (event/organization/group) on Signal via the modal share page component.  

This component lives in [frontend/components/modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal). 

Icons would ideally come from one icon source on [Icones](https://icones.js.org/) such as [CoreUI Brands](https://icones.js.org/collection/cib).  

You may also find the icons you need on [Bootstrap](https://www.figma.com/file/p0nFlH6k1C9rTBjYjjiAUV/Bootstrap-Icons-(Community)?type=design&node-id=0-1&mode=design&t=vUxUgAs8hBzCRHYc-0).

### Contribution

Happy to pair up with someone on this or tackle it myself :))

GrandSuccess87:
Note: [Modal Share Page Component #581](https://github.com/activist-org/activist/issues/581) is in progress

Mister-Tofu-tech:
@andrewtavis @GrandSuccess87 I created the pull request to add the icon. But am not sure what URL we are linking to from it.

andrewtavis:
Hey @Mister-Tofu-tech 👋 Thanks for sending along the PR. I'll make some comments as there are things that would need to change, and then maybe we can figure out what URL it should point to and you can implement that as well :)

UnknownSean8:
Hi, I would be happy to work on this. Thanks!

andrewtavis:
@UnknownSean8 will be helping a bit on the share part of this once I bring in @Mister-Tofu-tech's PR :) Thanks, @UnknownSean8!

andrewtavis:
Ping @UnknownSean8 👋 With ea8f4fe and #855 being merged we're ready to work on the share functionality :) Thanks for the work here, @Mister-Tofu-tech!

andrewtavis:
Closed via #955 🚀 Thanks for the all the efforts here, all, and especially for the great execution, @UnknownSean8! We're not able to directly open Signal with a message, but copying the URL and opening the app should work well 😊

PR #855, opened 2024-04-30
Add Signal icon to Modal Share page

Mister-Tofu-tech:
### Contributor checklist

---

### Related issue

- #694


frontend/components/modal/ModalSharePage.vue
```diff
@@ -196,6 +200,7 @@ import type { Group } from "~/types/group";
 import type { Organization } from "~/types/organization";
 import type { Resource } from "~/types/resource";
 import type { User } from "~/types/user";
+import IconSignal from "../icon/IconSignal.vue";
```
andrewtavis, line 203: In Nuxt we actually don't need to import the components, so this can be removed (regardless of us not using a component anyway) 😊

frontend/components/icon/IconSignal.vue
```diff
@@ -0,0 +1,10 @@
+
+<template name="iconsignal">
```
andrewtavis, line 2: We actually don't need this component, @Mister-Tofu-tech :) My suggestion would be to use the built in `<Icon />` component via Nuxt Icon. You can pass in an icon that you find on [Icones](https://icones.js.org/) in the following way, with `simple-icons:signal` being one I'd suggest:

```
<Icon name="simple-icons:signal" size="1.5em" />
```

This can be directly dropped into `ModalSharePage`. We do not need a separate component for it :)

andrewtavis:
Once you do the changes above, @Mister-Tofu-tech, check out the [contributing guide](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md) for extensions that you can add that will format your code with Prettier. This will then clear up the test error that we're getting here. From there, @GrandSuccess87 and I can look into what the links for this can be, and we can pass on some info on what the next steps are! 🚀

frontend/components/icon/IconSignal.vue
andrewtavis: Hey @Mister-Tofu-tech 👋 Given the changes recently in #868, the code for this would now be:

```ts
<Icon :name="IconMap.SIGNAL" size="1.5em" />
```

andrewtavis (approved):
Thanks for the efforts here, @Mister-Tofu-tech! Appreciate the communication in the issue and the work to get Signal added as an easy page share method 😊

PR #859, opened 2024-05-01
added login function and changed serializer to use username instead o…

to-sta:
…f email

### Contributor checklist

---

### Description
In this PR, I have started working on the login. I have changed the serializer to use the `username` instead of the `email`. There are still some missing pieces:

* User input validation
* Error handling
* routing

<ins>Notes:</ins>
* I was thinking about the password indicator on login. Since the user already has an account and a password, it seems unnecessary to me.

* How are we showing login errors to the user? 

* I used the `$fetch` instead of `useFetch` because nuxt, displayed a suggestion in the console:

```
[nuxt] [useFetch] Component is already mounted, please use $fetch instead. See https://nuxt.com/docs/getting-started/data-fetching
```

I have tested a successful login and a bad request.

### Related issue

- No related issue

### DIFF
--- a/backend/authentication/serializers.py
+++ b/backend/authentication/serializers.py
@@ -136,11 +136,11 @@ class SignupSerializer(serializers.ModelSerializer[User]):
 
 
 class LoginSerializer(serializers.Serializer[UserModel]):
-    email = serializers.EmailField()
+    username = serializers.CharField()
     password = serializers.CharField(write_only=True)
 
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Thanks for all this, @to-sta! We can figure out the final interface a bit later, with it likely just needing to get props from the user interface in types or something along those lines 😊

Issue #858, opened 2024-04-30
Org and event index pages don't route to about on resize to the page being larger


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In #856 a bug was fixed that would route to a double id on an about change on resize the page to being smaller. The tradeoff though was that now the route doesn't change to about if we start on the small breakpoint and move to the larger ones. Changes for this would specifically need to happen within [pages/organizations/[id]/about](https://github.com/activist-org/activist/blob/main/frontend/pages/organizations/%5Bid%5D/about.vue) and [pages/events/[id]/about](https://github.com/activist-org/activist/blob/main/frontend/pages/events/%5Bid%5D/about.vue), with the [commit that fixed the bug](https://github.com/activist-org/activist/commit/f2fc0165328961b3591af3973534fe6c79f734dc) being a good place to check on all of this (resizing from small to large used to work).

This is a great `good first issue` for someone wanting to try working on activist! Happy to discuss and support as needed 😊

nicomaz:
Hi, I've been having a look at this and if it's still up for grabs, I'd love to work on it.

andrewtavis:
Closed by #860 🚀 Appreciate the contribution, @nicomaz! Would be happy to work with you again :) #857 is likely the same? Do you want to pick that up as well?

nicomaz:
Appreciate the opportunity and glad to help! Sounds good, I'll have a look now. :-)

andrewtavis:
Feel free to write in there if you're interested and I'd be happy to assign :)

PR #860, opened 2024-05-01
update route when changing from smaller to larger breakpoints

nicomaz:
### Contributor checklist

---

### Description
Route now changes to about when starting from a smaller breakpoint and moving onto a larger one. Specifically for [pages/organizations/[id]/about](https://github.com/activist-org/activist/blob/main/frontend/pages/organizations/%5Bid%5D/about.vue) and [pages/events/[id]/about](https://github.com/activist-org/activist/blob/main/frontend/pages/events/%5Bid%5D/about.vue). 

### Related issue

-  #858

### DIFF
--- a/frontend/pages/events/[id]/index.vue
+++ b/frontend/pages/events/[id]/index.vue
@@ -146,8 +146,11 @@ const handleResize = () => {
 
 onMounted(() => {
   handleResize();
-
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Tested and working perfectly, @nicomaz 😊 Thanks so much!

Issue #645, opened 2024-01-06
Open QR Code in new page on click


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In #473 we were able to add in the ability to download QR Codes for pages in multiple formats. This functionality can be seen on say the [about page of an event](https://activist.org/en/events/123/about) where clicking the QR Code icon will open the modal where the download functionality is. It would be great if we could add in one final functionality to this modal, which would be that the user could click the QR Code and have it open in a new tab. We'd need to alert the user that this would happen by applying `cursor-pointer` to `QRCode` in [ModalQRCode](https://github.com/activist-org/activist/blob/main/frontend/components/modal/ModalQRCode.vue), and beyond this adding a tooltip on hover that indicates that clicking will open it in a new page would also be helpful.

### Contribution

Happy to support someone who wants to work on this!

Vidhanvyrs:
hi @andrewtavis Can i try to work on this issue

andrewtavis:
Hey @Vidhanvyrs! Nice to hear from you again 😊 Sure thing! Please let me know if I can do anything to help :)

Vidhanvyrs:
hey @andrewtavis i need some help ! I just want to know how can i get the url of the qr code because my function is ready with the tooltip and all its just i was unable to find the url of the qrcode which we wanted to open in a new tab

andrewtavis:
This is a great question, @Vidhanvyrs :) Made the issue and then realized that this could be more complex than I expected 🤔 I guess that we want to be able to open the image that's being saved in a new window, but there's no URL for it as it's generated in the frontend. Is there a way of copying the functionality of "open in a new window" without a URL? I guess not. We could make a URL for it that's based on the route for the event or organization and then at that route it'd just be the code? Can you think of anything else?

Vidhanvyrs:
As we have a dynamically generated image on the frontend and we want that same image to open it in a new window, we can indeed create a temporary URL for it. One approach is to convert the image data into a data URL and use that as the source for the new window Now that is something to be researched iny give me some time may be I could come up with something that might solve it🙂

andrewtavis:
[...]

PR #887, opened 2024-05-23
feat: add ability to open QR code in new tab on click.

Robot8lover:
### Contributor checklist

---

### Description

Introduce basic functionality for opening a the QR code in a new tab as request in #645. Opens a new window containing an HTMLImageElement with the QR code as a png data url src. Tested on Chrome and Safari, and it works as expected though pop-up windows must be allowed.

### Related issue

- #645


andrewtavis:
Thanks for this, @Robot8lover! Will get to it by early next week at the latest :)

andrewtavis:
Something that I'm realizing here @Robot8lover is that no matter what ID we're on we're returning `1` for the ID for the QR code... This is the case for organizations, groups and events. Would you have interest in looking into this?

The only other comment is on the way the the `Open image in a new tab` label flickers when we hover over it itself. Could we figure out a fix for this? Specifically the cursor tends to be over it when the QR code modal is opened, so then the first thing the user experiences is the label flickering.

Really appreciate all the work here!

andrewtavis:
Also, let me know if you can see the errors in [the workflows](https://github.com/activist-org/activist/actions/runs/9201644423/job/25371910903?pr=887). Would be happy to assist with those as well! I'll be away for the weekend, but will be happy to assist and bring this in next week!

Robot8lover:
@andrewtavis I will take a look at that. And yes, I can see the errors, and I'll try to fix those.

andrewtavis:
Thanks @Robot8lover :) Let me know if there's anything I can do to help!

Robot8lover:
@andrewtavis I'm a bit confused what you mean when you say that no matter what ID we're on we're returning 1 for the ID. Where does this ID show up?

Robot8lover:
Oh, I think I know what you mean – the QR codes point to /en/events/1 (or organizations/1 or whatever it is) instead of what they actually are from. This is based on linkURL, which is based on useLinkURL from [frontend/composables/useLinkURL.ts](frontend/composables/useLinkURL.ts), which bases itself off of the props object. Within [frontend/pages/organizations/[id]/about.vue](frontend/pages/organizations/[id]/about.vue) and the other organizations/[id] pages is the line `const organization = testTechOrg` so I believe that that's what is setting the id to 1. This would no doubt change in production though I'm not quite sure how it works yet and what creating an organization/event would actually do.

andrewtavis:
Fantastic and totally makes sense, @Robot8lover! Was worried we had something hard coded, but if it's based on the ID of the object then we're good :) Thanks for the investigation!

Will get to the full review in the coming days 😊

andrewtavis (approved):
Thanks so much for this, @Robot8lover! Sorry that the review took as long as it did. Has been a very busy last couple of weeks, and I wanted to clear some other things so that I can focus more exclusively on activist 😊 Appreciate the discussions in the PR and the extra commits to get this all right! Will try to get the others in soon as well, and hope to be able to work with you more!

Issue #864, opened 2024-05-02
Remove Tailwind based component hiding in favor of TS breakpoint v-if conditional rendering


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Original issue:

Something that's come up is that it seems that [SidebarLeft](https://github.com/activist-org/activist/blob/main/frontend/components/sidebar/left/SidebarLeft.vue) and [MenuMobileNavigationDropdown](https://github.com/activist-org/activist/blob/main/frontend/components/menu/mobile/MenuMobileNavigationDropdown.vue) are running concurrently regardless of them being hidden at compliment breakpoints in the [sidebar layout](https://github.com/activist-org/activist/blob/main/frontend/layouts/sidebar.vue).

Further context:

The reason that this is happening is that just because the component is hidden via CSS doesn't mean that it's not loaded along with it's TS. What needs to happen is we need to switch the instance of `hidden md:block` and the like to conditional rendering based on [breakpoints in breakpoints.ts](https://github.com/activist-org/activist/blob/main/frontend/types/breakpoints.ts). With this, these components will then not be rendered at all, and we'll get dramatic speed increases for the platform 🚀

Robot8lover:
@andrewtavis Hi, is this just a matter of switching `hidden md:block` and `md:hidden` to `v-if` (with appropriate conditionals based on breakpoint) in the [sidebar layout](https://github.com/activist-org/activist/blob/main/frontend/layouts/sidebar.vue) or is there more to it?

andrewtavis:
Closed by #890 🚀 Thanks again for all of these amazing changes, @Robot8lover! The contributions have really been amazing, and specifically you chose some extremely important things to work on. Really hope that we can work with you again in the future, and I'd be more than happy to discuss the project with you at some point on a call if you have questions :)

Thanks again, and all the best!

PR #890, opened 2024-05-30
Fix component hiding

Robot8lover:
### Contributor checklist

---

### Description

Exchange instances of `hidden md:block` or `md:hidden` and similar CSS based hiding techniques for `v-if` statements using breakpoints. As mentioned in #864, this has the advantage of not rendering these components, which is better for performance and usually desirable unless the window is being resized. If accounting for resizing windows is preferred, though it is usually not the case, then `v-show` could be used instead.
I tried to keep any `display` property classes unless it was obvious they could be removed (e.g. block on a `div`). It is possible I reversed some of the logic due to a forgotten exclamation mark, which is one of the downsides of what I did. It may be better to explicitly write `greaterOrEqual` or `lesser` to avoid such errors. There were also some conditionally assigned hiding classes, which involve two boolean checks.

### Related issue

- #864


andrewtavis:
Really appreciate all the work here, @Robot8lover! We have the dev sync on Saturday that you'd be more then welcome to :) If you can't make the sync, also let me know and I'd be happy to set up a call and answer any questions you might have!

I'll try to bring all the PRs in over the coming days 😊

andrewtavis (approved):
This change really is great, @Robot8lover 😊 Exactly what we needed, and was really easy to follow! :)

I added/changed a few things:
- I updated the style guide so that people know to follow these conventions
- I changed the names of instances of `useBreakpoint` per your suggestion to make it a bit more clear in the code that the conditional has to do with breakpoints
  - Rather than `medium` we now have `aboveMediumBP`

From my view the conditionals have all been applied correctly, but we can fix something if need be later on!

frontend/components/header/HeaderWebsite.vue
```diff
@@ -10,7 +10,7 @@
     }"
   >
     <!-- MARK: Mobile Header -->
-    <div id="mobile-header" class="flex px-4 py-3 md:hidden">
+    <div v-if="!aboveMediumBP" class="flex px-4 py-3">
```
aasimsyed, line 13: removing this id breaks some tests

andrewtavis: This one's on me, @aasimsyed! Sorry about that!

Issue #876, opened 2024-05-05
Protect frontend routes based on user logged in status


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that’s needed is to make sure that certain routes aren’t reachable given the status of the user. Specifically the following routes should be blocked if the user isn’t signed in:

- /organizations/create
- /groups/create
- /events/create
- /resources/create

Signed in users should also be redirected to /home if they navigate to auth pages directly via a URL. Beyond this we have the settings pages, which should be protected eventually, but only after they switch over from being template pages. 

Suggestions for other pages that need attention would be welcome!

### Contribution

I’ll likely work on this once we get the final routes decided and once there’s more connection between the front and back end.

wkyoshida:
I imagine that protecting based on roles would also fall under here + simply doing things like only rendering certain components (or parts of components) based on whether user has certain roles

sulabhkatila:
Is adding a middleware that deals with these a feasible option?

andrewtavis:
Note from the sync: we can do an on mounted check and then directly route them to `auth/sign-in` if they for some reason don't have a token 🪙

UnknownSean8:
Hi, can I try to work on this feature, I have done some authentication/authorization in React but if it's possible I would like to try to implement it in VueJS as well. Thanks! :)

andrewtavis:
Sounds good, @UnknownSean8! This one would be more high priority than the map tooltip, so it'd be good if you'd try to get a solution for this up sooner. Let us know if there's anything we can do to help!

andrewtavis:
@momanyisamuel and I are talking about this now :) The best way to do this would be to enable Pinia and have the middleware handle the routes and their respective protections.

UnknownSean8:
Ahhh, but the current pinia is not fully implemented yet right, I'll be implementing it? Thanks!

UnknownSean8:
Okay! I'll try to attend if possible. As the time zone difference is pretty big rn as I'm back in my home country for a month. :/

UnknownSean8:
[...]

PR #896, opened 2024-06-03
feat/protect frontend routes

UnknownSean8:
### Contributor checklist

---

### Description

I just implemented the middleware in the events create route as I think most of the others are done. I also added a check in the admin guard as there's already a userIsAdmin variable in the store.

A few questions,
1. Is it better if we make the guard a global middleware instead of the current implementation where we put the metadata into each file that we want to protect? If we change it into a global middleware I think we can just put in the route name in the middleware file instead.
2. I don't understand the part where the signed-in user accesses the URL directly. Does this mean that the user is authenticated but is not an admin, so he will be redirected to the /home page or anyone who accesses the auth pages directly through the URL?

I might have missed it if it was discussed in the dev sync, sorry in advance and thanks!

### Related issue

- #876
- https://github.com/activist-org/activist/commit/3cf5fb2c25159194f0ba354489bd690f6ee5b4a1

### DIFF
--- a/frontend/middleware/adminOnly.ts
+++ b/frontend/middleware/adminOnly.ts
@@ -1,7 +1,15 @@
 export default defineNuxtRouteMiddleware(async () => {
-  const { roles } = useUser();
+  const { userIsAdmin, userIsSignedIn } = useUser();
+  const localePath = useLocalePath();
 
-  if (!roles.includes("admin")) {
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Thanks for this, @UnknownSean8! Appreciate the quick additions and doing this for events as well!

Issue #841, opened 2024-04-25
Add form validations to all create flows


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that's needed is for all create flows to have form validations such that submission is not possible without certain fields being filled. The following flows need validation:

- `organizations/create`
- `groups/create`
- 'resources/create'

amilliiee:
I've been doing some digging during work today and I might be able to get this done if you haven't resolved this in any way yet.

andrewtavis:
Would be happy to get the help, @amilliiee! Will assign you now, and let me know if there's anything I can do to help 😊

amilliiee:
I think right now the only question I have is how we want this organized. I'm assuming a single validation file?

andrewtavis:
Would be great, @amilliiee 😊 Thanks for checking in!

andrewtavis:
Hey @amilliiee 👋 Looping @to-sta into this discussion as he's working on the login for the backend :) Let us know if you're definitely going with formkit and what the plans are 😊

amilliiee:
@andrewtavis @to-sta I just figured since it can handle validation on it's own, it would cut down on what we'd need to do to get it working :)

I've been playing around at work and [this](https://codesandbox.io/p/devbox/sad-star-jk3qp8) devbox is a messy first attempt. Since I don't have everything I can't do a full mockup of the create groups page but the general idea is there. There's also a way to condense all of the styling but I need to look into it more.

The topic selection requires the pro version, so if we really do want to go this route I have no problem making the one time payment for it.

Let me know what you guys think, and I can have another go at validation without formkit when I get home today!

andrewtavis:
Hey @amilliiee 👋 Sorry for not checking this sooner :) Was gone for a week, and lots to do at work once I was back. Will check this over the weekend, and also feel free to drop into the dev sync tomorrow where we can discuss this as a group 😊

CC @momanyisamuel, this is the issue that's looking at form validation :)

andrewtavis:
Hey @amilliiee 👋 We're trying to check the sandbox within the dev sync now, but then we're getting something went wrong... Would you be able to open a PR with the changes? :)

andrewtavis:
[...]

PR #897, opened 2024-06-09
Form validation

amilliiee:
### Contributor checklist

---

### Description

FormKit validation PR to determine if we need to pivot.

### Related issue

- #841


to-sta:
@andrewtavis, can you look into the styling of the error messages. 

In case you want to change or add error message, these are part of the schema now. 

```js
export const organizationSchema = z.object({
    name: z.string().min(1, "Name must have at least 1 character").max(55, "Name must be less than 55 characters"),
    location: z.string().min(1).max(100),
    description: z.string().min(1).max(1000),
    tagline: z.string().max(100).optional(),
    social_accounts: z.array(z.string()).optional(),
    topics: z.array(z.string()).min(1).optional(),
  });
```

You can insert a custom error message for every constraint (e.g. `name`).

andrewtavis:
Is looking good, @to-sta! One thing for now is maybe we don't need to do a min topic as that won't be in the MVP?

to-sta:
> Is looking good, @to-sta! One thing for now is maybe we don't need to do a min topic as that won't be in the MVP?

Hmm gotcha, but this field is optional but if it is provided it has to have at least one element. Thought that does the job for the current MVP and beyond 😄

andrewtavis:
Makes sense 👍 Thanks for the explanation 🙏

andrewtavis:
Quick note on this via a conversation with @to-sta: We do want to make sure that the errors are apparent to the user - specifically being formatted differently and being red text. This is something that should be achievable via adding in a custom `<style>` block, or could be done in the FormKit config file (the latter is likely best practice, if possible). If it's not possible to do this in the config, then we could certainly write a custom CSS file that can just be imported into the `<style>` blocks :)

andrewtavis:
Thanks for your patience here, @amilliiee! Just now have gotten through the rest of the pressing PRs, and this one will be next :) It'll likely be sometime over the weekend before it comes in, but I'll get to it when I can 😊

andrewtavis (approved):
I'm assuming that the frontend error is related to the reset of the yarn.lock file, as prettier has been ran on the file. Really sorry about this review, @amilliiee... Have had a lot on my plate recently, and this + a decision to look into VeeValidate as something that styling should come a bit more easily to had this on the back burner for some time. Bringing in the changes with the FormKit work reverted, and I'll look into finalizing with the new process this week. The forms themselves need a lot of work, so I'll look to get as much of them switched over to Headless UI and other frameworks such that we can have as much accessibility and functionality coming from third parties as possible :)

Thank you! 😊

PR #900, opened 2024-06-12
direction arrow added

rabelmervin:
changes made


andrewtavis:
Thanks for this, @rabelmervin! I'll take a look later this week :)

rabelmervin:
Its great to work with you sir @andrewtavis  and Thanks for helping me. I would like to connect with you on socials.

frontend/components/media/MediaMap.vue
```diff
@@ -1,3 +1,5 @@
+
```
andrewtavis, line 1: Going through this now, @rabelmervin. Firstly we don't need the extra spaces here :)

frontend/components/media/MediaMap.vue
```diff
@@ -6,10 +8,15 @@
 </template>
 
 <script setup lang="ts">
+
+import { onMounted } from 'vue';
+import { useI18n } from "vue-i18n";
+import maplibregl from 'maplibre-gl';
+import direction-arrow from "frontend/public/icons/direction-arrow.png";
```
andrewtavis, line 15: This is breaking the header rules here, @rabelmervin. I'll see what I can do to fix it 🤔😊

andrewtavis (approved):
All's been fixed in the most recent commit :) Thanks for the initial work here, @rabelmervin! I decided to go with a black arrow instead of the one from the other library, as for the teal bike directions it was hard to see. Really great that we have this!

Issue #832, opened 2024-04-22
Map option to switch from walking to biking directions


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Now that work has been finished on #685, we're now ready to provide an option for the user to choose their routing options. The code to switch the profile is currently [commented out in the MediaMap component](https://github.com/activist-org/activist/blob/de5578e26f81c7593729f838879e4505da92d01c/frontend/components/media/MediaMap.vue#L140), with the base designs for this being found [on Figma](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=2673%3A20111&mode=design&t=0Y3ZLuQIW8ks96uu-1). For this first issue, it'd be great if we gave the user the option to switch between walking and biking direction services, with another issue then following that would then add the full functionality of switching between walking, biking and driving. The default would be walking, and if the user clicks the walking icon, then the api and directions profile would be switched to the biking directions.

- Note that we can add a MediaMap option to clear the directions via the work done in #791, with the code for this starting [here](https://github.com/activist-org/activist/blob/de5578e26f81c7593729f838879e4505da92d01c/frontend/components/media/MediaMap.vue#L191). The big challenge for this will be to add in a menu option that also has an icon that's also reactive to light and dark mode.

### Contribution

Happy to support with this or get to it myself eventually! 😊

andrewtavis:
CC @GrandSuccess87, would you have interest in working on this?

andrewtavis:
Thanks, @GrandSuccess87! Let me know if there's anything I can do to help :) And have fun working on the map! 🗺️😊

andrewtavis:
Further question that I have on this for maybe another issue, @GrandSuccess87: 

What about adding in a tooltip or something to the second waypoint that would show the time estimate for how long it'd take to get there? I'm realizing that the route is often times going to be the same for biking and walking, but having a time estimate would be a major distinction for it 😊

Beyond that, let's keep in mind that we'll likely want to clear there directions when the profile is changed to make sure that the user redraws the path and gets accurate directions with the new one. Would likely be ideal if we were able to reroute, but that's a separate issue. 

[...]

PR #901, opened 2024-06-14
#832/feature map option walking to biking

GrandSuccess87:
### Contributor checklist

---

### Description

I created a feature branch from main in my forked repo to develop and test locally.  

* Run `npm docker compose up --build`
* Navigate to the events page
* Add two wave points to the map
* You should see the path outlined for the "foot" route profile 
* The user can now toggle between "foot" and "bike" routing profiles and vice versa 
* Clicking the walking icon, should update the UI to display the bike icon (some CSS missing at the moment for the bike icon :p)
* Adding two wave points, now results in a path outlined for the "foot" route profile, however; so I can look into this some more or perhaps just create a new ticket for this. I'm thinking the latter (so this isn't delayed), which can also include CSS updates. Let me know which would be best!

### Related issue

- #832


andrewtavis (approved):
With the hydration changes merged in the event page is a bit broke, @GrandSuccess87, but this is working :) I updated the icons a bit to match the default ones a bit more, and specifically we needed to call `directions.destroy()` rather than `directions.clear()` within the directions profile switch, which then allows us to reassign the directions and lets the new profile take effect 😊 We'll be able to see for sure once we start calculating route durations, but the directions are teal and the image does switch 🙌 

This map is getting so cool! 🗺️

GrandSuccess87:
@andrewtavis That's wonderful news!! I'm so glad it's working 🎉🎉 It looks really nice too, thank you!

PR #913, opened 2024-06-30
fix creation of superuser

to-sta:
### Contributor checklist

---

### Description
Solves a problem with user creation. Email was not set due to wrong ordering of positional arguments.

### Related issue

- #ISSUE_NUMBER


andrewtavis (approved):
Thanks for this, @to-sta!

Issue #915, opened 2024-06-30
Adjust org API create method  


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We want to create the organization_text when we create a orginization. For this the create method in the OrginizationViewSet need to be adjusted. 
 

### Contribution

I will work on this myself :)

andrewtavis:
#916 was brought in, @to-sta, so you're good to work on this now :) Do you want to also add in the toast notification to the org creation if it's not successful?

andrewtavis:
Closed by #921 :) Thanks, @to-sta! 🚀

PR #921, opened 2024-07-01
org_text adjustments

to-sta:
### Contributor checklist

---

### Description
I adjusted the org API to create the OrganizationText for POST requests. 

ToDo:
* Toast needs to be added on the frontend, for bad requests.

### Related issue

- #915


andrewtavis:
Thanks for this, @to-sta!

Are you planning on adding Toast yourself, or should I check this? And let me know what changes should generally happen in the [organization store](https://github.com/activist-org/activist/blob/main/frontend/stores/organization.ts) :)

to-sta:
@andrewtavis I will give it a try, with [vue-sonner](https://github.com/xiaoluoboding/vue-sonner). 

We don't need the OrganizationText POST request anymore in the store. I can remove it myself while working on the toast 🤗

andrewtavis:
Sounds good! Thanks, @to-sta!

to-sta:
Removed OrganizationText request and added in the vue sonner toast in the bottom left.

andrewtavis:
Nice, @to-sta! Should I look into the conflicts from here? Let me know :)

to-sta:
> Nice, @to-sta! Should I look into the conflicts from here? Let me know :)

Yeah, that would be great. 

But I think instead of resolving 150 conflicts for versions, it would be more effective to re-compile the dependencies. You could remove the `yarn.lock` before merging it in and re-compile afterwards.

andrewtavis:
Ah ya, 100% @to-sta :) I don't deal with `yarn.lock` conflicts. Always just swap in the most recent version of things and recompile 😊

frontend/pages/organizations/create.vue
```diff
@@ -117,6 +118,8 @@
 
 <script setup lang="ts">
 import type { OrganizationCreateFormData } from "~/types/entities/organization";
+import {Toaster, toast } from 'vue-sonner'
+import BtnAction from "~/components/btn/action/BtnAction.vue";
```
andrewtavis, line 122: Quick note that we don't need to import the component in Nuxt, but then I think this was likely an auto import? :)

andrewtavis:
Quick note here, @to-sta: I'm not able to navigate to the organization about page post creation with the current state of things. I can take a look at this a bit later when I fix the `yarn.lock` file :)

frontend/pages/organizations/create.vue
to-sta: Yes, I didn't even realize that it was added to the imports 😄.

to-sta:
> Quick note here, @to-sta: I'm not able to navigate to the organization about page post creation with the current state of things. I can take a look at this a bit later when I fix the `yarn.lock` file :)

What was the response? I just checked if I removed the redirect, but it is still in their and it works fine on my end.  It could be the db volume since the models have changed.

andrewtavis:
I'll give it another shot on Sunday and let you know :) Thanks for the note that it's working on your end!

PR #928, opened 2024-07-06
added custom management command wait_for_db

to-sta:
### Contributor checklist

---

### Description
When we spin up docker compose and the database is just a slower than the backend container, we have to restart the container. We solved that with:

```yaml
restart: unless-stopped
```
the restart policy. But I think a custom command that catches the connection error and tries again is the cleaner solution. I would also recommand to keep the restart policy in place. 

To test this, I ran the container and checked the logs.

### Related issue

- No related issue

### DIFF
--- a/backend/backend/management/commands/wait_for_db.py
+++ b/backend/backend/management/commands/wait_for_db.py
@@ -1,0 +1,39 @@
+import sys
+import time
+from argparse import ArgumentParser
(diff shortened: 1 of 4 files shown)


andrewtavis (approved):
Thanks for thinking of this and implementing, @to-sta! Let's definitely present this for a sec at the next sync 😊

PR #929, opened 2024-07-06
Gh pytest action

to-sta:
### Contributor checklist

---

### Description
In this PR I have added pytest to the `pr_ci_backend` check. In order to get this up and running I removed the throttling test since it was failing. Throttling itself works, I checked it via SwaggerUI, it's just the tests. 

I think we should tackle this later on and only test it once instead of every app on its own. Its a general setting, no difference in different apps. 

I uncommented a factory test because it was not in sync with the model anymore. Factories should be updated everytime models are updated. I think a custom gh action would help here, that targets changes in models.py files. Since a change there means factories should be updated and the Figma schema.

I ran pytest successful, so at the moment there shouldn't be any errors. 

### Related issue

- #432


to-sta:
Also added a migrations check 😄

andrewtavis (approved):
Super great to have this in here now, @to-sta! Thanks for putting it all together! 🚀✅

PR #933, opened 2024-07-14
populating the database with dummy data

to-sta:
### Contributor checklist

---

### Description
We have been talking about populating the database with dummy data @andrewtavis @aasimsyed. I have added a custom management command to django `populate_db` with the option to specifiy the number of users via the `--users` flag. 

The file populate_db can be extend to load more dummy data (events, orgs).

### Related issue

- No related issue


andrewtavis (approved):
Sorry the review here took so long, @to-sta! Looking great on my end, and looking forward to the further fixtures coming in! 😊

PR #935, opened 2024-07-27
feat: addition of signal social media

UnknownSean8:
### Contributor checklist

---

### Description

Added Signal social to the modal share page. Thanks!

### Related issue

- #694


andrewtavis:
As discussed in the sync, let's look into how we would automatically launch Signal on a person's device and include a default message for sharing :)

UnknownSean8:
Hi, so far regarding the URL link for sharing to Signal, there is no feature to support sending a pre-defined message yet. However, it is possible to have a share-to-signal link. Thanks!

Ref:
https://github.com/signalapp/Signal-iOS/issues/706

andrewtavis:
Hey @UnknownSean8 👋 As discussed, would be great if we could still get Signal opened at least, and then we can figure this out later when that issue for Signal is closed 😊

andrewtavis:
Comment from the dev sync: Copy the URL to clipboard and open the app :)

UnknownSean8:
Hi @andrewtavis, the feature is working correctly for now. But there are still some settings/preferences that can be improved. Could you take a look and let me know your thoughts?

I think that the copyToClipboard function can be separated from the redirection so we can time the waiting time. But maybe the functions shouldn't exist in the file itself?

Other than that, since the link is using the same contentCopied variable, the copy URL button will also change when the Signal link is clicked. Do you want this behaviour or want it to be separated?

Thanks!

andrewtavis:
Hey @UnknownSean8 👋 So all's working well, but I do agree that we want to split the functionality of copying the link and the opening Signal. The copy link shouldn't change in this case, but the Signal option should change to a checkmark and the same green and `URL copied` should be the text displayed to the user until Signal opens. Can you make those quick changes and then I'll merge it in? :)

andrewtavis:
Also, don't worry about how some of the texts are a bit weird at the moment! This is coming from needing to combine a bunch of strings for the rework I talked to you about. It'll get figured out once all the PRs are in :)

Issue #918, opened 2024-06-30
Update to Django 5


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One thing that the community should get to sooner rather than later is update underlying frameworks early in the process so we don't need to think about this later when the platform grows. The first of these is Django, which hit version 5.0 at the end of 2023. This isn't needed for the first release, but should be gotten to by the end of 2024 :)

The upgrade guide is here: https://docs.djangoproject.com/en/5.0/howto/upgrade-version/

### Contribution

Happy to support with this! CC @to-sta. We'd discussed this a bit I believe, and wanted to put it in here so we can plan for it 😊

andrewtavis:
Closed via #937 😊 Thanks for getting to this so quickly, @to-sta!

PR #937, opened 2024-07-28
django upgrade to 5.0

to-sta:
### Contributor checklist

---

### Description
I have upgrade the django package from version 4 to 5. There are no breaking changes for us but there are some intresting new features:

* [Database computed default values](https://docs.djangoproject.com/en/5.0/releases/5.0/#database-computed-default-values)
* [Generated Model Field](https://docs.djangoproject.com/en/5.0/releases/5.0/#database-generated-model-field)

The could switch to `Database Computed Defaults` to improve the performance and consistency of our application. Would be especially use full for all time related defaults.

Of the top of my head, I don't see a use case for the `Generated Model Field` yet.

### Related issue

- #918


andrewtavis (approved):
Thanks for the upgrade here, @to-sta! Feel free to make any issues you can think of for how we can use `Database Computed Defaults` 😊 I'll update the docs with Django 5 now :)

Issue #870, opened 2024-05-03
Combine index menu option definitions


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The definitions for index menu options were copied in some cases, and it would be great if they could be combined. Specifically on the following two pages:

- [frontend/pages/organizations/[id]/index.vue](https://github.com/activist-org/activist/blob/main/frontend/pages/organizations/%5Bid%5D/index.vue)
- [frontend/pages/events/[id]/index.vue](https://github.com/activist-org/activist/blob/main/frontend/pages/organizations/[id]/index.vue)

we have `organizationButtons` and `eventButtons` respectively that should likely be using the logic in [useMenuEntriesState](https://github.com/activist-org/activist/blob/main/frontend/composables/useMenuEntriesState.ts).

### Contribution

Would be great if someone wanted to help with this! Happy to assist as needed or get to it myself eventually 😊

tasawar-hussain:
I would like to take it up!

andrewtavis:
Thanks for checking the issues and writing in so many, @tasawar-hussain! Assigning you here, and please let us know if there's something we can do to help!

andrewtavis:
Hey @tasawar-hussain 👋 Checking in with you here :) Is there something we can do to support?

tasawar-hussain:
Nothing right now, will start this week, and get back to you if there is something

andrewtavis:
Sounds great, @tasawar-hussain :) Looking forward to it!

t3azr:
Hello!! I'd like to try at this if that's okay : ). If I'm understanding/reading it right -- in both the pages, the organisation menu buttons are being manually created, and instead you want it to be done through useMenuEntriesState? Please feel free to correct me // add anything on !!

andrewtavis:
Yes, exactly that, @t3azr! Thanks so much for your willingness to work on this! :)

andrewtavis:
Give the contribution guide a Quick Look for the sections you'll need for this, and please let us know if there's anything you can do to help!

t3azr:
could i just ask , in the pages provided , i can only see organisedButtons , no eventButtons . did the original post mean to also change it to eventButtons in the event page or am i missing something ? :)

[...]

PR #940, opened 2024-08-01
Combine index menu defs

t3azr:
[x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

Use useMenuEntriesState to create the index menu options for events/id/index and organizations/id/index. Add idStr to prevent NaN from breaking urls.

Used useMenuEntriesState to create the menu options for both events and organizations. Changed the id used to create URL routes in useMenuEntriesState . 

When tested using an example organization, redirected to the correct URL asfaik. 

- #870 

Hope this is good!! I think the commit name may have become messed up when trying to commit if you need that fixed, thank you !!


andrewtavis:
Thanks so much for the PR, @t3azr! Would you want to give the [frontend warnings](https://github.com/activist-org/activist/actions/runs/10203261176/job/28288295441#step:6:1) a quick look? I can also support it needed! 😊

t3azr:
sorry for taking so long !! im pretty sure the only errors were because I forgot to use prettier to format the file , but if there's anything else wrong please contact me !! hopefully these changes have been updated and work !!

frontend/composables/useMenuEntriesState.ts
```diff
@@ -3,8 +3,11 @@ import { IconMap } from "~/types/icon-map";
 const createMenuEntry = (label: string, basePath: string, iconURL: string) => {
   const { locale } = useI18n();
   const router = useRouter();
+  // here, afaik, this is returning NaN as the id of a given organization is alphanumeric -- not a number.
+  // added the raw id which can be used with the routeURL , while maintaining the number version
+  const idStr = router.currentRoute.value.params.id;
   const id = Number(router.currentRoute.value.params.id);
```
andrewtavis, line 9: Ya this is a holdover from prior setups where the route URL was an integer. Nice catch! Let's just set id to your value - without the `Number()` :)

andrewtavis (approved):
All's working really well, @t3azr! Really appreciate you taking the time to make these changes and that you reacted to the tests as well! Sorry my review was a bit on the late side :) Getting back into things after needing to focus elsewhere 😊 Hope you pick up another issue soon!

Issue #923, opened 2024-07-04
Modals still open and can't be closed if we navigate away while they're open and then back


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Something that came up during the recent testing of the command palette was the realization that when we navigate away while they're open, then if we come back they can't be closed (thus breaking the entire UI). Steps to reproduce this are:

- Go to a page where a modal works, say `/home`
- Click on the activist logo to go to the landing page
- With the back button of the browser, navigate back to `/home`
- Open the command palette with command or control `k`
- Navigate back to the landing page by using the forward button (do not close the command palette)
- Navigate back to the `/home` page using the back button

After the above steps, the command palette will still be open, and further all functionality to close it will also not function. The `X` to close won't work, and clicking outside the modal will also not close it.

Would be great if we look into this!

andrewtavis:
Hey @mattburnett-repo 👋 Would this be something you'd be interested in looking into?

andrewtavis:
Let me know if more explanation on how to reproduce would be helpful! :)

mattburnett-repo:
@andrewtavis I'll get on this. Go ahead and assign it to me.

andrewtavis:
Hey @mattburnett-repo 👋 Thanks for jumping in to help here! My assumption would be that this is for all modals given they're all using the same store, but I didn't test others :)

mattburnett-repo:
The behavior described in the original issue is resolved. However, changes were made that will likely affect all other modals. This needs to be investigated and resolved.

We were using a prop called  `isOpen` to control / toggle the display of the modal.
Recently we started using a Pinia store called `modals` to control the modal display.

The behavior described in this issue arose from the fact that we were not changing the value of the `isOpen` prop when navigating away from the path underlying the displayed modal. That is now changed.

The very short version is 
- Took out all of the :isOpen props.
- Moved modal display flow away from :isOpen props, towards using the `modals` Pinia store.
- Made changes to `ModalBase` to watch the Route / Router object thing, and to close the modal if the use changes the route / path / url / etc. while the modal is still open.

[...]

PR #941, opened 2024-08-04
923 - Modal still open ...

mattburnett-repo:
### Contributor checklist

---

### Description

This PR addresses issue 923. The changes affected modal components that use an 'isOpen' prop; the 'isOpen' props and related functionality were removed.

Modals related to Organizations were affected. Other entities (ie Events) were not checked.

Testing was done following the original description of the problem in issue 923 (click activist icon, use forward / backward arrows to navigate, etc.). However, there appears to be an error with reloading the organization data when clicking the back button to check if modal is not displayed, so we can't be completely sure that everything is fixed.

That being said, the affected modals appear to work correctly without the 'isOpen' props, meaning that they appear to operate normally.

### Related issue

- #923


andrewtavis:
One question on this, @mattburnett-repo: What the purpose of all the `Watch for closeModal emit and do cleanup when it happens.` in the codebase now? Can we remove these somehow? And can we clean up the command palette code a bit to remove the TODOs from it? Maybe you could let me know what issues could be made from those TODOs, and then I can make the issues and we can remove the comments?

mattburnett-repo:
The 'Watch for closeModal emit...' are there just in case they are needed in the future. If you prefer, they can be removed.

The TODO messages in the command palette are notes-to-self. Later this evening I will go through them and let you know which of these are issue-relevant.

mattburnett-repo:
Here's a list of issues from the TODO's removed from ModalCommandPalette.vue

Search works, but it's picking up the literal values of 'displayName' (eg. "_global.home"). Search should instead look through the rendered values.

Sometimes router.push() loses the i18n part of the path (eg: /en/the.path.name)

andrewtavis (approved):
Thanks for this, @mattburnett-repo! Sorry for the wait in bringing this in... Let me know what your thoughts are on the next steps for modals :) Really is great that you're focusing in on them so much with them being such a large part of the UX 😊

Issue #938, opened 2024-07-28
Fixtures / Factories


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

We use [factory_boy](https://factoryboy.readthedocs.io/en/stable/) to create fixtures and test data. I've set up a custom management command to create random users: python manage.py populate_db --users=100.

Our plan is to extend this command to include organizations, events, and more.

TODOs:

 Add organizations
 Add events
 ...
As we add new data types, we should also validate and improve the corresponding factories.

You can find the factories in the ../factories.py files, which should accurately reflect the database models.

Note: This issue is dependent on PR #933.

### Contribution

I am happy to help :smile

andrewtavis:
Unblocked as #933 is now merged :)

andrewtavis:
@to-sta, maybe we can do this next so we can get back to a bit of what we had when we just had the integer routes and people could directly see the org pages, etc?

andrewtavis:
Closed by #947 🚀🚀🚀🚀🚀🚀🚀

PR #947, opened 2024-08-11
added orgs to populate db management command, improved org factory

to-sta:
### Contributor checklist

---

### Description
I have added the organizations to the populate db command. 

WIP: 
* org_text needs to be added to the OrganizationFactory
* Events should be added
* Group should be added

### Related issue

- #938

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -310,10 +310,20 @@ Please see the [activist style guide](https://github.com/activist-org/activist/b
 
 ### Backend
 
-Please run the following from the project root to test the backend:
+Please run the following commands from the project root to test the backend:
 
 ```bash
-pytest ./backend -vv
+# Start the Docker container:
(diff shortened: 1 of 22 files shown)


to-sta:
I removed the StatusType query within the OrganizationFactory and switched to factory_boy implementation. This we are removing the need of db access for fixtures.

andrewtavis:
I've reduced the number of entities/users/events created to 10 as the frontend build speed is faster, and everything's running in parallel now, so the frontend was actually loading faster than the backend with 100 (and even 50).

andrewtavis:
This still needs work to get the tests running, and the organization text table isn't returning a description – navigating to `/organizations` doesn't work as descriptions aren't coming through.

andrewtavis (approved):
Thanks so much @to-sta, @momanyisamuel and @wkyoshida!

Issue #718, opened 2024-02-24
Add layers options to events map


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Based on discussions in the most recent dev sync, one final element that should be added to the map for events is a menu option to control certain layers. We would like this to have three options:

- One to toggle public transit maps
- One to toggle biking routes
- One that's a radio option for how the directions should be routes (walking, biking or driving)
  - This is related to #685 

Note that this is blocked by #673 that is where we're learning as a community how to make a clickable map option that calls a function to clear the directions.

### Contribution

Happy to work on this myself or support someone with interest when this is unblocked!

andrewtavis:
As of now this issue would just focus on the bike route and transit map layers, and we could make another issue for switching the routing options once #685 is closed. I've just updated the [Figma designs for this](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=9492%3A10277&mode=design&t=iAcJvM6U2RZjd60X-1), which can further be seen below:

![Mode=Light, Mobile=False, Type=Map, Disabled=False](https://github.com/activist-org/activist/assets/24387426/118240e0-fc1d-46e9-b7d8-d8c4916871a9)

![Mode=Light, Mobile=False, Type=Map, Disabled=False (2)](https://github.com/activist-org/activist/assets/24387426/e01399dc-909f-4cec-85bc-e5df92c463ca)

In the second image above we have the biking option as being selected (not reflected in the map). The selected option would change color, as would the layers color to indicate that one layer has been added.

CC @nicholas-vecchio, in case you'd have interest :)

andrewtavis:
First step would be researching a way to add in the layers for the transit maps or biking maps onto the map, but hopefully there are extensions/plugins for this from OSM or another provider 😊

nicholas-vecchio:
Thanks for the mention @andrewtavis I'd be happy to look into this :)

andrewtavis:
Nice, @nicholas-vecchio! Really excited to hear what comes from this! Let me know if you need some support here, and happy to do a call to discuss at some point 😊

andrewtavis:
[...]

PR #949, opened 2024-08-19
feat(map): Add cycling raster layer from CyclOSM and a visibility toggle

wkyoshida:
### Contributor checklist

---

### Description

Discovered that the URLs available for [CyclOSM](https://www.cyclosm.org) are in the following format:
`https://{switch}.tile-cyclosm.openstreetmap.fr/[cyclosm|cyclosm-lite]/{zoom}/{x}/{y}.png`

- The `cyclosm-lite` version is nice in that it only contains the bike paths in the rendered layers.
- The full `cyclosm` version does highlight bike paths and other cycling infrastructure, but it does render a full "map" - so to speak.

With `cyclosm-lite` though, we can simply overlay it on top of the default layer that we currently use to include the bike paths on top of the existing map.

The property below (L210) configures the layer to not be initially visible, but toggling to the bike profile activates the layer visibility. 
```
  layout: {
      visibility: "none"
  },
```

### Related issue

- #718


andrewtavis:
Checking on the idea here, @wkyoshida: you're saying that we can just have the bike paths up when the user has bike routing selected? So then for the layers option it would just be to add in public transit routes, and we wouldn't provide an option to toggle the bike routes? (I think that this might be a good way of going about this, btw!)

wkyoshida:
> .. you're saying that we can just have the bike paths up when the user has bike routing selected? So then for the layers option it would just be to add in public transit routes, and we wouldn't provide an option to toggle the bike routes? ..

Yeee - thinking that we could really pretty much just add on the cycling layer and the public transit layer when their respective routing profiles are toggled on.

I'm looking into what our options are for the public transit one though... Transitland also seems like it requires an API key if I understood. Perhaps dealing with a key is what we end up having to do; so might simply be a matter of seeing which provider aligns better and whose free tier is generous :nerd_face: 

---

Quick notes on attribution in regards to 462d1ee: 
- Modified the references linked in the attributions
   - Link instead to https://www.openstreetmap.org/copyright as suggested in the [OSM Attribution Guidelines](https://osmfoundation.org/wiki/Licence/Attribution_Guidelines)
   - For cycling, link also to [CyclOSM](https://www.cyclosm.org) and [OpenStreetMap France](https://openstreetmap.fr)  
- Modified the attributions text to align a little closer to what OSM appears to use:

**BEFORE:**

- Standard

   image

- Cycling

   image

**Commit 462d1ee:**

- Standard

   image

- Cycling

   image

https://www.openstreetmap.org:

- Standard

   image

- Cycling

   image

frontend/components/media/MediaMap.vue
```diff
@@ -167,7 +175,18 @@ onMounted(() => {
...
                 attribution:
-                  '<a href="https://www.openstreetmap.org/about" target="_blank">Data &copy; OpenStreetMap contributors</a>',
+                  '<a href="https://www.openstreetmap.org/copyright" target="_blank">&copy; OpenStreetMap contributors</a>',
+              },
+              "cycle-raster-tiles": {
+                type: "raster",
+                tiles: [
+                  "https://a.tile-cyclosm.openstreetmap.fr/cyclosm-lite/{z}/{x}/{y}.png",
+                  "https://b.tile-cyclosm.openstreetmap.fr/cyclosm-lite/{z}/{x}/{y}.png",
+                  "https://c.tile-cyclosm.openstreetmap.fr/cyclosm-lite/{z}/{x}/{y}.png",
```
wkyoshida, line 185: Adding thoughts here..

It could potentially be good if we're able to load the tile source URLs from env vars - instead of configuring these hard-coded. Same as with `https://tile.openstreetmap.org` for the default layer source URL and the geocoding API URL `https://nominatim.openstreetmap.org`.

To note - as they are, the latter two are probably easy to do as we can simply replace them with an env var each. What about multiple-subdomain URLs though as in here for CyclOSM? We might have to configure a way to  be able to add more or less env vars if the number of subdomains change.

Or - we decide to not concern ourselves too much given that the mentioned services are likely to be and continue fairly stable in the near future? :thinking::shrug:

andrewtavis: Ya I'd say for now let's not concern ourselves too much with this, @wkyoshida :) We can revisit it later when we have a bit more capacity 😊

andrewtavis (approved):
Really really great, @wkyoshida! Appreciate that the map is continuing to be such a focus and that we're developing something so high quality 😊

PR #950, opened 2024-08-24
signout functionality enabled in mobile view

khushaal-nandwani:
The sign out button had to be added in mobile view in the side panel view. To do this, I basically mimicked what was happening in SideBarLeftFooter.vue i.e. the way it was working on bigger screens.

Not sure why `userIsSignedIn` was set to `false` previously. Wonder if something  will break by doing this?

Manually tested login logout in mobile view.

- #ISSUE_NUMBER: 821


andrewtavis:
Thanks so much for this, @khushaal-nandwani! I'll try to get to it in the coming days :)

andrewtavis (approved):
Was as easy as you found it to be, @khushaal-nandwani 😊 Thanks so much for your help! Sorry that the review took a little bit of time :)

Issue #871, opened 2024-05-03
Suppress warnings of "unmet peer dependencies" in frontend docker build


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

A general issue with the frontend Docker build that's come up is that we're consistently getting errors like the following when it's starting:

```bash
nuxt_frontend   | warning "nuxt > nitropack > @netlify/functions > @netlify/serverless-functions-api > @opentelemetry/otlp-transformer > @opentelemetry/core@1.23.0" has unmet peer dependency "@opentelemetry/api@>=1.0.0 <1.9.0".
```

The solution in the past has been to just install these dependencies, but if we keep on doing this then the `package.json` is going to get huge. It'd be great if we could look into options for running Nuxt within Docker that would allow us to ignore these warnings, and along with this we should then go through and find those dependencies that are not directly used in the frontend and remove them (assuming that they were installed to make these warnings go away).

Note - related issues are:
- #724
- #872

Happy to support on this or get to it myself! 😊

andrewtavis:
CC @aasimsyed :) With the most recent yarn updates from the damn `yarn install` we have even more, so I think your suggestion of ignoring these should be acted on 😊 Just pinging you so you have an overview of this!

andrewtavis:
Note that this should be worked on when we make all frontend dependencies explicit in #872 :)

andrewtavis:
Closed by #951 :) I agree that this should be a problem anymore, @arescrimson! We can readdress it if it comes back in a new issue. Appreciate the support and you also pointing out how this one can also be closed!

Issue #872, opened 2024-05-03
Make all frontend dependencies explicit version numbers rather than minimums


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The way that we've been doing frontend dependencies for the longest time is to have a minimum version number and let yarn figure out the rest. An issue with this is that it's making maintaining PRs for the frontend dramatically more difficult, as each time there's an added dependency we also get minor version updates that lead to overly complex merge conflicts with `yarn.lock`. With this in mind, it'd be great to have only explicit version numbers for the frontend in the [pageage.json](https://github.com/activist-org/activist/blob/main/frontend/package.json) file.

We can set all current minimums as the explicit version number and work it out until we have a working build, and then from there we can update them regularly as needed :)

Note - related issues are:
- #724
- #871

### Contribution

Happy to support or work on this myself eventually! 😊

andrewtavis:
CC @aasimsyed again :) Thanks for the chats on dependencies and for your work on that there `yarn.lock`! Let's try to have you not need to do that again :)

andrewtavis:
One thing that's needed from here is also removing all of the packages that we don't need except to suppress the warnings :)

tasawar-hussain:
I would like to take it up!

andrewtavis:
Hey @tasawar-hussain 👋 Thanks for the interest! We talked yesterday that @aasimsyed was going to take this one, but I'll get you on the other two you wrote in :)

andrewtavis:
@aasimsyed, can you write in here and I'll assign?

arescrimson:
Hey Andrew! Was this related to the issue that came up earlier this year when we resolved the bunch of issues that were coming up when building the frontend in Docker? If so, could maybe take a look or assist anyone who's currently working on the issue. Could also look at the related issue #871. Just let me know what your plans are for these issues, etc 😁

andrewtavis:
Hey @arescrimson! Happy to have you work on this one :) Big thing is just setting up the current dependencies to all be explicit, which might take a bit of trial and error, but I'm sure you'll be able to handle it 😊

Assigning, and great to hear from you!

arescrimson:
[...]

PR #951, opened 2024-08-26
Specify frontend package dependency versions 

arescrimson:
### Contributor checklist

---

### Description

This is for issue #872, where we are trying to specify frontend dependencies. This was done through just removing the carets and leaving a specific version number for each package. This has the effect of potentially resolving some issues where a released minor/major versions introduces new, unnecessary dependencies. 

**NOTE**: This is still a W.I.P - @andrewtavis I've left it as a W.I.P just in case you wanted to hop on the branch and look at the initial changes for yourself, see if this is on the right track, etc. 

**W.I.P List:** 

Resolving unmet peer dependency issues - I assume this is because some of the packages were using versions that don't match the ones that are now specified, so I'll resolve those. 

image
^ Looks like these warnings are mostly from one package, so hopefully straightforward fix. 

Testing to see if the application will still run the same previously with the new packages - this I'm open to suggestions/guidance for :) 

### Related issue

- #872


arescrimson:
Looks like checks are failing because of **@nuxt/eslint-config**, which is the same one that's also producing those warnings. Honestly, from what I understand about **eslint**, I think we could maybe leave that one unspecified. Since it's just a linter, minor changes/releases hopefully just still be fine? 

Either way, I'll start looking for the proper version that fulfills that missing peer dependencies - maybe that's also what's causing the checks to fail as well.

andrewtavis:
Thanks for this, @arescrimson! I'll be able to help here once a few more PRs are in. I'm finalizing the i18n ones tonight and then will start on others 😊

andrewtavis:
Hey @arescrimson 👋 For the tests to see if things are working properly, for now sadly it's just "is stuff loading" as we don't have proper tests. I'll click around to make sure that things are progressing, and then we can make use of the browser tests we're building up eventually. Ultimately there's nothing to test against... Do you want to fix the dependency errors and merge conflicts and then we'll bring this in? Happy to help as needed!

arescrimson:
Hey @andrewtavis just some notes on this - I had to leave **@nuxt-eslint** and **eslint** as loose dependencies since they were the ones generating the errors. This is also because I think if they were tied to one specific version, so many different packages rely on the different versions of **eslint** that they kind of have to be loose or we'll get a bunch of missing/incorrect peer deps. 

Let me know if it looks the same on your end, or if you want to change anything about this. So far, it looks like there are no warnings and that every package is working fine in their specified version, but just want to make sure 😄

andrewtavis:
This is great, @arescrimson! I'll check it out tomorrow and bring it in before doing the Dependabot Nuxt update from the other PR. Appreciate you looking into this! Makes sense that we still need some loose dependencies, and hopefully those won't cause us to need to update dependencies when we're doing frontend linting tests. I think likely we won't 😊

arescrimson:
Yep! I will say though @andrewtavis it was pretty straightforward to specify the versions, etc... but I think we all know how annoying the Node ecosystem can get sometimes when dealing with package dependencies, and I'm kind of thinking about how packages will interact going forward. For now, hopefully everything goes well, but in the future feel free to ping me if something starts breaking because that's something I'm a bit worried about 😅 

In addition, hopefully this also resolves issue #871 - unless any new packages are added for now, hopefully shouldn't have any unmet peer dependencies. I just tested it out with a fresh lockfile and it looks like there's still one error when fetching packages, but this should be how it looks for a person installing the project: 
image

andrewtavis (approved):
Thanks so much for the efforts here, @arescrimson! Really great to have this figured out :) I'll play around a bit with the "trying to install in the same destination" error a bit so hopefully we can get a clean build 😊

PR #955, opened 2024-08-31
feat: copy url and redirect to signal when share link is clicked

UnknownSean8:
### Contributor checklist

---

### Description
This is the same PR from the other signal additional PR. I have a problem with my git so I had to recreate the branch.

For the implementation, I feel like there is still improvement that can be made, especially with the duplicated copy URL code. Feel free to let me know if there are any ideas/comments.

Also, just an idea, maybe we could add in a "Redirecting to Signal" infobox or something to let the user know that it is redirecting to signal.

### Related issue

- #694

### DIFF
--- a/frontend/components/modal/ModalSharePage.vue
+++ b/frontend/components/modal/ModalSharePage.vue
@@ -94,10 +94,59 @@
               iconSize="1.5em"
             />
           </s-facebook>
-          <!-- <div class="flex h-full w-full cursor-pointer items-center gap-3">
-            <Icon :name="IconMap.SIGNAL" size="1.5em" />
-            <p>{{ $t("components.modal_share_page.signal") }}</p>
-          </div> -->
+          <div
+            @click="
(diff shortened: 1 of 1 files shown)


UnknownSean8:
@andrewtavis Hi, I've fixed the comments that you listed yesterday. Do let me know if there are any changes needed. Thanks!

andrewtavis:
Hey @UnknownSean8! The idea of maybe a toast to that says that Signal is going to open is a good one 😊 Let's maybe not include it in here, but have a separate issue that implements this for all the share functionality that needs it? Would you want to work on that? I could make the issue :)

UnknownSean8:
Sounds like a good idea! I would be happy to work on it. Thanks!

andrewtavis (approved):
Really really nice, @UnknownSean8! Let's maybe change the `Copied` to `URL copied`, but I'll do that in the keys :) Will make the issue for the toasts now!

Issue #902, opened 2024-06-15
Standardize i18n keys


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Something that came up recently is that the i18n keys are not standardized - i.e. sometimes we have kebab case or snake case. This issue is to check the keys and change them all to either snake case or camel case, with the reason for this being that we want to be able to easily replace full sections of the key.

jannesblobel:
@andrewtavis  I came upon your issue by chance. 

[What we are currently discussing internally in our team might also be of interest to you. Simply use human readable and random ids and let the tooling generate them. This will solve many problems 

https://github.com/opral/monorepo/issues/1892

jennethydyrova:
hi @andrewtavis! I would like to help with this issue, is it still available?

andrewtavis:
Hey @jennethydyrova 👋 If you wanted to send along a PR that converts all of the kebab case to snake case, it'd be much appreciated!

andrewtavis:
Let me know if there's anything I can do to help :) I'll also send along a commit to this that will make a check to see if the keys can be effectively combined. We did discuss this a bit internally, and feel that the current system is valid as it allows the localization team to easily filter for parts of the app 😊

jennethydyrova:
I just wondered if would it be fine if I send all changes in one PR or would it be easier for you to review if I split it to multiple PRs?

jennethydyrova:
I've got one more question. According to style guide folder names should be separated by dash

> Localization keys should be defined based on their component or page within the platform and the content that they refer to (CONTENT_REFERENCE below). Please use the following rules as a guide if you find yourself needing to create new localization keys:
Separate directories and references by . and CamelCase file name words by - in keys
Ex: "components.search-bar.CONTENT_REFERENCE" for the SearchBar component

then, for example,  `components.btn-action.save-settings` key should be renamed to `components.btn-action.save_settings`, is that right?

andrewtavis:
Glad to have you working on this, @jennethydyrova! Looks like the styleguide needs to be updated as well. Let's use underscores for the whole thing. That was a poor choice by us earlier :)

Would you be able to send along an update to the styleguide with this as well?

[...]

PR #966, opened 2024-09-03
Rename and add missing key to the keys list

jennethydyrova:
### Contributor checklist

---

### Description

1. Renamed `components.meta-tag-video.view-vide` to `components.card_search_result.view_video` to match new naming convention.
2. Added renamed key to the list of keys as it was missing.

### Related issue

- #902


jennethydyrova:
The pipeline is failing because of some invalid i18n keys, I can fix them in this PR if that's okay, @andrewtavis?

andrewtavis:
Sure thing, makes sense, @jennethydyrova! :) Please do fix them as directed by the errors. Means that the refactoring we recently did worked out :)

andrewtavis (approved):
Really appreciate you keeping up with the localization upkeep, @jennethydyrova! Would be interested in getting your feedback on the progress. Let me know if you think some things should be changed or if we can improve the process somehow 😊

Issue #968, opened 2024-09-08
map: refactor to use localized strings within map controls/interactions


wkyoshida:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Like elsewhere in the application, it would be good for the map controls/interactions to also make use of localized strings.

### Contribution

Happy to support someone on this or get to it myself 😁

akdave2003:
Hi this is Ansh, Zach and I are ready to be assigned!

andrewtavis:
Nice! Thanks for writing in, @akdave2003 :) Let us know if there's anything we can do to support!

akdave2003:
Hey I had a question. So in the file I started out by refactoring these strings: FOOT, BIKE, DRIVING, CAR. I also added the corresponding variables to the json file. My partner and I were wondering, would this apply to every string in the file or are there some strings we should leave alone. For example, there are two strings called api and profile, do those need to be refactored?

andrewtavis:
Those would not need to be refactored, @akdave2003 :) The big thing is that we need `Change profile` also localized, as well as `hosted by` :) Those look to be the main ones that are needed 😊

andrewtavis:
The API request is coming out of the component anyway, and it will be replaced with loading in the event object and then gettin the bounding box, latitude and longitude from it :)

andrewtavis:
Hey @akdave2003 👋 I just did some major changes to how events work, so I'd suggest pulling them in. You can now navigate to an event page and see the results of your localized controls via `View events` on the development home screen and then selecting an event 😊

Please let us know if there's anything that we can do to support!

andrewtavis:
Closing this issue because of the work in #1045, with further work to localize the preloaded Maplibre GL GS controls happening in #1071 :)

Thanks for the great contribution, @akdave2003! Hope we can work with you again soon :)

Issue #961, opened 2024-09-02
map: attribution not fully responsive


wkyoshida:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Noticed that the map attribution is not quite fully responsive. The following main points were observed:

---

- In dark mode, parts of the attribution are unreadable due to the text color:

![image](https://github.com/user-attachments/assets/b0346c72-cc36-4c67-b5f9-0ce8b99802c7)

..which was unfortunately overlooked, as that wasn't the case in local development and in the preview for PR #949:

![image](https://github.com/user-attachments/assets/2c8efc38-9c20-4110-bae4-99a36fdbc907)

---

- Depending on screen size, the attribution can also sometimes sit on top of the `Clear directions [x]` custom control that we have on the bottom-left corner of the map:

![image](https://github.com/user-attachments/assets/a622df03-5077-494f-a231-3ce15550d2c6)

..on this point, something that I'm considering is actually moving the `Clear directions [x]` custom control. Perhaps it could make sense to place it right under the route profile toggle button? The protruding size of the custom control might also not be an issue if a button to hide it is also implemented (as proposed in #960 - in the same vein as how the attribution control works).

---

- Finally, not quite a fan of how the attribution wraps to fit in the text honestly... As we see in the below screenshot:

![image](https://github.com/user-attachments/assets/494fe025-9e15-47b2-8eee-1a124fa42780)
..only the word `France` is pushed down to the second line to fit everything. It could be more interesting if that entire segment of the attribution had been pushed down, e.g. in this case, `CyclOSM tiles hosted by OpenStreetMap France` so that the attribution might have looked something like: 
```
MapLibre | © OpenStreetMap contributors |
CyclOSM tiles hosted by OpenStreetMap France
```
This point though is perhaps a little more tricky to wholy address. It also could simply be me being a bit too nitpicky :sweat_smile: so fully open to reconsideration. Mentioning this as I'm actually fine if this third point is set aside.

andrewtavis:
[...]

PR #969, opened 2024-09-08
refac(map): Update attribution text to condensed version and change profile on key press [p]

wkyoshida:
### Contributor checklist

---

### Description

Slowly but surely chipping away at some map stuff..
- Was confused by the `keydown` event on `[x]` causing both the clearing of directions and toggling the route profile. Modified so that the route profile is changed instead on `keydown` events on `[p]`. Let me know if there was a reason though for having both actions caused by the same event - wasn't sure
- As per part of the discussion in #961, modified the attribution texts to be more condensed. Went with:
   - `© OpenStreetMap` as that is also a fairly common alternative ([ref](https://osmfoundation.org/wiki/Licence/Attribution_Guidelines#Attribution_text)).
   - `CyclOSM hosted by OSM France` as opposed to `CyclOSM tiles by OSM France` from https://github.com/activist-org/activist/issues/961#issuecomment-2323958279 , given that I'm actually unsure if the CyclOSM styles are officially an OSM France project. However, the tiles themselves are definitely hosted by their infrastructure - so that seemed more appropriate. 

Also - as I was looking at this, thought about how eventually using localized strings for the map controls/interactions would be a good idea actually - created #968 for it.

### Related issue

- Part of #961

### DIFF
--- a/frontend/components/media/MediaMap.vue
+++ b/frontend/components/media/MediaMap.vue
@@ -90,6 +90,7 @@ const routeProfileMap: RouteProfile[] = [
 
 const walkingRouteProfileControl = `
   <div
+    title="Change profile [p]"
     id=${routeProfileOptions.FOOT}
     style="
     background-image: url(${walkDirectionsIcon});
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Working really well, @wkyoshida, and thanks for adding in the hotkey for the profiles! I think that the route profile update on `x` likely came from @GrandSuccess87 and I trying things back when we were first getting it to work.

Coming along so nicely! 🗺️

Issue #964, opened 2024-09-03
map: on event creation, allow location search and store location geometry


wkyoshida:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Related to issues:
- #621
- #628

I believe I might've mentioned this idea before, but noting it down here finally..

---

When creating an event, it could be interesting for the user to be able to search for locations _à la_  nominatim and for the creation page to then display a preview map for the user to check if a selected location is correct for their event.
- For this, once a user searches for a location and selects it, the location's geometry that is returned in the response could then be used to pan to the location in the preview map.

With the event details and location confirmed, activist could then store the location's geometry to later use it to display on the event page's map. 
- For this, the idea could be for the location table (of issue 621) to be the one to store said geometry - by leveraging the [PostGIS](https://postgis.net) extension for Postgres.
- With the geometry table, we have some flexibility in how we choose to store the location:
   - We could simply store the centroid coordinates for the location. This would be the easiest and cheapest.
   - We could store the bounding box (bbox) of the location. Imagine that this is essentially a box that surrounds the location. A bbox is helpful, since it's a cheap way to also still very roughly know how large/small a location is. This is useful for loading the map on the correct zoom level for that location. (My vote would be for this option).
   - We could store the full geometry for the location also if needed. For Berlin, for instance, imagine this is the multi-vertices polygon that represents the city limits/boundaries.

With the event page, we would then have a way to more appropriately set the zoom level and the map center that fits for the event's location. 

### Contribution

Happy to support someone on this or get to it myself 😁

wkyoshida:
Noting that:
- This could replace the current `offline_location_lat` and `offline_location_long` fields in
[Event](https://github.com/activist-org/activist/blob/f4608bbe35aaa93102a031a4928c2500cf9ddef7/backend/events/models.py#L29-L30)
- This could replace the current implementation in the map using a [nominatim request](https://github.com/activist-org/activist/blob/f4608bbe35aaa93102a031a4928c2500cf9ddef7/frontend/components/media/MediaMap.vue#L158-L164) to attempt to determine the location coordinates to display based on the location's name

andrewtavis:
[...]

PR #970, opened 2024-09-09
refac(map): Update `maxZoom` at map-level and map fit on location bbox

wkyoshida:
### Contributor checklist

---

### Description

- Configured `maxZoom` at the map-level rather than per-layer:
   - Different per-layer configurations resulted in the CyclOSM layer hiding while the default layer remained visible - which is the expected behavior definitely, but may be confusing if the user specifically toggles to see the cycling profile. A configured `maxZoom` on the entire map avoids this.
   -  Went with `maxZoom: 19` as that is the lowest max zoom available as per [docs](https://wiki.openstreetmap.org/wiki/Featured_tile_layers):
      - Default layer is `z19`
      - CyclOSM layer is `z20`

Showcasing some of what I was thinking with #964 ... 
- Modified to fit the map on the bbox of the event location:
   -  This frees the map from relying on a generic zoom level for all events and fits on a zoom more appropriate for a given location.
   - Configured some settings, but these can be modified definitely if alternatives are preferred:
      - `duration: 6000` sets the duration of the map fitting/zooming animation to 6 seconds
      - `padding: 100` fits the map with an additional 100 pixels of padding around the bbox - useful so that the fit is not too zoomed-in
   - Note FYI - The bbox coordinates passed into `fitBounds()` may appear funky, but that is just due to the different bbox formats used by Nominatim and MapLibre
      - [Nominatim](https://wiki.openstreetmap.org/wiki/Bounding_box): `south Latitude, north Latitude, west Longitude, east Longitude`
      - [MapLibre](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/LngLatBoundsLike/): `west Longitude, south Latitude, east Longitude, north Latitude`

### Related issue

- Kinda related to #964 ?


andrewtavis:
Points on this, @wkyoshida :)

- Definitely agree that this is a better approach so that the zoom can be more appropriate to the location
- Do we have to have the zoom, or is it possible to just have a one 1-2 second one from not too far away into the map location? I think there might even be accessibility things for users who don't like things moving so much with a full globe zoom?
- Is it possible to have a 20 zoom level, as as of now the user wouldn't be able to necessarily orient the location to other surrounding to know where it is without zooming out? Or maybe we add a bit to the padding?

Thanks so much for continuing to crack away at all this!

andrewtavis (approved):
Apparently we needed to rerun this on Netrlify post clearing out the cache :) It does work as seen in https://app.netlify.com/sites/activist-org/deploys/67467bf1bae46d4f9da015b9. I've also opened #1019 that has the same changes on here and it did finish, so we're good to go here 😊

Thanks, @wkyoshida! Wanted to bring this is in so it can be factored into the locations table work I'm starting on. Hope the removal of the zoom is ok for now until we can figure out how to zoom from not so far out 🗺️🔍

PR #977, opened 2024-09-21
#938 Expand fixtures to produce human relatable results

andrewtavis:
### Contributor checklist

---

### Description

Various fixes to the fixtures :)

### Related issue

- #938


andrewtavis:
Note: Before this is merged, I will add in directions for onboarding new users. Navigate to XYC in order to see organizations, etc.

backend/backend/management/commands/populate_db.py
```diff
@@ -18,99 +20,85 @@
...
 
     def handle(self, *args: str, **options: Unpack[Options]) -> None:
-        n_users = options.get("users")
-        n_orgs_per_user = options.get("opu")
-        n_groups_per_org = options.get("gpo")
-        n_events_per_org = options.get("epo")
+        num_users = options.get("users")
+        num_orgs_per_user = options.get("orgs_per_user")
+        num_groups_per_org = options.get("groups_per_org")
+        num_events_per_org = options.get("events_per_org")
```
wkyoshida, line 41: The errors with the `pr_ci_backend` workflow at least, appear to be due to the above 4 possibly being either an `int` or `None` (despite the defaults that are specified in the `parser` arguments on lines L32-L35).

Would doing for instance `options["users"]` as opposed to `options.get("users")` still work? Or at least specifying the default here as in `options.get("users", 10)` so that the `.get()` calls don't possibly return `None` by default.

wkyoshida (commented):
:rocket:

backend/backend/management/commands/populate_db.py
to-sta: The get method return type is `int | None`. Because the method will return `None` instead of throwing a KeyError, if the key does not exist. We can just use square bracket notation `options["users"]` and there errors are gone 😃.

andrewtavis:
I still don't know what's up with the frontend error, but is what it is for now. I'm hoping that it goes away when we update the frontend dependencies, which is needed at this point.

df04381 Adds in a `devMode` Pinia store, with `devMode.active` being used to see if `loclhost:3000` is in the route. This is the same technique used for disabling the captcha in dev mode, but now universal. With this I've gone in and changed the landing page in dev mode to replace the `Request access` button with `View organizations` and `View events`. I also figured out the line clamp for search result cards.

The last thing I want to do here is play around with the `events` page a bit to make sure that we're loading in events for that, and that clicking on one gets to a fully populated page from the backend. I won't finish all the hydration, but I think that it's important to make it easier for people who are working on events 🗺️

andrewtavis:
Frontend error's still a problem, but I'm going to merge. Tons of changes included in here at this point 😊 New ones are that the events are also being listed, and the work for the backend to do this has also been gotten to. Likely are some minor things that need to get taken care of. Please let me know if you see something! Also feel free to send along a PR with any needed changes :)

Issue #835, opened 2024-04-22
Click on map full screen triggers minimization of mobile browser tab bar


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The [MediaMap](https://github.com/activist-org/activist/blob/main/frontend/components/media/MediaMap.vue) component can be seen on [https://activist.org/events/ANY_INTEGER/about](https://activist.org/en/events/123/about). One thing that would be a nice feature would be to make sure that the browser tab bar Safari, Chrome and other browsers would retract/minimize so that the map is the full screen and we don't have the browser search bar as well. An example with the search bar still visible at the bottom of the screen is:

![IMG_0172](https://github.com/activist-org/activist/assets/24387426/cab9b76b-77bc-4813-95b0-b31a0a10e961)

The documentation for the FullscreenControl class can be seen [here](https://maplibre.org/maplibre-gl-js/docs/API/classes/FullscreenControl/). By the looks of it there's a `fullscreenstart` event that could be used to potentially trigger the search/tab bar being minimized. 

### Contribution

Would be happy to support on this or get to it myself later 😊 This would be a great feature to add into the already strong map component, as the user experience is a bit reduced when the browser controls are visible on full screen.

Geet-S:
hey, still need help with this?
From what I've checked on PC and android with Chrome browser this doesn't seem to be a problem but it could be different on Safari

andrewtavis:
Would be great to get some help with this, @Geet-S! Let us know what your thoughts are, or also feel free to send along a PR :) Happy to help as needed 😊

andrewtavis:
Hey hey @Geet-S 👋 Checking in to see if there's anything that we can do to assist here :)

Geet-S:
Hey sorry, I weirdly never got notified on your reply to me I don't know what's up with that.

So, I don't really have any apple products to check on safari if it's not in full screen like you described but let me know if there is a way for me to test that out through some emulator of sorts, I'm kinda new to this :P

andrewtavis:
Hey @Geet-S 👋 No stress at all. You'd be welcome to send along something and I could test it on my end :) Seems like the following pages might have suggestions we could try:
[...]

PR #979, opened 2024-09-26
#835 Close Safari buttom bar for fullscreen map

andrewtavis:
### Contributor checklist

---

### Description

Sending along a PR so this can be checked on Mobile via Netlify previews. Goal is that when the user clicks fullscreen on the map that the bottom bar on Safari will automatically close. Without this, on smaller iPhones the map can be a bit small in the fullscreen mode.

### Related issue

- #835


andrewtavis:
Closing this after trying a ton of options. Doesn't look like this is possible.

PR #981, opened 2024-10-01
Bug: 980 (Cannot read properties of undefined (reading 'isOpen') in MediaImageCarousel.vue)

mattburnett-repo:
Added if statement to watch(route) so that closeModal() is called only when modals are present.

### Contributor checklist

---

### Description

This PR attempts to correct an error that occurs when a user navigates the app and there are no modals in the modal Pinia store. It is part of a larger effort to make the modal implementation more consistent.

### Related issue

- #ISSUE_NUMBER 980

### DIFF
--- a/frontend/components/modal/ModalBase.vue
+++ b/frontend/components/modal/ModalBase.vue
@@ -96,8 +96,11 @@ const closeModal = () => {
   modals.closeModal(modalName);
 };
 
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Thanks for the quick fix here, @mattburnett-repo! Let's be in touch on a call later to look into this :)

Issue #987, opened 2024-10-15
BUG: Language translation issue in Portugese


Sushil010:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

When clicking on Portugese language from dropdown the main page language should change to portugese but it is still in English.

![image](https://github.com/user-attachments/assets/4cb413b2-2a67-4763-8db5-b4861352cfab)

The language on main page after clicking on "PT" dropdown should be in portugese but is still in English language. But it works for other language dropdown options.

andrewtavis:
Hey @Sushil010 👋 This isn't so much of a bug in so far as we just don't have Portuguese localizations right now. Might make sense to remove the option at this point. Would you want to do that and then we can start including languages in the localizations only when we have the localizations?

Sushil010:
I am fine with that and there seems to issue with spanish translation also. Only some heading and some part has been translated. Is the reason with Spanish translation also same?

andrewtavis:
Spanish is only partially translated. We can maybe remove that one as well for now :)

andrewtavis:
Closed by #988 :) Thanks for the issue and the quick fix, @Sushil010!

PR #988, opened 2024-10-16
fix: removed the languages from dropdown whose localization didn't exist.

Sushil010:

This request is made to remove the languages whose localizations was not available for translation. It can be seen by moving at main page and by clicking on the dropdown button in language

-#987
Below is the image after removing the issue:
image


andrewtavis:
Thanks for this, @Sushil010!

andrewtavis:
I do the final review soon :)

andrewtavis (approved):
Sorry the review for something so easy is coming so late, @Sushil010 :) I had a lot of unexpected work for a coding outreach program over the last few weeks. Thanks so much for the contribution here! Spanish is already at 25%, but we'll add it back in once all the strings in the current user facing interface are done 😊

PR #997, opened 2024-10-27
Page/Component Object refactor; Added new tests for Landing/Home/Sign-in pages

aasimsyed:
### Contributor checklist

---

### Description

- updated Playwright
- ignored tests/results/reports from watch
- added an html reporter for axe
- updated components and pages with id locators for use in tests
- refactored the page object model design, updated page/component objects, tests for landing, home, and sign-in pages
- added a utility method to run accessibility tests on any given page

- tested locally

### Related issue

- #ISSUE_NUMBER
N/A


andrewtavis:
Lemme know when this is ready for a final review, @aasimsyed :) Would be good to do a call this week as well, if it can work on your end 😊

andrewtavis:
One note on this is that the `yarn.lock` file is the current version from main as I didn't want to deal with the merge conflicts.

frontend/tests/utils/PageObjectBase.ts
```diff
@@ -0,0 +1,66 @@
...
+  ): Promise<void> {
+    const timeout = options?.timeout || 10000;
+    await this.page.waitForURL(expectedUrlPattern, { timeout });
+  }
+
+  public async currentTheme(): Promise<string> {
+    return (await this.page.locator("html").getAttribute("class")) ?? "";
+  }
+
+  public getLocator(selector: keyof typeof this.locators): Locator {
```
cquinn540, line 61: Is there a reason we don't want to use locators the way Playwright recommends?  https://playwright.dev/docs/locators

This implementation discourages that, and also encourages us to use ids to locate most of our elements as opposed to using roles and labels the way a user would.

cquinn540:
@andrewtavis I've been going through the tests now, and I'm starting to think we should just stick with ids for this change.  Some of the components aren't accessible enough yet to query them by role, and fixing that would expand the scope of this change by quite a bit.

I think for now it makes sense to get the tests passing and maybe rename the `tests` directory to `test-e2e` just to make it clearer what it is.

Then I can make another issue to improve one test and the accessibility of the components under test and use that as the basis for creating an e2e testing guide.

After that we can keep improving the tests and accessibility one test at a time.

andrewtavis:
All of the above sounds perfect, @cquinn540 :) Thanks for the suggestion! Will check in with you on a call this weekend to review the changes related to this PR  ✅

andrewtavis:
Closing this as #1041 brings in the changes here and adds further edits to the e2e tests :) Thanks for the great work here, @aasimsyed! Hope all's well 😊

Issue #747, opened 2024-03-07
Check and load in env variables for backend/db only builds


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

An issue that a contributor faced recently was that for builds that are not the full Docker build, but rather only the backend and database separately, that the environment variables within [.env.dev](https://github.com/activist-org/activist/blob/main/.env.dev) were not loaded in. Ideally we would check what the build situation is, and then in [settings.py](https://github.com/activist-org/activist/blob/main/backend/backend/settings.py) for the backend we would pass an argument to `dotenv.load_dotenv()` to load them in. Specifically, the line should be changed to:

```py
dotenv.load_dotenv(override=True, dotenv_path=Path(__file__).resolve().parent.parent.parent / ".env.dev")
```

This again should be triggered via an if clause that determines the Docker state and whether it would be necessary :)

### Contribution

Happy to look into this or support as needed!

to-sta:
@andrewtavis I think best practice in the django world is to split the settings file into:

* base.py (shared settings amongs all enviroments)
* local.py (this would include the line above)
* development.py (development specific settings)
* production.py 

Then include the line above inside the local.py settings file. Determine the appropriate settings file based on an environment variable (e.g., mode="Production"). If the mode is neither set to `Production` nor `Development`, default to the `Local` settings.

bharath637462:
@to-sta @andrewtavis can i work on this issue?

to-sta:
Sure, @bharath637462. Please name the env. variable `DJANGO_ENV`.

bharath637462:
@to-sta 
when i run the docker it does not read the env variables.
if i add DJANGO_ENV in env and 

settings.py
if DJANGO_ENV:
dotenv.load_dotenv(override=True, dotenv_path=Path(__file__).resolve().parent.parent.parent / ".env.dev")

how will it work? i Think it cant read DJANGO_ENV


i have put statically the var true and false

docker_build_condition = False  # Example: Set to True for demonstration

django_stubs_ext.monkeypatch(extra_classes=(viewsets.ModelViewSet,))
if docker_build_condition:
[...]

PR #1005, opened 2024-11-07
Switch from container to local with env variable

cquinn540:
### Contributor checklist

- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#testing)

---

### Description

I made a bit of a design decision here.  I wanted to make this user friendly, so now if you set the env variable `DJANGO_ENV=LOCAL_DEV` you can run the backend locally, but you can also spin up the full container deployment without having to edit the env files.

This worked seamlessly on my mac with django running locally talking to the postgres container.  Unfortunately I don't have machines with other OSs so I can't verify that it will work as seamlessly in other environments, but you should be able to further customize `.env.dev.local` if you need to as any variables in there will overwrite the ones set in `.env.dev`.

I tried running both the UI and the backend locally and the file watchers opened more files than my system can handle, but I confirmed that the UI container will work with both versions of the backend.

Let me know if you think this approach makes sense or if you want more detail in the docs!

### Related issue

- #747


cquinn540:
Looking at the errors it looks like stricter type checking is flagging some unrelated files?

andrewtavis:
Could be, @cquinn540. In that case I can take a look :)

andrewtavis:
But if you have suggestions let me know 😊 Great to have this PR in!

cquinn540:
Figured out most of the type errors.  One I think is best to ignore because I don't think there is any easy way to track down the needed generic type.

I'm not sure how to run the frontend tests locally

andrewtavis:
You can check the [testing section of the contributing docs](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#testing), which should have everything you need. Let us know if it's a bit incomplete though!

andrewtavis (approved):
Really appreciate the consideration of the developer experience here, @cquinn540! Thanks so much 😊

Issue #903, opened 2024-06-15
Experiment with and apply global middleware file rather than file based


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would investigate and implement a file that would serve as the global middleware for the full platform. So rather than having a file based system where we put the middleware in the route that needs it, we'd instead have one file that then defines all the routes that get a specific requirement in order to access them.

### Contribution

@UnknownSean8 said he had interest in this one :) Can you write in here and I'll assign? Very happy to support!

Robot8lover:
Two approaches I've found (using Nuxt's middleware guide) are using global middleware (so run the middleware on every page, check if the page matches some criteria, and do something based on that) and setting middleware at build time (use the pages:extend lifecycle hook as part of the nuxt.config.ts file to go through the pages and add middleware based on criteria). The former has the advantage of potentially being more readable since it places all the middleware related files in the middleware directory and seems to be more what you were going for based on the description. The latter has the advantage of not running every time on every page and is more similar to the current system except that all of the middleware additions can be in one file.

andrewtavis:
I'd say that performance should trump readability for this, @Robot8lover, so let's go for the second option :) Can we have one separate file that's loaded into the `nuxt.config.ts` so it's at least separated out from that larger file?

andrewtavis:
Happy to assign if you'd like to work on this!

Robot8lover:
Sure, I'd love to work on this.

Robot8lover:
And yes, we can just import into the `nuxt.config.ts` so all that can be separated.

andrewtavis:
Thanks for the confirmation and picking this up, @Robot8lover!

Robot8lover:
Right now, the only pages that use middleware are the create pages, which check if the user is signed in. Are there any other pages that should get middleware at this point?

andrewtavis:
For right now those are fine, and then we can expand it with individual issues moving forward :)

Robot8lover:
My apologies for getting back to this so late, but I think this PR should accomplish it.

andrewtavis:
No stress at all, @Robot8lover! Great to hear from you :) We'll check it this weekend as we get to things 😊

[...]

PR #1017, opened 2024-11-22
Build hook middleware

Robot8lover:
### Contributor checklist

- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#testing)

---

### Description

Use global middleware instead of file based middleware. Rather than assigning middleware in the file for a specific route, we can assign the middleware in one file. This still uses named middleware from the frontend/middleware directory.

Note that as this uses a build hook, I believe you will need to restart the environment/frontend to reflect changes in the middleware (at least the part assigning middleware to pages since that is what is done from the hook), which can take some time.

I tested on various systems and browsers to ensure that the admin/user protection works as before.

### Related issue

- #903


andrewtavis (approved):
Thanks for the great work here, @Robot8lover! Always a pleasure 😊 Please let us know if another issue would be of interest, or I'd be happy to suggest one as well :)

Issue #550, opened 2023-11-18
Base page filters component


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Although not a part of the MVP, we could start making the component for the frontend that houses the filters for the various organization and event pages. Designs for this component can be seen on Figma in the [Organizations](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=517%3A4852&mode=design&t=hbrmPFJFE6NT0uV9-1) and [Events](https://www.figma.com/file/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?type=design&node-id=62%3A2&mode=design&t=hbrmPFJFE6NT0uV9-1) pages, with an example being below:

![Page Filters](https://github.com/activist-org/activist/assets/24387426/e2719a7a-69ff-4599-8822-f49d39235aa8)

We'd want a `filter` directory within `components/page` that has `PageFilter.vue` that defines the basic styles including some slots and conditions based on which type of page it's on. From there we'd have `PageFilterEvents.vue` and other versions of it that add all the functionality and texts needed for the given page. This issue would be to make `PageFilter.vue`, and then further issues would make the sub-components.

### Contribution

Happy to support someone or get to this myself! 😊 This base component is a `good first issue` that would be great for someone who wants to contribute for the first time. Also of course available to those already in the community :)

hariFED:
Hello @andrewtavis I am happy to work on this issue.

andrewtavis:
Awesome, @hariFED! Assigning you now :) Please let me know if there's anything I can do to assist!

hariFED:
“We need to create a PageFilter.vue component. Could you explain the functionalities that need to be added to this component?

andrewtavis:
Thanks for checking in @hariFED :) I'd envision `PageFilter.vue` to function in a similar manner to [CardSearchResult.vue](https://github.com/activist-org/activist/blob/main/frontend/components/card/search-result/CardSearchResult.vue) in that we'd be passing a `pageFilterType` to it that would then trigger one of the subclasses. Basically the functionality of `PageFilter.vue` is common styling so that we don't need to copy it into the other ones.

[...]

PR #1027, opened 2024-11-30
feat(filter): added filter common component

Hasteerp:
### Contributor checklist

---

### Description

I added a common reusable Filter component. 
ScreenRecording2024-11-30at12 56 16PM-ezgif com-video-to-gif-converter

The Filter component is a reusable component. 
It takes the following props: 
- HandleFilterChange
- tabs (id, name)[]
- sections (title, tags[])
     - tags: (id: number, name: string, selected: boolean)
     
This way you have full control over the tag title and tag buttons you want to populate. Additionally, decision can be made to check tab values and name based on each page requirement. It also takes handleFilterChange from the parent component that allows different filteration logic based on parent component requirement.

### Related issue

- [#550 ](https://github.com/activist-org/activist/issues/550)


Hasteerp:
I'll be removing the filter component usage from organizations before merging. It is currently there for reviewer to visualize how the filter component looks.

andrewtavis:
Looking forward to the review here, @Hasteerp! I'll try to get to this tomorrow or early next week 😊

Hasteerp:
@cquinn540 I was having a difficult time getting the test to run. Would you have any suggestions?

cquinn540:
> @cquinn540 I was having a difficult time getting the test to run. Would you have any suggestions?

I would need to know more about what issue you are facing.  I'll be a bit busy today, but I should have more free time to help over the next few days.

Hasteerp:
@cquinn540 I was facing some issues trying to get my component to render I think the custom render function needs to change but I am not exactly sure. This is the error message that I got. Please feel free to check it out thanks!

`This error originated in "test/components/filter/pageFilter.spec.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "test/components/filter/pageFilter.spec.ts". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- If the error occurred after the test had been completed, this was the last documented test before it was thrown.
`

cquinn540:
@Hasteerp Would you have time today or sooon to get on a call in Matrix chat so I can see what you are seeing?  That's the top level error message which is generic and doesn't give much information.  There's usually another stack trace somewhere in the output with more details.

Hasteerp:
I can get on a call tomorrow if that works for you.

cquinn540:
Sorry @Hasteerp I got into the zone working on another issue and didn't see this until now.  Let me know on the Matrix chat when you are available today.

Hasteerp:
@andrewtavis @cquinn540 Please check it out whenever you can. I added the test case.

frontend/test/components/filter/pageFilter.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { describe, it, expect } from "vitest";
+import { render, screen } from "@testing-library/vue";
+import PageFilter from "@/components/page/filter/PageFilter.vue";
+
```
cquinn540, line 4: Okay, I need to go back on what I said originally a little.  There is an appropriate case for a mock function here.  You should create a mock function like:
```typescript
const mockHandleFilterChange = vi.fn();
```
and you'll want to reset it after each test
```typescript
afterEach(() => {
  mockHandleFilterChange.mockClear();
});
```

frontend/test/components/filter/pageFilter.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+          { id: 2, name: "Tag 2", selected: false },
+          { id: 3, name: "Tag 3", selected: false },
+          { id: 4, name: "Tag 4", selected: false },
+        ],
+      },
+    ],
+    tabs: [
+      { id: "tab1", name: "Tab 1" },
+      { id: "tab2", name: "Tab 2" },
+      { id: "tab3", name: "Tab 3" },
```
cquinn540, line 21: Then you should use `mockHandleFilterChange` as the `@filter-change` prop

frontend/test/components/filter/pageFilter.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    ],
+  };
+
+  it("should render", () => {
```
cquinn540, line 25: Then you want a test where you click the different options and confirm that `mockHandleFilterChange` was called with the right arguments in the right order.

And another test where you tab through the options and hit enter and again confirm the mock was called correctly.

You can see the call list with [calls array](https://vitest.dev/api/mock#mock-calls) on your mock function.

Ideally you would also have a test checking that the focusing works correctly, but given your time limit I can be flexible on that.

frontend/test/components/filter/pageFilter.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+          BtnTag: true,
+          TooltipBase: true,
+        },
+      },
+    });
+
+    expect(screen.getByText("Popular Tags")).toBeTruthy();
+    expect(screen.getByRole("heading", { name: "Filter" })).toBeTruthy();
+    expect(screen.getByLabelText("Search to filter")).toBeTruthy();
+    expect(screen.getByRole("textbox", { id: "input-search" })).toBeTruthy();
```
cquinn540, line 44: With the other behavior tests testing for truthyness shouldn't be needed anymore

cquinn540 (changes requested):

andrewtavis:
I've made #1051 to do the follow up work to finalize the tests. We are on a bit of a deadline here for @Hasteerp, which I hope that we haven't missed, and the contributions till now are enough that we should bring them in and then continue when we have a bit more time.

frontend/test/components/filter/pageFilter.spec.ts
andrewtavis: Removed in my commit ✅

Hasteerp:
@andrewtavis thanks for the pull request and allowing me to merge this one. We are currently seeing an error in linting due to unused vars (this is coming from index.vue on organizations page). This is because we commented the code out but the parameters were defined. We can either remove them completely or comment out the code for now. Please let me know what you would suggest I do and I can make a commit accordingly. Thank you for taking into consideration my deadline. Appreciate it! I'll work on the test cases in the new PR in the coming week.

andrewtavis (approved):
Approving these with the assumption that the most recent commit fixes all the frontend linting issues, but I'll get to them if there are any other changes.

Really appreciate this as a first contribution, @Hasteerp, and hope that we can work with you again :) Let's definitely keep #1051 in mind in the coming weeks, but this is a great start on a relatively complex component 🚀

@cquinn540: Don't mean to disregard your review, and really appreciate the support you're providing to @Hasteerp! This component isn't MVP, so under the circumstances I'd like to bring it in and iterate 😊

cquinn540:
No problem this seems like a reasonable compromise

Issue #1028, opened 2024-12-01
Modify the Organization Model to Return Related Data on API Request


momanyisamuel:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

The current implementation of the Organization API does not include related data such as org_text and locations in its response. This makes it difficult for the front-end to retrieve complete information in a single request.

The task is to modify the Organization model and API view to return related data seamlessly, ensuring a more efficient and user-friendly response.

andrewtavis:
Closed by #1029 🚀 Thanks for the issue and being so persistent that we need a better solution, @momanyisamuel, and thanks for the implementation, @to-sta!

PR #1029, opened 2024-12-01
#1028 organization model and serializers to support event association…

to-sta:
…s and update admin display

### Contributor checklist

---

### Description

I worked with @momanyisamuel on this today. There is still quite some work that needs to be done here. The organization api, get the org_text now and we started working on org_events.

### Related issue

- #1028


andrewtavis:
Really nice, @to-sta and @momanyisamuel :) Seeing the responses from organizations for the first time just now 😊

andrewtavis:
Ok so post 58ed272 the `/organizations` route and organization About pages load in :) Still some work to do, but it's sooo much easier now! Thank you both so much!

andrewtavis:
I'm also not sure on the i18n pre-commit errors/warnings 🤔 @mattburnett-repo was getting the same as @to-sta, I believe, but then when I took them out they all ran fine for me. Let's keep an eye on this!

frontend/stores/organization.ts
```diff
@@ -96,96 +92,32 @@ export const useOrganizationStore = defineStore("organization", {
...
-          //       {}
-          //     )
-          // ),
-        ]);
+      const [responseOrg] = await Promise.all([
+        useAsyncData(
+          async () =>
+            await fetchWithoutToken(`/entities/organizations/${id}`, {})
+        ),
+      ]);
```
cquinn540, line 100: `Promise.all()` is no longer necessary since there is only one async request

andrewtavis: Thanks for pointing this out, @cquinn540!

andrewtavis (approved):
Alright :) So with this we're bringing all of the various data for orgs, groups and events in one call each. Thanks so much for the initial work here, @to-sta and @momanyisamuel! 🙌 Really is making things so much easier to work with. Thanks also for the review, @cquinn540!

Current state isn't perfect, but is good enough to bring in. The event page is loading the location from the backend, but the navigation is a bit buggy and some of the page elements aren't perfect (map marker color for some reason isn't reactive).

I'll look into this more this weekend! Would be great if people had time to check things a bit :)

Issue #963, opened 2024-09-02
Check and fix routing based on non-localized routes


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Because of the necessary change in #962, we now have non-localized routes for English as Netlify was breaking on them. We thus need to rework the way that routes are done to allow for non `/en` routes as well, so routes that lack any localization at all.

I'll look into this!

aadityanairaaq:
Hello @andrewtavis, could you please assign me to this issue?

Mohtelsayed:
Hello @andrewtavis, could you please assign me to this issue as well? Thank you!

andrewtavis:
Closed by #1059 and other related commits :) Thanks all for the great work! 🚀

PR #1059, opened 2024-12-11
 Fix Routing Logic for Non-Localized Routes and Ensure Robust Locale Handling

aadityanairaaq:
This pull request addresses issue #963, where routes were not being handled properly. Previously, isRouteActive assumed a localized route structure. Switching between localized and non-localized routes led to unexpected behavior in terms of resulting pages.

**Changes:**

1. **Locale Stripping Logic:**

Introduced helper functions to strip locale segments from path-based routes and route names. Ensures that core functionality (checking active routes and checking subpages) is stable, regardless of locale prefix. 

2. **Normalized Path Comparisons:**
Updated isRouteActive and related functions to first remove any leading/trailing slashes and locale prefixes before performing comparisons. This makes the logic more resilient to both localized (/en/events) and non-localized (/events) routes.

**Benefits:**
Ensures that tab switching and route comparisons work consistently, even after multiple interactions.
Simplifies maintenance by centralizing locale removal logic.

**Testing:**
Verified that switching between /en/organizations and en/events (as well as in other languages) now correctly reflects the active route and subpages.
Confirmed that non-localized routes (e.g., /organizations) are handled without errors.
Checked that existing localized functionalities remain intact.

### DIFF
--- a/frontend/utils/routeUtils.ts
+++ b/frontend/utils/routeUtils.ts
@@ -1,15 +1,16 @@
-// A helper array for supported locales. Adjust as needed.
+// Define the supported locales. Adjust as needed to match your project.
 const LOCALES = ['en', 'fr', 'de'];
 
 /**
- * Removes leading and trailing slashes from a path.
+ * Normalize a given path by removing leading and trailing slashes.
  */
 function normalizePath(path: string): string {
   return path.replace(/^\/|\/$/g, '');
 }
 
 /**
- * If the first segment of the given segments is a known locale, remove it.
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Thanks for the contribution, @aadityanairaaq :) Please note that giving maintainers edit rights to PRs helps us make minor adjustments to the PRs before bringing them in. The ideal of loading in the locales for the checks is a good one, but we can also load this in from the file where the locales are defined. We'll do any further changes that are needed!

Issue #698, opened 2024-02-10
Add Facebook Messenger to Modal Share Page Component


GrandSuccess87:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue will focus on adding support to share activist pages (event/organization/group) on Facebook Messenger via the modal share page component.  

This component lives in [frontend/components/modal](https://github.com/activist-org/activist/tree/main/frontend/components/modal). 

Icons would ideally come from one icon source on [Icones](https://icones.js.org/) such as [CoreUI Brands](https://icones.js.org/collection/cib).  

- You may also find the icons you need on [Bootstrap](https://www.figma.com/file/p0nFlH6k1C9rTBjYjjiAUV/Bootstrap-Icons-(Community)?type=design&node-id=0-1&mode=design&t=vUxUgAs8hBzCRHYc-0).

### Contribution

Happy to pair up with someone on this or tackle it myself :))

GrandSuccess87:
Note: [Modal Share Page Component #581](https://github.com/activist-org/activist/issues/581) is in progress

UnknownSean8:
Hi, I can work on this issue together with the other social media icons for the modal share page since I've worked on the Signal icon. Thanks!

andrewtavis:
Thanks @UnknownSean8! Let us know if we can support somehow :) :)

UnknownSean8:
Hi, sorry for the delay, I will finish this by this weekend.

andrewtavis:
Nice, @UnknownSean8! Looking forward to the PR :)

UnknownSean8:
Mhm, question: i18n checks doesn't support formatted strings?

For example: `toast(t(`components.btn_share_icon.opening_${props.text.toLowerCase()}`));`

andrewtavis:
No they don't. It is a drawback of the system, but it's the only way to do it without getting into some really weird code parsing.

andrewtavis:
Closed by #1070 🚀 Thanks for the dedication to getting the share modal up to a great level, @UnknownSean8! Really great to see this taking shape 😊 Would you have interest in looking into #904 now that we have a component to easily slot this styling into?

Issue #904, opened 2024-06-15
Mark some options in the share modal be "suggested"


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would look into design changes to the [frontend/components/modal/ModalSharePage.vue](https://github.com/activist-org/activist/blob/main/frontend/components/modal/ModalSharePage.vue) such that some of the options would be suggested and others wouldn't be. This would be a step of curation for the platform to communicate to users that some options are more safe than others to use. We'd also need a way to convey the reason why some options are suggested to the user, which is open to suggestions :)

### Contribution

Happy to work on this with people and discuss how best to have this function!

andrewtavis:
@GrandSuccess87: let me know what you think on this one!

Mahendradoak:
![github](https://github.com/user-attachments/assets/e6576764-b95d-4ee3-8df7-7803e8b7ab12)
should it be similar to this?

andrewtavis:
I'd say that we can maybe simplify it and add say an apostrophe to the icon and have a line under it that says that those are the ones that we'd suggest people use. This suggestion visual element should be included in the component for a `ModalSharePage` element such that we can toggle it. 

Would you like to work on this, @Mahendradoak? :)

Mahendradoak:
i would love to try

andrewtavis:
Sure thing :) Please let us know if you need any assistance! 

CC @UnknownSean8 here as well as he's been working on the share component and made the component we'll be editing 😊

Mahendradoak:
yes, will do

Mahendradoak:
The key updates i have tried to include:

New showSuggestions prop: This allows the component to conditionally display a suggestion note and icons next to specific social media buttons.

suggestedOptions array: A new array holds the names of suggested social media components (e.g., Telegram, Mastodon, Email, Signal), and these buttons are marked with a small suggestion icon when relevant.

Updated button components: Conditional rendering of the suggestion icon for the listed platforms.

andrewtavis:
Hey @Mahendradoak 👋 Checking in to see if you still plan on opening a PR for this issue :) Please let us know if you need any support! 😊

andrewtavis:
[...]

PR #1070, opened 2024-12-19
feat + refactor: added messenger sharing

UnknownSean8:
### Contributor checklist

---

### Description

Refactored the buttons from the share page as a component and added messenger sharing to the modal. Lmk what do you think about the refactoring, thanks!

### Related issue

- #698


andrewtavis:
Will pick this up and bring it in tomorrow morning, @UnknownSean8 :) If you have an idea on the failing frontend tests, then let us know. Specifically the type check errors in `pages/organizations/[id]/events.vue`.

frontend/pages/organizations/[id]/events.vue
```diff
@@ -27,7 +27,10 @@
         />
       </div>
     </HeaderAppPage>
-    <div v-if="organization.events?.length > 0" class="space-y-3 py-4">
+    <div
+      v-if="organization.events && organization.events.length > 0"
```
Comment (andrewtavis) [inline frontend/pages/organizations/[id]/events.vue:31]: Was what I was thinking too, but also weird that it's not needed for groups 🤔

andrewtavis (approved):
Thanks much for the refactor and the frontend fixes, @UnknownSean8! Really great to have all of this so well structured :) Will make future work like #904 a lot easier 😊

frontend/pages/organizations/[id]/events.vue
Comment (UnknownSean8) [reply frontend/pages/organizations/[id]/events.vue:31]: Mhm, I'll take a look at the groups later after work.

Comment (UnknownSean8) [reply frontend/pages/organizations/[id]/events.vue:31]: Oh, it's because the `Events` type was previously under the `dom-events` type by default, but since there is no default type for groups, it's defined as an `any`. The `any` type wouldn't have any problem under this case.

Issue #1076, opened 2025-01-11
Adding a licence identifier to files


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

We plan to add a license identifier to all `.vue`, `.ts`, and `.py` files. After some research, we’ve identified a concise and effective approach:

For `.ts` files:
```ts
// SPDX-License-Identifier: AGPL-3.0-or-later
```

For ` .vue` files:
```html
<!-- SPDX-License-Identifier: AGPL-3.0-or-later -->
```

For `.py` files
```python
# SPDX-License-Identifier: AGPL-3.0-or-later
```

The comment should be placed at the top of each file.

This task could be a great first issue, as it can be automated using a Python script. In the future, we could also integrate this script as a GitHub Action to check/enforce license checks.

### Contribution

Happy to help 😃

andrewtavis:
Maybe we can take a look at [GitHub projects that might include such checks](https://github.com/search?q=license%20check&type=repositories) @to-sta. Seems like there are a few, but who knows if one is doing what we'd like it to in a simple manner.

andrewtavis:
Closed by #1083 🚀 Thanks for the change, @Mahendradoak!

PR #1087, opened 2025-01-14
Issue #1076: added comment header for all files in the project corres…

Suz99:
…ponding styles for each type of files such as for .py it is //SPDX-License-Identifier: AGPL-3.0-or-later and so on

### Contributor checklist

---

### Description

Added a pre-commit hook to the project so that every time a project member commits a file, the comment will automatically be shown as a header.

### Related issue

- #1076


andrewtavis:
Hey @Suz99 👋 Thanks for the expansion here :) The shell script does a lot of changes to the files that in some cases are breaking 🤔 I also decided that I'd prefer if `__init__.py` not get this notice as I don't think we need to license an empty file. Would you be able to revert these changes?

I wrote [utils/check_license.py](https://github.com/activist-org/activist/blob/main/utils/check_license.py) to check the work in #1076. Would you maybe want to call that in a pre-commit hook and add it to the ones that run on commit? Any other ideas on how to expand this?

Suz99:
Hi @andrewtavis may I see the errors from the console and what part is it breaking? I can revert the change for `__init__.py`.

Regarding the [utils/check_license.py](https://github.com/activist-org/activist/blob/main/utils/check_license.py), We can add a new script in the `.git/hooks/pre-commit` directory to invoke `check_license.py` every time a commit is made. This will check all staged files for the correct SPDX header and block the commit if any files are missing or have incorrect headers.
We might need to adjust `check_license.py `to make it suitable for a pre-commit environment like Returning a non-zero exit code if it finds files without the required headers.
Hope this suggestion works, please let me know the next steps are.

Suz99:
Hi Andrew! I was going to make the PR but this issue is closed

andrewtavis:
Yes it's already been covered by another contributor. Sorry that it couldn't work out!

Issue #1047, opened 2024-12-09
Remove inheritance from e2e page models


cquinn540:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

Currently the page objects models use a combination of inheritance and pseudo-inheritance via a Proxy that make the code difficult to debug and maintain.

The page objects should be simple and just use composition with a page field and fields for components on the page that are their own objects.  Also the use of key value pairs for the locators makes the tests harder to read.

### Contribution

I will work on this.

andrewtavis:
Closing based on the work in #1091 🚀 Thanks, @cquinn540!!!

PR #1091, opened 2025-01-16
Simplify page objects

cquinn540:
### Contributor checklist

---

### Description

This refactors the page objects for the e2e tests.  The objects have been split into smaller, more independent components.  Less logic is hidden inside page object methods.  Also the tests have been split into directories based on the browser environment they will run in.

Some tests are still flaky on my machine.  Some of that can be fixed with waiting for components to be visible before clicking on them, but others I think are due to running playwright and the full application at the same time being resource intensive.

### Related issue

- #1047


andrewtavis (approved):
Looking at this locally with @cquinn540 right now :) We're dealing with an issue where GitHub Actions doesn't have the resources to run the e2e tests remotely, but tests do pass locally. Approved and all great on my end 🚀

Issue #1100, opened 2025-01-21
Continue i18nMap conversion


andrewtavis:
### Terms

- [x] I have searched [open localization issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+is%3Aopen+label%3Alocalization)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

In #1094 we did a partial update of all strings within the frontend to use the new `i18nMap` object. There are still some strings left to be converted. Finding entries that need to be changed can be done via searching for, `i18n.t('`, `i18n.t("`, `t('` and `t("` in the codebase :)

Would be great if someone had interest in converting the rest of the hard coded strings! 🌐🚀

andrewtavis:
CC @tfuent1, do you have interest in working on this some more? It looks like there are still some strings that need to be converted to using the `i18nMap` object :)

tfuent1:
@andrewtavis no problem at all I must have missed a few!

andrewtavis:
Thanks, @tfuent1! 🚀

andrewtavis:
Hey @tfuent1 👋 Just got off a call with @aasimsyed who's looking for an issue to ease back into the project :) We thought that this would be an easy one for him. Hope it's ok if he jumps in to help. You'd be welcome to review the PR and also finalize any changes if there are any! 😊

aasimsyed:
https://github.com/activist-org/activist/pull/1118/files

andrewtavis:
Ok @aasimsyed, I'm realizing there are a few other ways of finding keys that need to be replaced. Specifically we have cases where the keys are assigned with objects via `:` assignment, and also cases without `t()` where they're translated within the component. You can find them in the following ways:

- Searching for `="_global`, `="components`, `="pages` and similar ways that keys in `en-US.json` start (check the file for the options `types` as well comes to mind)
- Similarly `"_global`, `"components`, `"pages`, etc, so without the equals sign (I'd do this after the above so you get rid of the subset with equals first, with this being `:` assignment usually)
- Maybe there's another pattern you can see when checking the above?

So each of these needs to change to `:label="i18nMap.KEY", or similar for the ariaLabel or other attributes on components :)

We should be done after the above 😊 Feel free to open another PR and we'll finalize this 🚀

andrewtavis:
[...]

PR #1118, opened 2025-02-05
refactor(i18n): standardize localization key references across components

aasimsyed:
Issue: https://github.com/activist-org/activist/issues/1100

- Migrate all inline i18n keys to use i18nMap object for type-safe references
- Add consistent 'i18n.' namespace prefix to translation keys in:
  - EmptyState component buttons
  - CardGetInvolvedOrganization links
  - CardSearchResult aria labels
  - MediaMap control tooltips
  - ModalEditFaqEntry text checks
  - Group/resource creation pages
  - Sidebar navigation labels
- Update HTML title attributes to use translated strings via i18nMap
- Import i18nMap in files where missing (MediaMap, groupSubPages)
- Fix key consistency in discussion input component

### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/frontend/components/EmptyState.vue
+++ b/frontend/components/EmptyState.vue
@@ -3,7 +3,7 @@
   <div class="flex w-full flex-col items-center bg-layer-0 text-primary-text">
     <PageContent
       :imgUrl="BOOTSTRAP_CLOUD_MOON_URL"
-      imgAltText="components.empty_state.img_alt_text"
+      :imgAltText="i18nMap.components.empty_state.img_alt_text"
     >
       <div>
         <!-- Header -->
@@ -40,7 +40,7 @@
             $t(i18nMap.components.empty_state.message_no_permission)
           }}</span>
           <PageCommunityFooter
-            header="components.empty_state.cta_header_no_permission"
+            :header="i18nMap.components.empty_state.cta_header_no_permission"
             ><BtnRouteInternal
               class="w-full"
               :cta="false"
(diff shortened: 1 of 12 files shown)


aasimsyed:
@andrewtavis I didn't know how to fix this one from /frontend/components/combobox/ComboboxTopics.vue (lines 117-129):

```function displayValue() {
  if (inputFocussed.value) {
    return "";
  } else {
    return selectedTopic.value.id == 1
      ? "Filter by topic"
      : selectedTopic.value.name;
  }
}

const displayValueHandler = () => {
  return i18n.t(displayValue());
};```

**Comment by andrewtavis** [comment]
I'll get the formatting check in the review here, @aasimsyed :) No stress 😊

File: frontend/components/EmptyState.vue [hunk 1]
```diff
@@ -61,37 +61,37 @@
               v-if="pageType == 'organizations'"
               class="w-full"
               :cta="true"
-              label="components.empty_state.create_organization"
+              label="i18n.components.empty_state.create_organization"
```
andrewtavis, line 64: Bit of Vue learning here, @aasimsyed: What we do here is preppend `label` with `:`, which is short for `v-bind:label` ([see v-bind docs on W3 Schools](https://www.w3schools.com/vue/vue_v-bind.php)). This allows us to pass a variable instead of a string. I'll fix these quickly 😊

andrewtavis:
And in the [comment above](https://github.com/activist-org/activist/pull/1118#issuecomment-2637971105), good catch! We should include that string in the i18n values :) I'll add that in as well :)

andrewtavis (approved):
Nice, @aasimsyed! This is great 😊 I'm realizing that there are a few more that can be found in cases that are not the search methods that I defined in the issue, so let me detail those and then you can take those down as well :)

Wonderful to work with you again!

PR #1119, opened 2025-02-05
fix: standardize i18n string references

aasimsyed:
Issue: https://github.com/activist-org/activist/issues/1100

changes suggested by @andrewtavis 

- Replace hardcoded i18n path strings with i18nMap references in components
- Update button labels and aria-labels to use i18nMap paths
- Remove unnecessary i18nMap imports where direct strings are preferred
- Maintain consistent i18n usage pattern across components

### Contributor checklist

---

### Related issue

- [#ISSUE_NUMBER](https://github.com/activist-org/activist/issues/1100)


andrewtavis:
And would be amazing if you could get the merge conflicts as well, but let me know if you need some help with those! :)

andrewtavis:
I spoke too soon :) Thanks, @aasimsyed! 🚀

aasimsyed:
> Quick note that we'll need many of these component attributes to be `v-bind:` equivalents, where rather than `label=...` we'd have `:label=...` which then interprets the value in the string as a variable rather than a string, @aasimsyed :) In case you haven't installed the Vue plugin for VS Code, I'd do that as that'll then show you explicitly that it's reading in a variable instead of a string :) Sorry if you do have it installed, just making sure as there have been some changes 😊 You can check the recommended extensions in the dropdowns in the Environment Setup section of the readme :)

I'm not sure I understand what you mean here. You want those changes into this PR?

andrewtavis:
Yes these changes would need to happen on each component attribute where we're trying to pass the result of the object as a variable. It it's written now in some cases then the string `"i18nMap..."` will be passed. With the colon at the start the string is interpreted as a variable. We still leave the quotes on either side so it looks the same on the right side, but the colon on the left changes what's interpreted.

Let me know if it's still confusing!

aasimsyed:
> Yes these changes would need to happen on each component attribute where we're trying to pass the result of the object as a variable. As it's written now in some cases then the string `"i18nMap..."` will be passed. With the colon at the start the string is interpreted as a variable. We still leave the quotes on either side so it looks the same on the right side, but the colon on the left changes what's interpreted.
> 
> Let me know if it's still confusing!

@andrewtavis please review. I think I did what you were asking for. Let me know if there are any other issues.

andrewtavis (approved):
A couple more instances needed to get changed, @aasimsyed, and with that I did a few updates for things that weren't quite working after some dependency updates.

Thanks so much for bringing this home, @aasimsyed! Really awesome to have this setup finalized 🚀

CC @OmarAI2003 that the `i18nMap` object is now in full use over here. One extra thing I realized for `i18n-check` is that we need to give the user the ability for the map file to be written with a license header. I made the changes to add the current way we do it over here into the project, and then we'll just need to add an option for what the license for the project is :)

andrewtavis:
PS I ran the e2e tests and we're at the ~7 that are already failing, so we're good in those regards as well :)

Issue #839, opened 2024-04-24
Upload image functionality - Backend


bozmen:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Initially I was thinking of doing this, but I do not have enough time these days.

Here are the requirements for it: https://github.com/activist-org/activist/issues/670

I am right now working on the front-end, we can coordinate on this.

andrewtavis:
In discussing this with @to-sta, we're thinking that it's important that we keep the images in a place that's separate from the VPS for security reasons. This could be an Amazon S3 bucket, but ideally something other an S3 such as Digital Ocean (Hacktoberfest 🎃🧡).

andrewtavis:
Note for this that I need to make a [Hetzner](https://www.hetzner.com/) account still :)

mattburnett-repo:
writing in.

andrewtavis:
Closed by #1124 😊 Thanks so much for the great work, @mattburnett-repo, and also for the issue, @bozmen! 🚀

Issue #670, opened 2024-01-28
Image Upload Validation


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Let's discuss how we can protect against malicious user uploaded images/files?

We should consider protection mechansim on all level of our tech stack:
* Frontend 
* Webserver (we plan to use nginx)
* Backend 

Here is note from the django documentation on that:
[Django](https://docs.djangoproject.com/en/4.2/topics/security/#user-uploaded-content-security)

andrewtavis:
Information that I found on this during Code Brunch 🍞☕:
- `identify --warnings` from https://imagemagick.org/index.php
- Scrubbing image metadata on entry to the server as a way to make sure that if there is a breach that this information doesn't get into the wrong hands
- Have a different VPS for images
  - Copy images over into a compressed format that's then used on the main server

The following is also a useful chatbot output on the question "What are some steps that should be taken in a Django app to make sure that image uploads are not malicious?":

```
1. Use Django's Form Handling with FileField or ImageField: Django's forms provide built-in validation for file uploads. Use FileField or ImageField in your Django model forms to handle file uploads securely.

2. Implement Client-Side Validation: Utilize client-side validation using JavaScript to ensure that only certain file types are allowed to be uploaded. You can also check the file extension and file size before submitting the form to the server.

3. Server-Side Validation: Even though client-side validation is important, it can be bypassed. Therefore, always perform server-side validation as well. Django provides validators for file uploads. For example, you can use validate_image_file_extension to validate that the uploaded file is an image and not a disguised malicious file.

4. Use a Secure File Storage Backend: Django provides different file storage backends. Use a secure file storage backend like FileSystemStorage or Amazon S3 with proper access controls and permissions.

5. Implement Image Processing Libraries with Caution: If your application requires image processing, use trusted libraries like Pillow. Be cautious when accepting user input for image processing operations to prevent security vulnerabilities like Remote Code Execution (RCE).

6. Scan Uploaded Files: Consider integrating antivirus or file scanning services to scan uploaded files for malware or malicious content.

7. Limit File Upload Size: Configure Django settings to limit the size of uploaded files to prevent denial of service attacks and disk space exhaustion.

[...]

PR #1124, opened 2025-02-11
WIP - Issue 839

mattburnett-repo:
### Contributor checklist

---

### Description

This PR is a WIP and addresses the functional requirements of Issue #839.

The code in this PR:
- Allows organizations to upload files.
- Saves the binary image file to local folder backend/media/images.
- Links image id and org id in the database.
- Displays uploaded images in the frontend MediaCarousel components.

To do:
1. ~~Change the name of 'file_location' to 'file' or 'file_object' to make it clearer.~~ Done.
2.  ~~Error handling for cases in which user doesn't select an image file, but clicks 'Upload'.~~ Done. 'Upload' button not displayed if no files selected.
3. ~~Clean up 'Unknown @apply' error in MediaImageCarousel.vue.~~ Done. There's a Tailwind-recommended solution that adds a line to .vscode/.settings.json, and requires a .vscode/tailwind.json file.
4. ~~Provide the user with feedback that the upload either succeeded or failed. This can be a minimal line of text in the main upload modal.~~ Done. Added 'components.media_image_carousel.upload_error' to MediaImageCarousel.vue, en-US.json, and i18m-map.ts files.
5. Implement full CRUD.
   - We might not need Update functionality at this point, but we should at least stub in something to handle this in the future.
   - Do we want to have captions for images? That's the only thing I can think of that would make Update functionality necessary. However, I see captions as a potential security issue.
      - Free text fields are always possible entry points for bad actors.
      - Free text in this case allow users to inadvertently enter compromising information. 
   - We should have Delete functionality. Figure out where a control/button for this would sit in the UI.
6. Replace file storage at backend/media/images with an actual storage server.
   - There is discussion of this in #839.
   - Check MEDIA_BASE (or whatever it's called in DRF) variable.
   - Check docker compose file for image volume assignment.
      - There's a
      ```bash
          volumes:
            - ./backend/media:/app/media
       ```
       entry in the 'backend' section.
8. ~~Expand existing functionality to allow for image upload from Events / other entities TBD.~~
   ~~- Refer to social links functionality for a pattern to follow, regarding page type (group/org/event).~~ This doesn't look like a requirement at this time.
9. Assess and mitigate potential security issues associated with file upload.
   - Issue #670 has more discussion about this.
      - I've started consolidation / adding to the existing discussion.
11. Assess and possibly implement alternative file upload UI packages such as [filepond](https://pqina.nl/filepond/).
   - Maybe this is better suited as a separate issue.
13. ~~Clean up modal open/close functionality in 'ModalImage.vue' to use the newer modal store open/close functionality.~~ Done.
14. ~~The 'fullscreen' display option should display the current carousel image. Right now it displays one of the default 'fist' images.~~ This is fixed. Embedded and fullscreen/modal image carousels properly receive imageUrl data. This could be refactored into a composable or a store, if we're going to use file/image upload elsewhere in the app.

### Related issue

- #839 
- #670


backend/content/serializers.py
```diff
@@ -42,34 +41,33 @@ class Meta:
...
-                    code="invalid_extension",
-                )
-
-        else:
-            raise serializers.ValidationError(
-                _("The file location must be a string."), code="invalid_file_location"
-            )
-
+        # Remove string validation since we're getting a file object
+        if "file_location" not in data:
```
cquinn540, line 45: Is the `file_location` field a reliable way to tell if a file object is present or not?

Also is the `file_location` field just for validation, or is it being used / stored anywhere?

mattburnett-repo: 'file_location' is the actual file sent in the request. As far as I can tell it's the only way to tell if a file object is present or not. Not sure I understand what you mean by 'reliable' in this context. It's called every time a request is made, so to me that's 'reliable'. Maybe I don't understand your question completely.

A little further down in the file, around line 64-68, the file data gets added to the database. So it's used for validation as well as storing.

Not sure if I'm addressing your questions. Please let me know.

andrewtavis:
@mattburnett-repo, @cquinn540: I wanted to bring [filepond](https://github.com/pqina/filepond) to your two's attention. This is more of a quick "Hey could this be useful?" check in here. Was something that folks at Vue.js Berlin vouched for, but as I understand it this would be more to replace some of the frontend code we have?

mattburnett-repo:
filepond looks interesting, especially the streamlined upload, as well as image editing capability.

My first reaction to this is that it makes more sense as something we can add later in the project's life cycle. Right now we're still working on getting the basic image upload / management functionality in place. Once that's done, we should definitely look at ways to improve this functionality (especially from the perspective of the userbase), and filepond looks like a path towards that improvement. 

I'm also open to other possible solutions besides filepond, that we don't yet know about.

backend/content/serializers.py
cquinn540: I see, I didn't realize the `file_location` is the actual file object itself.  Based on the name I would assume that it's a string file path.  Is there a way to change the name of that field to say `file` or `file_object` to make it clearer?

mattburnett-repo: Yes, I was thinking of something similar after answering your last comment.

I'll add this to the list of to-do items at the top of this PR.

andrewtavis (approved):
🚀🚀🚀🚀🖼️🖼️🖼️

Great work @mattburnett-repo :) Really appreciate you bringing this super important feature home 😊

PR #1125, opened 2025-02-11
Added PyTest Coverage to backend

hboppana:
### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER


andrewtavis:
@to-sta is going to take a look at this and finalize the PR tomorrow, @hboppana :)

andrewtavis:
Do we want to pivot away from [MishaKav/pytest-coverage-comment](https://github.com/MishaKav/pytest-coverage-comment) then, @to-sta? I'm not sure if it makes sense to use it if we need to expand write permissions to forks 🤔 We can just have the warning for it dropping below 70 for now, and then this allows us to make some testing issues to expand the coverage on a file by file basis and we can then move it up to 80 and eventually 90?

andrewtavis:
If that's workable for now, can you let me know which args we wouldn't need from the `pytest` command in the workflow? Right now I have this locally:

```yaml
    ...
      - name: Run Django Migrations - Database Migrations Check
        run: |
          . .venv/bin/activate
          python ./backend/manage.py makemigrations
          python ./backend/manage.py migrate

      - name: Run pytest - Unit Tests
        run: |
          pytest ./backend --cov=backend --junitxml=pytest.xml --cov-report=xml --cov-report=term --cov-fail-under=70 | tee pytest-coverage.txt
```

But then I'm pretty sure that the command can just be`pytest ./backend --cov=backend --cov-fail-under=70`, right?

andrewtavis:
And could we maybe make a `. coveragerc` file so that we don't include `__init__.py`s and other non-important files in the tests? This'd make the output easier to read :)

I also tried to see if there's a per-file fail threshold so we can say no file is under X%, but looks like not.

to-sta:
> If that's workable for now, can you let me know which args we wouldn't need from the `pytest` command in the workflow? Right now I have this locally:
> 
> ```yaml
>     ...
>       - name: Run Django Migrations - Database Migrations Check
>         run: |
>           . .venv/bin/activate
>           python ./backend/manage.py makemigrations
>           python ./backend/manage.py migrate
> 
>       - name: Run pytest - Unit Tests
>         run: |
>           pytest ./backend --cov=backend --junitxml=pytest.xml --cov-report=xml --cov-report=term --cov-fail-under=70 | tee pytest-coverage.txt
> ```
> 
> But then I'm pretty sure that the command can just be`pytest ./backend --cov=backend --junitxml=pytest.xml --cov-report=xml --cov-report=term --cov-fail-under=70` without `| tee pytest-coverage.txt`, right?

This should be enough:

```
pytest ./backend  --cov=backend --cov-report=term --cov-fail-under=70
```

We only need the terminal (term) output in the workflow.

to-sta:
> And could we maybe make a `. coveragerc` file so that we don't include `__init__.py`s and other non-important files in the tests? This'd make the output easier to read :)
> 
> I also tried to see if there's a per-file fail threshold so we can say no file is under X%, but looks like not.

We do have a config for pytest, it is in the `pyproject.toml` file. I am on my phone otherwise I would have cited the code, but it should be simple to find. 

There are already excluded files like `tests_` and migration files. You can add `__init__.py` there.

I can also make the changes later if you want :smile:

andrewtavis:
I can get to it shortly :) Thanks, @to-sta!

andrewtavis:
All's green here now, but ignoring files from the coverage report output isn't working 🤔 I did edit `pyproject.toml`, but actually what's there already isn't being ignored either as can be seen in [the workflow run](https://github.com/activist-org/activist/actions/runs/13552208876/job/37878273863?pr=1125). The coverage is still being ran on `migrations` files. 🤔

.github/workflows/pr_ci_backend.yaml
```diff
@@ -68,12 +72,12 @@ jobs:
...
-        run: pytest ./backend -vv
-
       - name: Run Django Migrations - Database Migrations Check
         run: |
           . .venv/bin/activate
           python ./backend/manage.py makemigrations
           python ./backend/manage.py migrate
+
+      - name: Run pytest - Unit Tests
+        run: |
```
Abhi-Bohora, line 82: @andrewtavis maybe we can try this explicitly telling  pytest-cov to use our `pyproject.toml` settings, when i ran this locally it ignore the `__init__.py` files and i didn't see any `migrations` dir with files
```yaml 
run: |
            pytest ./backend -vv --cov=backend --cov-report=term --cov-report=html --cov-config=backend/pyproject.toml
```

andrewtavis: Thanks, @Abhi-Bohora! I'll try this out this evening and hopefully we can finalize this 😊

to-sta:
I made the change @Abhi-Bohora suggested and I added `factories` files to be excluded. I made some changes to the report. The report is more consice now because we skip all covered files and we show which lines are missing testing.

andrewtavis:
Interesting 🤔 I did the calculation of the percentages for the coverage threshold and it's only based on the files that are not fully covered no matter if you have `:skip-covered` or not. I included the other files as I feel it'll make it easier to onboard people into backend testing if they can see the full coverage report in the PRs, but we can add that back in at some point if it gets too long. Also dropped the threshold to 60% for now, but we can raise it once we have a few more files covered :)

andrewtavis (approved):
Thanks for the great initial commit here @hboppana as well as the review @Abhi-Bohora and the fine tuning, @to-sta! This is a great result for the last issue for our starting with testing process ✨ Wonderful work and collaboration, all! 🚀

andrewtavis:
FYI all I just made #1132, #1133 and #1134 in response to the coverage report 😊 We can finalize those and then move on to other files! 😊

Issue #1097, opened 2025-01-19
Update i18nMap on frontend build


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

activist adopted the use of the `i18nMap` object in #1094. One thing that's needed is that the functionality to update the map should be ran each time the frontend is built to make sure that new developers in the community don't run into issues when they update the i18n keys and forget to update the object. Ideally this would be the [frontend/i18n/check/i18n_generate_map_object.py](https://github.com/activist-org/activist/blob/main/frontend/i18n/check/i18n_generate_map_object.py) script being ran on yarn build, but we can also look into other solutions here :)

### Contribution

Happy to help devise a solution to this and work with people to implement this functionality! 😊

Suz99:
Hi @andrewtavis! Can I work on this?

andrewtavis:
Yes, @Suz99! Thanks for your willingness to take this on! 😊

Suz99:
Hi @andrewtavis! I had a doubt, if I wanted [i18n_generate_map_object.py](https://github.com/activist-org/activist/blob/main/frontend/i18n/check/i18n_generate_map_object.py) to run on yarn build, I would have to update the a `json` file with `"frontend/i18n/check/i18n_generate_map_object.py && yarn webpack"` to execute the py script. Can you direct me to where I should add this to the project?

andrewtavis:
Hey @Suz99 👋 Checking in here to see if you have an idea on the above? :) I think your idea could work. We'd need to make that file executable, but then this will likely shift to a Python dependency in the coming weeks, so maybe it makes sense to hold off on that and then we can use the CLI to run the generation? Let us know if this makes sense :)

Suz99:
Hi @andrewtavis It makes sense. My idea was to add this `frontend/i18n/check/i18n_generate_map_object.py && yarn webpack` in the build of package.json. This needs to be part of the documentation to run this script.

andrewtavis:
[...]

PR #1130, opened 2025-02-18
#1097 Remove i18nMap and switch to regex based invalid key check

andrewtavis:
### Contributor checklist

---

### Description

This PR addresses development and performance issues introduced by the `i18nMap` object by replacing it with a method that's more suitable for the project. Note that the [i18n-check](https://github.com/activist-org/i18n-check) project has gotten this change in parallel.

Issues with the `i18nMap` include:
- It's something that would need to be updated, which is confusing for developers
    - This necessitated #1097, but rather than fix this we should find a better solution
- In loading in `i18nMap` to most frontend files, we will doubtless experience some performance issues in that we'll have a JSON object with 1,000s of keys being loaded into most components and pages

The solution to the above issues is twofold:
- Switch over the check of invalid i18n keys (those that we've incorrectly typed, etc) from a type check to a regex check
- In order to allow for a regex check of all i18n keys, they should all be prepended with `i18n.`
    - This allows us to find them with `r"\"i18n\.[_\S\.]+?\""` and similar patterns

With this the i18n process for the project is set. Further updates to this will come after [i18n-check](https://github.com/activist-org/i18n-check) has been released via a version update over here :)

### Related issue

- #1097, to a degree as it will be closed


andrewtavis:
CC @mattburnett-repo, @cquinn540, @aasimsyed:

Wanted to bring these changes to your all's attention. We can talk about this a bit more in the sync, but the `i18nMap` was a meh idea that ultimately lead to a better one 😊 If we're saying that i18n keys should be accessed via an object such that we'll have `i18nMap.*` for all i18n references, why not just prepend all strings with `i18n.` such that we can use regex to find them? This means that:
- We don't need to update the map object as we're back to using the strings
- We have a new regex check to see if any keys used in the project are not included in the `en-US.json` file
- We don't need to import `i18nMap` into most files in the frontend and doubtless deal with negative drawbacks of dev experience and performance from loading a large object everywhere

@omarai2003: The changes to the checks have already been made in `i18n-check`, so you're still free to work on https://github.com/activist-org/i18n-check/issues/2 😊

andrewtavis:
The above changes came post check our strategies at the Vue.js Berlin Meetup :)

andrewtavis:
Bringing this in so that we can finalize this quick pivot 😊

Issue #564, opened 2023-11-18
Add rich text functionality to CardDiscussionInput


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to investigate various Vue/JS/TS plugins to allow us to add markdown and rich text functionality to [CardDiscussionInput.vue](https://github.com/activist-org/activist/blob/main/frontend/components/card/discussion/CardDiscussionInput.vue). We should consider a few possible options, and then this issue can shift to implementing the changes 😊

### Contribution

Would be happy to discuss options with people and work on implementation or support someone with it!

andrewtavis:
Looking for options for this, for RTE we have the following:

- [tiptap](https://github.com/ueberdosis/tiptap)
- [tinymce](https://github.com/tinymce/tinymce)

For markdown we could use the following:

- [vue-markdown-editor](https://github.com/code-farmer-i/vue-markdown-editor)
- [Mavon Editor](https://github.com/hinesboy/mavonEditor)

andrewtavis:
A quick note on this, it'd be great if we'd also do some styling as a part of this such that the options are only present for a text field if it's currently active in the same what that it is for GitHub :)

andrewtavis:
Note that this issue is now only for RTE, and likely the solution we'll be using [tiptap](https://github.com/ueberdosis/tiptap) :)

bharath637462:
@andrewtavis can i work on this?

Mister-Tofu-tech:
@andrewtavis . Hopefully I can help out with this.

andrewtavis:
Give it a shot, @Mister-Tofu-tech :) Happy to support! We can also do a quick call to discuss if that would help.

andrewtavis:
Hey @Mister-Tofu-tech 👋 Let me know what the status here is :) We're getting to a point where this issue will be a bit higher priority. I can also take it over if you're pressed for time :)

Mister-Tofu-tech:
Hey @andrewtavis , sorry I wasn't able to make progress towards this issue. Life has been busy.
Hopefully, I can work on something with lower priority?

andrewtavis:
[...]

PR #1135, opened 2025-03-01
markdown tiptap feature

UnknownSean8:
### Contributor checklist

---

### Description
Two features are still pending:
1. Does The mention require a model? or we just use the built-in one from tiptap?
2. The link as of now only supports links pasted into the editor, however, it can be configured to insert a link with a model to enter the url.
3. The preview function is not done, it now only supports write and preview together.

Lmk what you think about the current product and if there's anything to change :) Thanks!

### Related issue

- #564


UnknownSean8:
Three*

andrewtavis:
Checking this out now, @UnknownSean8 :)

1. Let's stick with the built in one for now and we can move to a custom one if needed later once we start actively experimenting here.
2. Would this be with the link button? Would be nice to have this in case people don't understand how it's working :)
3. Maybe we can also just not do preview like this for now and we can go more towards WYSIWYG editor with markdown added in later.

Do you want to work on point 2 above and then we can merge?

UnknownSean8:
Hii

1. Okay, they were also using external dependency, tippy.js in this case, should I proceed?
2. Yeap, the only problem now is I can't type `/` in the editor, because it will always bring me to the search bar, is there a solution for this already?
3. Okay, I will also try to implement it in the meantime, if not I will focus on the first two points.

Thanks!

andrewtavis:
Hiiii 👋

1. Fine to use that dependency :) 
2. Maybe we disable the search bar hotkey when we're focusing the input? How does that sound? 
3. Sounds good! 😊 

Thank you! 😊

UnknownSean8:
Hii, fixed all of it except the preview feature. Thanks!

andrewtavis:
Thanks, @UnknownSean8! I'll review this and bring it in and then we can discuss further issues :) This should be good to close #564 and we can make new issues for specific additions to this 😊

andrewtavis (approved):
Looking really great, @UnknownSean8! Thanks so much for the amazing work here and the fun review session 😊

Issue #1149, opened 2025-03-09
Update mentionSuggestion.js to TypeScript


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

For the discussion input component we have a [frontend/utils/mentionSuggestion.js](https://github.com/activist-org/activist/blob/main/frontend/utils/mentionSuggestion.js) utility file to control the @ mentions within the component. This file is written in JavaScript, and it would be great if it could be ported to TypeScript with all the necessary types.

Happy to support on this and review! 😊

annaleexjohnson:
Hi! I'm interested in picking this up!

andrewtavis:
Thanks for picking this up, @annaleexjohnson! Please let us know if you have any questions 😊

Danbaekman:
Hi!
I'd love to work on this issue and port mentionSuggestion.js to TypeScript.
Could you confirm if I can proceed with this task?

Looking forward to your response!

andrewtavis:
Hi @Danbaekman 👋 @annaleexjohnson already has a PR up for this in #1154. You'd be welcome to make suggestions to improve it! Aside from that, please let us know if you find another issue of interest 😊

Thanks for your interest in contributing!

andrewtavis:
Closed by #1154 🚀 Thanks for the amazing first contribution, @annaleexjohnson! Hope that we can work with you again :) Please let us know if there's another issue of interest!

PR #1154, opened 2025-03-12
Update mention suggestions to typescript

annaleexjohnson:
### Contributor checklist

- [X] This pull request is on a [separate branch](https://github.com/annaleexjohnson/activist/tree/update-mentionSuggestions-to-typescript) and not the main branch
- [X] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description
Refactored mentionSuggestion.js to Typescript and added a types file

When running yarn typecheck, there was a type error when initializing the **popup** variable. I set the type to _Instance_ based on the [tippy.js docs](https://atomiks.github.io/tippyjs/v6/tippy-instance/), but it appears there's still an error.
tippy_type_error

Please let me know if there's a better way to assign this type! Thank you!

### Related issue
- #1149

### DIFF
--- a/frontend/components/card/discussion/CardDiscussionInput.vue
+++ b/frontend/components/card/discussion/CardDiscussionInput.vue
@@ -159,6 +159,7 @@ const editor = useEditor({
         class:
           "hover:underline font-bold rounded-2xl box-decoration-clone px-1 py-0.5",
       },
+      // @ts-expect-error Ignore mismatched types.
       suggestion: Suggestion,
     }),
   ],
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -6,6 +6,7 @@ header:
     - '**/*.ts'
     - '**/*.test.ts'
     - '**/*.d.ts'
+    - '**/*.js'
     - '**/*.py'
   paths-ignore:
     - 'dist'
--- a/frontend/components/media/MediaMap.vue
+++ b/frontend/components/media/MediaMap.vue
@@ -217,6 +217,7 @@ function resetRouteProfileControl() {
             requestOptions: {
               alternatives: "true",
             },
+            // @ts-expect-error Will break route profile change.
             mapLayers,
           });
 
--- a/frontend/components/media/MediaMap.vue
+++ b/frontend/components/media/MediaMap.vue
@@ -217,7 +217,7 @@ function resetRouteProfileControl() {
             requestOptions: {
               alternatives: "true",
             },
-            layers: mapLayers,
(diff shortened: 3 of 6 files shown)


andrewtavis:
Hey @annaleexjohnson 👋 Quick feedback on this :)

- Can we get the license header included in `frontend/utils/mentionSuggestion.ts` so that we're compliant with the AGPL
- I'm realizing that that was missing because [.licenserc.yaml](https://github.com/activist-org/activist/blob/main/.licenserc.yaml) was missing tracking on JS files :)
    - Could you add JS files to the ones that will be checked and make sure that the license header is on the remaining JS files that are in the repo?

From there I am seeing the TS errors, as you noted. 100% ok to open a PR with those still in it, and I'll see if I can find a fix for this after the changes above are sent along.

Thanks for the great first contribution so far! 😊

annaleexjohnson:
Thanks! Just made the changes- please let me know if I should add anything else!

GoodMorningA1i (commented):
Great conversion to typescript 👍

In terms of the error, did you try single quotation marks? So instead of "body", can you try 'body'?  I noticed single quotation marks were used in the tippy docs. 

The argument string "body" seems to be causing the error. In the docs, they typically used the attribute or content prop rather than the string as an argument to `tippy()`. I was specifically looking at this page in the docs: https://atomiks.github.io/tippyjs/v6/constructor/

frontend/components/card/discussion/CardDiscussionInput.vue
```diff
@@ -159,6 +159,7 @@ const editor = useEditor({
         class:
           "hover:underline font-bold rounded-2xl box-decoration-clone px-1 py-0.5",
       },
+      // @ts-expect-error Ignore mismatched types.
       suggestion: Suggestion,
```
andrewtavis, line 163: This was the one error that I wasn't sure on ignoring 🤔 If either of you have an idea on this, @annaleexjohnson or @GoodMorningA1i, then we'd welcome another PR to fix it!

andrewtavis: For now I'd like to bring this in :)

andrewtavis (approved):
Thanks for the great contribution here, @annaleexjohnson! As mentioned above, if you have an idea on the type error on the `suggestion: Suggestion` assignment within CardDiscussionInput, then by all means send along a change! Really appreciate the quick work here! 🚀

Issue #1147, opened 2025-03-08
Remove extra messages from frontend typecheck output


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

As of now we're getting a lot of noise back from the typecheck call that we run for GitHub Actions workflows. The command that's ran within the frontend directory is:

```bash
yarn typecheck
```

Outputs of this as of now include:

```
[@nuxt/content 7:19:44 PM]  WARN  No content configuration found, falling back to default collection. In order to have full control over your collections, create the config file in project root. See: https://content.nuxt.com/getting-started/installation


[@nuxtjs/i18n 7:19:45 PM]  WARN  Ignore Vue I18n configuration file does not exist at ./i18n.config.ts in PATH_TO_MY_ACTIVIST/activist-org/activist/frontend. Skipping...


[nuxt:tailwindcss 7:19:45 PM]  WARN  The provided Tailwind configuration in your nuxt.config is non-serializable. Check plugins. Falling back to providing the loaded configuration inlined directly to PostCSS loader.. Please consider using tailwind.config or a separate file (specifying in configPath of the module options) to enable it with additional support for IntelliSense and HMR. Suppress this warning with quiet: true in the module options.

ℹ Using Tailwind CSS from ~/assets/css/tailwind.css
FILE_CONTENTS_EXPORTED
```

It would be great if we could clean up the output of this so it's a bit more clear what the result of the typecheck is :)

Happy to support as I can here! 😊

andrewtavis:
Closing this as the important part was for the typecheck in the dependency check for #1090. I don't think there's a good way to remove it from the actual typecheck run.

Issue #1090, opened 2025-01-15
Issues with check_dependency_updates


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

In #724 we developed [check_dependency_updates](https://github.com/activist-org/activist/blob/main/.github/workflows/check_dependency_updates.yaml), but this workflow has proven to not work particularly well as seen in #1089 and #1011. Issues are:

- The data for the issue title isn't formatted
- The updates that would happen aren't added to the issue text
    - We'd like the full error text
- The frontend errors seem to not be being included

Ideally we'd have this workflow running every two months and we'd then get a report that would tell us which dependencies would cause new errors so that we can react to these changes.

From the original issue #724:

- We'll keep the current Dependabot warnings and react to them as needed
- We'll run a workflow every two months
- This workflow will update all dependencies to the next available minor release
- The backend and frontend checks will be ran such that new errors will be detected
- Results will be logged and new issues will be made to work on them
- This two month process will coincide with Code Night 🌙 such that we can work together to squash 'em all 🐞

Would be great if someone wanted to look into this! 😊

andrewtavis:
Would be great to get your help here, @Abhi-Bohora! Thanks for taking this on :)

Abhi-Bohora:
our current [check_dependency_updates.yaml](https://github.com/activist-org/activist/blob/main/.github/workflows/check_dependency_updates.yaml) creates two `error_log.txt` files one on the `frontend`dir  and one on the `root` dir.

![Image](https://github.com/user-attachments/assets/2cc2c9cf-d9cd-418b-80c0-8b58c5607236)

 the issue we are seeing here [#1011](https://github.com/activist-org/activist/issues/1011) comes from `error_log.txt` file which is on the `root`. It seems like `error_log.txt` which is on the `frontend` dir seems to be ignored, but in there also there is not much information just this only
 ```
Type check failed.
ESLint check failed.
```

[...]

PR #1155, opened 2025-03-12
fix: frontend and backend test workflow

Abhi-Bohora:
As said in the issue [comment ](https://github.com/activist-org/activist/issues/1090#issuecomment-2708444053) there was error while running the backend test in workflow `ModuleNotFoundError: No module named 'app'` and also as said in the [1147](https://github.com/activist-org/activist/issues/1147) the typecheck result in the frontend contains too much information and not the TypeErrors.

 I have updated the workflow to eliminate the backend error `ModuleNotFoundError: No module named 'app'` it was due to passing the wrong setting file in `DJANGO_SETTINGS_MODULE` and update the frontend test typecheck result to show only the typecheck errors if it exists.

### Contributor checklist

---

### Related issue

- [#1090](https://github.com/activist-org/activist/issues/1090)  [#1147](https://github.com/activist-org/activist/issues/1147)

### DIFF
--- a/.github/workflows/check_dependency_updates.yaml
+++ b/.github/workflows/check_dependency_updates.yaml
@@ -152,7 +152,7 @@ jobs:
 
           echo -e "\n### Django Tests" >> ../reports.md
           echo "\`\`\`" >> ../reports.md
-          DJANGO_SETTINGS_MODULE=app.settings.test python manage.py test 2>&1 | tee -a ../reports.md || true
+          DJANGO_SETTINGS_MODULE=core.settings python manage.py test 2>&1 | tee -a ../reports.md || true
           echo "\`\`\`" >> ../reports.md
 
       - name: Run Frontend Tests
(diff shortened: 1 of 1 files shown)


Abhi-Bohora:
@andrewtavis Hello 👋 for the frontend typecheck results i have extracted result to show only the errors  if its exists but i have one issue when i run this workflow locally. sometimes it shows it finds the typeerror and update `typecheck results` section in `reports.md` file and sometimes it doesn't show any typecheck error and empty when running workflow again.
### Type Check Results
```
ModalUploadImagesFileDropZone.vue(11,13): error TS18046: '__VLS_ctx.$refs.file' is of type 'unknown'.
[error] Process exited with non-zero status (2)
[error] Process exited with non-zero status (2)
``` 
 I don't know why this occurs maybe i am running this locally that's or why i don't know. Expected behavior is it should always show the type error in the `typecheck results` section if there is an typerror

andrewtavis:
Hmmm, I'm not too sure on why this would happen either 🤔 Maybe the frontend dependencies are being overwritten by the workflow and then there's a different result? This could be related to another thing that happens in the code base where when we run the Docker container the node modules are replaced for Linux based dependencies. Maybe something along those lines?

CC @cquinn540: Maybe you have an idea here?

andrewtavis:
9e18754 adds a step before that we use on other projects, @Abhi-Bohora :) Specifically it just makes sure that the workflow is only being ran on the remote main as we don't need this running on all of the forks 😊

andrewtavis:
I'm personally fine with merging this in now so long as the runs on the remote are consistent. I assume that that local runs are OS based differences creating minor changes. We can make good first issues to fix these issues as they come up 😊

How does this sound, @Abhi-Bohora?

Abhi-Bohora:
@andrewtavis 👋 yes that sounds good, let's see how this goes.....👀

andrewtavis (approved):
Bringing this in post discussions about it 😊 Thanks for your continued support here, @Abhi-Bohora! 🚀 Hopefully this does the trick :)

Issue #1077, opened 2025-01-11
API endpoint organizations 


to-sta:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

Currently, we retrieve a list of texts in the following format:

```json
"texts": [{"id": 75, "iso": "en", "primary": true}]
```
This provides all available `iso` codes. However, we want to enhance functionality to retrieve an organization's text based on a specific `iso` value, as demonstrated below:

`http://localhost:8000/v1/communities/organizations?iso=eng`

If no `iso` value is provided, or if a query for a value like "xyz" does not return a result, the default should fall back to "eng."

To achieve greater flexibility, we also want to transition to an `APIView` and `Serializer`, replacing the current `ViewSet` and `ModelSerializer`.

In order to make the api implement that more flexible we want to move to a APIView and the serializers.Serialer instead of the `ViewSet` and `ModelSerializer`

APIViews are not handling List and Detail Views, therefore we are going to split the views into:

```py
class OrganizationListAPIView(APIView):
      pass
        
class OrganizationDetailAPIView(APIView):
      pass
```

### Contribution

Happy to work on this myself :smile

cquinn540:
I have an additional idea for how design this in the backend.

I think any entity model with translations (Organization, Event, Group, etc.) should have a mandatory `default_iso` field.  Also, all translated texts should in the texts OneToMany Field, including the title. 

1. Create endpoints should require a `defaultIso` argument.  The create endpoints should always create a row in the entity's corresponding `texts` table for the default iso.  That way we know it will always exist (unless someone manually messed around in the db).
2. When we fetch entities we send the `iso` that we want.  If that texts row for that iso does not exist, then we use the `default_iso` to fetch the default texts row.
3. We should not allow users to delete the default iso row from the texts table for an entity.

cquinn540:
This design will allow our users to decide what they want the default language for their Event / Org / Group instead of defaulting to English

to-sta:
> I have an additional idea for how design this in the backend.
> 
> I think any entity model with translations (Organization, Event, Group, etc.) should have a mandatory `default_iso` field.  Also, all translated texts should be in the texts OneToMany Field, including the title. 

I think that is a good idea, that gives users an individual fallback language

[...]

PR #1159, opened 2025-03-15
implementing org apiview GET method

to-sta:
### Contributor checklist

---

### Description

Implemented a paginated org APIView for the organizations endpoint and added testing for the endpoint.

###  

- #1077


andrewtavis:
Not sure on the social link test in the backend, @to-sta, but for the license header check just add the one liner that you see on the top of most files to the ones that are missing it 😊

Thanks for opening this!

mattburnett-repo:
@andrewtavis @to-sta Saw something about the tests for social links; anything I can explain / help with?

andrewtavis:
Ya I was going to ping you on this, @mattburnett-repo. Do you have an idea on why the changes would be causing them to fail?

Line is:

```

FAILED backend/content/tests/test_social_links.py::test_organization_social_link_list_view - assert 404 == 200
```

mattburnett-repo:
The test is simple and  basic. Here's the line we care about:

response = client.get(f"/v1/communities/organizations/{org.id}/")

This should just pull a single big org object, which has social link info as part of the composite object. The test then just looks at the attached soc link info. Really simple, but the test fails before this part.

So a 404 means, for the get request, that either there is no more GET (list) method,  the url part is broken, or the org.id part is.

===

I noticed in the backend/communities/urls.py that is part of this PR, that lines 25-27

router.register(
    prefix=r"organizations", viewset=OrganizationViewSet, basename="organization"
)

have been removed and replaced with lines 58-60

urlpatterns = [
    path("", include(router.urls)),
    path("organizations/", OrganizationAPIView.as_view()),
]

It looks like the only real difference is that there are two different Views/ViewSets. Beyond that, I don't know anything about the new code.

So maybe this is the place to start, to figure out what the solution should be? I don't know what the intent of the new code is, so I can't say anything about it. One guess is that the '.../organizations/' url doesn't take an org id at the end of the path anymore?

===

If the new code / endpoint in the PR is what we want, we could simply adjust the test to fit the newer code. That would probably be a simple matter of changing the URL that the test uses.

to-sta:
@mattburnett-repo maybe I should added a `WIP` in the description 😄. Finished adding the GET (list) and POST method but the detail view is not implemented yet. Therefore your test is failing.

mattburnett-repo:
Makes total sense, @to-sta! Thanks for the detail info :)

to-sta:
I recommend changing the status field from a ForeignKey to a ChoiceField. The current implementation is overly complex and causes issues with pytest during teardown—likely a classic 'chicken-and-egg' problem 😄 . Ultimately, I resolved the issue by removing the default values.

andrewtavis:
Thanks for the explanation and checking in here, @to-sta and @mattburnett-repo :) Requested reviews for when this is ready. Please let us know when it's ready, @to-sta! 😊

mattburnett-repo (approved):

mattburnett-repo:
Re-requested review because I may have approved too early. Let me know when we're ready for review :)

to-sta:
Ok 😕. Locally the test are running just fine:

```python
communities/organizations/tests/test_api.py::test_organizationDetaiAPIView PASSED                                                               [ 40%]
communities/organizations/tests/test_org_events.py::test_multiple_events_per_org PASSED                                                         [ 42%]
```

But during the GitHub workflow django is not able to load the fixtures:

```python
FAILED backend/communities/organizations/tests/test_api.py::test_organizationDetaiAPIView - communities.models.StatusType.DoesNotExist: StatusType matching query does not exist.
ERROR backend/communities/organizations/tests/test_api.py::test_OrganizationAPIView - django.core.management.base.CommandError: No fixture named 'status_types' found.
```

to-sta:
Ok, got it fixed with using the pathlib libary to get the fixture folder location.

Almost done with the PR

- ~~GET (List)~~
- ~~GET (Detail)~~
- ~~POST~~
- ~~DELETE~~
- PUT

to-sta:
### Implemented Final Method & Extended API Documentation.

In this update, I implemented the last remaining method and extended the API documentation. However, I encountered an issue caused by the djangorestframework-camel-case package in combination with drf-spectacular.

### Issue Description
Requests made via SwaggerUI are incorrectly transformed. For example:

Expected URL:
```bash
http://localhost:8000/v1/communities/organizations/3fa85f64-5717-4562-b3fc-2c963f66afa6/
```

Actual URL:
```bash
http://localhost:8000/v1/communities/organizations/%7Borg_id%7D/
```

### Debugging Attempts
Suspecting the issue might be related to the camelCase parser, I tested my theory by temporarily disabling it. I commented out the following lines in the settings:

```python
"DEFAULT_PARSER_CLASSES": (
    "djangorestframework_camel_case.parser.CamelCaseFormParser",
    "djangorestframework_camel_case.parser.CamelCaseMultiPartParser",
    "djangorestframework_camel_case.parser.CamelCaseJSONParser",
),

...

"CAMELIZE_NAMES": False,
"POSTPROCESSING_HOOKS": [
    "drf_spectacular.hooks.postprocess_schema_enums",
    "drf_spectacular.contrib.djangorestframework_camel_case.camelize_serializer_fields",
],
```
Any insights or suggestions are welcome!

to-sta:
I found a solution / workaround, to not use snake_case url parameters in the url patterns.   

```python
urlpatterns = [
    path("", include(router.urls)),
    path("organizations/", OrganizationAPIView.as_view()),
    # old
    path("organizations/<uuid:ord_id>/", OrganizationDetailAPIView.as_view()),
    # new
     path("organizations/<uuid:id>/", OrganizationDetailAPIView.as_view()),
]
```

andrewtavis (approved):
Thanks for the paginated APIs here, @to-sta! Definitely a major step in the right direction :) I'll look into the groups and events for this later today/tomorrow 😊

Big thanks to @momanyisamuel for the call to check this out as well 🥳

PR #1166, opened 2025-03-18
missing test for backend/communities/organizations/models.py 🧪

Abhi-Bohora:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

This PR covers the missing test for `backend/communities/organizations/models.py           71      4    94%   71, 79, 88, 131`

This test the string representation methods of several model classes in `organizations/models.py`

### Related issue

- [#1134](https://github.com/activist-org/activist/issues/1134)

PR #1167, opened 2025-03-18
fix: load i18n in tests without typescript error

cquinn540:
### Contributor checklist

---

### Description

This should fix the issue pointed out in https://github.com/activist-org/activist/pull/1166

Our frontend vitest tests broke recently because a configuration field "locales" was being overwritten causing a typescript error.  This is happening because we have to reload i18n in vitest for our frontend tests to work.

There is an ongoing issue with `@nuxt/test-utils` that makes this necessary because the testing environment has multiple vue app instances in it and doesn't reference them in a consistent manner: https://github.com/nuxt-modules/i18n/issues/2637

This PR changes how i18n is initialized in vitest to get around the typescript error.  This PR also adds the `--silent` flag to running the tests in CI to reduce the amount of Nuxt related warnings that probably need to be resolved in Nuxt itself.

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/.github/workflows/pr_ci_frontend.yaml
+++ b/.github/workflows/pr_ci_frontend.yaml
@@ -53,4 +53,4 @@ jobs:
       - name: Run Vitest - Component Testing
         if: always()
         working-directory: ./frontend
-        run: sudo yarn test
+        run: sudo yarn test --silent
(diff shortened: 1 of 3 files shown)


andrewtavis (approved):
Thanks so much for this, @cquinn540! Really appreciate the quick fix :)

Issue #1132, opened 2025-02-27
Expand Views testing for backend


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

In #1121 and #1125 we added in a coverage report to the backend PR workflow. One thing that became apparent from that is that our testing for Views in the backend is quite low. Here's the current output for the View files in the backend:

```
backend/authentication/views.py                       90     13    86%   84-97, 194-204
backend/communities/groups/views.py                   92     65    29%   29-31, 34-39, 42-50, 53-74, 77-101, 104-126, 136-169
backend/communities/organizations/views.py            97     68    30%   36-38, 41-47, 50-61, 64-87, 90-113, 116-142, 152-185
backend/communities/views.py                           8      0   100%
backend/content/views.py                             153    113    26%   27-35, 41-52, 55-65, 71-81, 84-95, 101-110, 119-126, 132-155, 158-167, 170-181, 184-195, 198-207, 220-228, 234-245, 248-258, 261-272, 275-286, 289-299
backend/events/views.py                               96     69    28%   28-30, 33-38, 41-49, 52-73, 78-98, 103-122, 132-165
```

I'm expecting that the tests to expand the coverage of these files will be fairly similar, hence putting it all into a single issue. We can also break this up into file based PRs if that would be easier for the contributing :)

### Contribution

Happy to discuss the implementation here and review a PR once it's opened! I can also pick this up eventually if need be 😊

sh-ran:
Just wanted to ask if I had to write the remaining test cases in order to get a 100% coverage for each of the file.

andrewtavis:
Ideally 100%, but if you want to expand out the tests a bit with an initial PR that would be totally fine, @sh-ran! Feel free to send something along for your first try and we'll go from there :)

mattburnett-repo:
Hi there

Just wanted to mention that there already some  model, serializer and view tests for the content app (image upload and social media links) in the backend/content/tests folder. More test coverage is always better, though, so please feel free to add to the coverage for these things if you wish!

sh-ran:
@mattburnett-repo I had a doubt with this.

![Image](https://github.com/user-attachments/assets/c7481603-5566-4e0a-b5bd-8b42ffe7b489)
[...]

PR #1168, opened 2025-03-19
Missing test for deleting authenticated users

sh-ran:
### Contributor checklist

---

### Related issue
- #1132

### DIFF
--- a/backend/authentication/tests.py
+++ b/backend/authentication/tests.py
@@ -189,8 +189,7 @@ def test_sign_in(client: Client) -> None:
     # 4. User does not exists and tries to sign in.
     response = client.post(
         path="/v1/auth/sign_in/",
-        data={"email": "unknown_user@example.com",
-              "password": "Password@123!?"},
+        data={"email": "unknown_user@example.com", "password": "Password@123!?"},
     )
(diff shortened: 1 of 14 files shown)


sh-ran:
I will remove the if block and submit again ASAP. Thank you for the suggestion.

sh-ran:
So I tried removing the if block and when I ran the tests again, the test fails throwing an assertion error 401. In order for the user to be deleted, the user has be an authenticated user. And the only way I can think of ensuring that only authenticated users can delete is by using that IF block. Unless I am missing something. 

@mattburnett-repo

mattburnett-repo:
ah! That makes sense. Let's keep things simple and put the 'if' statement back in. 

Thanks for responding so quickly!

andrewtavis:
Minor change to left justify the function docstrings and break them onto a new line always. I find it makes it easier to read these things when they're all structured in a similar way 😊

andrewtavis (approved):
Thanks for the great first contribution here, @sh-ran! Really looking forward to further collaborations with you 😊 And also thanks for the review support, @mattburnett-repo! 🚀

PR #1192, opened 2025-03-27
WIP refactor: use composition pattern for FormTextInput

cquinn540:
### Contributor checklist

---

### Description

This refactors the FormTextInput from a component that handles both text and password inputs with a lot of imperative if statements to using a composition pattern instead.

I have also deliberately created a duplicate of the sign-in page so I could test the new pattern against the old pattern side by side.  I will remove the duplicates in my next PR.

I'm running into some weird Nuxt crashes on my local dev where Nuxt can't seem to find the new files.  No idea why atm.  I'm curious if the tests will pass in CI

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/frontend/components/form/text/FormPasswordInput.vue
+++ b/frontend/components/form/text/FormPasswordInput.vue
@@ -1,0 +1,32 @@
+<!-- SPDX-License-Identifier: AGPL-3.0-or-later -->
+<template>
+  <FormTextInput2 :hasError="hasError" :type="isPassword ? 'password' : 'text'">
+    <template #icons>
+      <slot name="icons"></slot>
+      <button @click="changeInputType">
+        <Icon
+          :name="isPassword ? IconMap.VISIBLE : IconMap.HIDDEN"
+          size="1.4em"
+        />
+      </button>
+    </template>
+  </FormTextInput2>
+</template>
+
+<script setup lang="ts">
+import { IconMap } from "~/types/icon-map";
+
+export interface Props {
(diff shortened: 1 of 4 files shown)


andrewtavis:
Quick note @cquinn540 that we need the directory structure of the component name to be integrated into the component name, so if it's in `components/form/text` then it needs to be `FormTextEtc` :)

cquinn540:
Okay, figured out why the tests were erroring so this PR is ready

andrewtavis (approved):
All's looking great here, @cquinn540! Is the plan now to move on to sign-in and then switch over the 2-series files to be the main ones while getting rid of the duplicates? Looking forward to the next PR!

andrewtavis:
Also FYI, if we could try to get the changes here done over the weekend that would be great so we don't have the duplicate files when I do the layout work :)

andrewtavis:
Note that the double file here is also causing issues with the i18n keys. I did update them so others don't deal with them, but let's remember that we'll need to switch them back :) This made it clear though that the rule we had for the i18n check running only when the i18n files were changed wasn't enough, so now it's running on all frontend changes 🌐

cquinn540:
Yeah next PR will be to get rid of the duplicate sign-up page and all instances of the old TextFormInput

andrewtavis:
Nice, ya that'll be great :) I knew that was the plan, but also wanted to write in to remind myself to fix it when the new one is up 😊

PR #1204, opened 2025-04-04
Doc: Added documentation in backend events folder

rohityadav2604:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

- This pull request contains documentation of the events folder in the backend. I have tried to keep the documentation concise and clean if there is anything more needed I would be happy to incorporate the changes. 

### Changes

- Backend/events folder

### Testing

- I have done the initial backend testing for by following this [doc](https://github.com/activist-org/activist/blob/main/CONTRIBUTING.md#testing-). Attached are the screenshots of testing
- Screenshot 2025-04-04 at 8 48 46 PM
- Screenshot 2025-04-04 at 8 49 10 PM
- Screenshot 2025-04-04 at 8 50 11 PM

### Related issue

- [Issue 1198](https://github.com/activist-org/activist/issues/1198)

Issue #1179, opened 2025-03-23
API endpoint events


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

This issue follows the work done in #1077 and the corresponding PR in #1159. The goal is that we'll similarly achieve greater flexibility for events via a transition to an `APIView` and `Serializer`, replacing the current `ViewSet` and `ModelSerializer`. In order to make the api implement that more flexible we want to move to a `APIView` and the `serializers.Serialer` instead of the `ViewSet` and `ModelSerializer`.

APIViews are not handling List and Detail Views, therefore we are going to split the views into:

```py
class EventListAPIView(APIView):
      pass
        
class EventDetailAPIView(APIView):
      pass
```

Note that the work in #1159 can generally be copied for this issue with the work for organizations there being translated where needed for events.

### Contribution

Happy to support on this and potentially get to it myself. Can also review as needed if someone wants to open a PR! 😊

andrewtavis:
CC @to-sta for the discussion here 😊

rishabmamgai:
Hey @andrewtavis, is someone already working on this? I think I can pick this one.

andrewtavis:
Assigned, @rishabmamgai! Maybe take a look at the conversation in https://github.com/activist-org/activist/issues/1178 for some questions that have come up for the person working on the same for groups, in case that would help :) Thanks for picking this up! 😊

rishabmamgai:
So I am facing few issues while testing out my changes with `EventSerializer -> validate()` -

1. Calls below method
https://github.com/activist-org/activist/blob/b9b19a5473d4233fcac286764688015bbb683baa/backend/utils/utils.py#L9
When I try to test the `EventListAPIView -> post()` to create an object, `creation_date` field doesn't exists in data since it'll only be assigned a value/created when object is created because of `auto_now_add=True`.
Also I believe `creation_date` field ISO validation is not needed, since this is an auto field which will be immutable.

2. Performs the below check -
https://github.com/activist-org/activist/blob/b9b19a5473d4233fcac286764688015bbb683baa/backend/events/serializers.py#L58
`data["start_time"]` and `data["end_time"]` are somehow not str when I sent a json payload. Although I did `str(data["start_time"])` for now.

Any suggestions for this?

[...]

Issue #1178, opened 2025-03-23
API endpoint groups


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

This issue follows the work done in #1077 and the corresponding PR in #1159. The goal is that we'll similarly achieve greater flexibility for groups via a transition to an `APIView` and `Serializer`, replacing the current `ViewSet` and `ModelSerializer`. In order to make the api implement that more flexible we want to move to a `APIView` and the `serializers.Serialer` instead of the `ViewSet` and `ModelSerializer`.

APIViews are not handling List and Detail Views, therefore we are going to split the views into:

```py
class GroupListAPIView(APIView):
      pass
        
class GroupDetailAPIView(APIView):
      pass
```

Note that the work in #1159 can generally be copied for this issue with the work for organizations there being translated where needed for groups.

### Contribution

Happy to support on this and potentially get to it myself. Can also review as needed if someone wants to open a PR! 😊

andrewtavis:
CC @to-sta for the discussion here 👋

IWang20:
Hi, can I be assigned to this issue?

andrewtavis:
Thanks for your interest in working on this, @IWang20! Please let us know if you need any support :)

IWang20:
I'm writing tests similar to #1159, and I'm only running the backend cause the entire project takes a while to start on my windows machine. 
However, running only the backend using the instructions from CONTRIBUTING.md doesn't have the docker_backend container, which is needed to run the pytests. 

Will I have to run the entire project just to run pytests? 
Thanks

to-sta:
> I'm writing tests similar to #1159, and I'm only running the backend cause the entire project takes a while to start on my windows machine. 
> However, running only the backend using the instructions from CONTRIBUTING.md doesn't have the docker_backend container, which is needed to run the pytests. 
> 
> Will I have to run the entire project just to run pytests? 
> Thanks

It is enough to run the backend container (it will automatically spin up the database container as well, because the backend depends on the db) and then you can run pytest within the container. 

`docker compose up backend --build`

IWang20:
Hi, running into an issue where calling 
`response = client.get("/v1/communities/groups/")` 
[...]

PR #1209, opened 2025-04-06
Groups APIView

IWang20:
### Contributor checklist

---

### Description
This issue follows the work done in #1077(PR #1159) and #1179(PR #1204). It pretty much follows the same implementation and testing structure as #1077, by using an `APIView` and `Serializer` instead of `View` and `ModelSerializer`. The original class was split into GroupList(GET, POST) and GroupDetail(GET, PUT, DELETE) and includes a test_api file to ensure that each request was functioning properly. 

This issue also includes a fix for the custom paginator class that was done in #1077, to address the query parameter auto transformation(from camelCase to snake_case) due to middleware `django-camelcase`. 

- ISSUE #1178


backend/communities/groups/tests/test_api.py
```diff
@@ -0,0 +1,198 @@
...
+        call_command("loaddata", str(fixture_path), verbosity=2)
+
+
+@pytest.fixture
+def new_user() -> UserDict:
+    return create_user("Activist@123!?")
+
+
+@pytest.fixture
+def created_by_user() -> UserModel:
```
cquinn540, line 62: I think this function makes more sense to be named `create_user`.

The existing `create_user` function could be named `create_user_and_password`

cquinn540: Or if the convention here is to use past tense for fixtures, `created_user` would work.

`created_by` implies that this function creates some object that was created by a user, rather than creating the actual user itself

backend/communities/groups/tests/test_api.py
```diff
@@ -0,0 +1,198 @@
...
+from rest_framework.test import APIClient
+
+from authentication.factories import UserFactory
+from authentication.models import UserModel
+from communities.groups.factories import GroupFactory
+from communities.groups.models import Group
+from content.factories import EntityLocationFactory
+
+
+class UserDict(TypedDict):
```
cquinn540, line 17: This section has a lot of fixtures and functions that create or login in a user and only have small variations in the arguments and return types.

I think a simpler way to do this is with a class:

```python
class UserLogin:
    def __init__(self, plaintext_password="Creator@!123?"):
        self.user = UserFactory.create(plaintext_password=plaintext_password, is_confirmed=True)
        self.password = plaintext_password
```

Since the class constructor has a default plaintext password, you shouldn't need 3 different fixtures that also provide a default password.

backend/communities/groups/tests/test_api.py
```diff
@@ -0,0 +1,198 @@
...
+
+@pytest.fixture
+def created_by_user() -> UserModel:
+    """
+    Create a user and return the user object.
+    """
+    return create_user("Creator@123!?")["user"]
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
cquinn540: With the User Login class both could just be
```python
def logged_in_user(new_user) -> dict:
    return login_user(new_user)

def logged_in_created_by_user(created_by_user) -> dict:
    return login_user(created_by_user)
```

backend/communities/groups/tests/test_api.py
```diff
@@ -0,0 +1,198 @@
...
+    client = APIClient()
+    response = client.post(
+        "/v1/auth/sign_in/",
+        {
+            "username": user_data["user"].username,
+            "password": user_data["plaintext_password"],
+        },
+    )
+    assert response.status_code == 200
+    return {"user": user_data["user"], "token": response.data["token"]}
```
cquinn540, line 43: I think it would also be helpful for this return type to be a class

```python
class UserLogin:
    def __init__(user: UserModel, token: str):
        self.user = user
        self.token = token
```

backend/communities/groups/tests/test_api.py
```diff
@@ -0,0 +1,198 @@
...
+    # GET request
+
+    1. Create a new group and verify it exists in the database.
+    2. Test the detail view endpoint for the group.
+    3. Verify the response status code is 200 (OK).
+    4. Ensure the response data matches the group data.
+
+    # PUT request
+
+    1. Check if groups can be edited without the proper credentials
```
cquinn540, line 160: Really good security test here!

cquinn540: Also check if the group can be deleted without proper credentials

backend/communities/groups/views.py
```diff
@@ -15,117 +19,154 @@
...
+    serializer_class = GroupSerializer
 
-        else:
+    @extend_schema(
+        responses={
+            200: GroupSerializer,
+            400: OpenApiResponse(response={"error": "Group ID is required"}),
+            404: OpenApiResponse(response={"error": "Failed to retrieve the Group"}),
+        }
+    )
```
cquinn540, line 77: I really like the use of extend_schema to make it easier to see all the expected responses

backend/communities/groups/views.py
```diff
@@ -15,117 +19,154 @@
...
+        except Group.DoesNotExist:
             return Response(
                 {"error": "Group not found"}, status=status.HTTP_404_NOT_FOUND
             )
 
         if request.user != group.created_by and not request.user.is_staff:
             return Response(
-                {"error": "You are not authorized to delete this group"},
-                status.HTTP_401_UNAUTHORIZED,
+                {"error": "You are not authorized to delete this Group"},
```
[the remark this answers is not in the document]
IWang20: Would it be ok to have `request.user != group.created_by` stay in the same conditional as `not request.user.is_staff`?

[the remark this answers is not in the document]
cquinn540: Sorry, was busy the last couple days.

In this case I don't think so.  `request.user.is_staff` should be the first conditional checked before you even query the database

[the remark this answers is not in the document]
cquinn540: It's best to think of them as two different errors.  If `request.user != group.created_by` means the user can't delete that specific group.

If `not request.user.is_staff` that means the user is not allowed to delete any group at all.

to-sta:
@IWang20 the `authentication_classes` and `permission_classes` are missing on the `GroupDetailAPIView`. To be fair the authentication_classes defaults to `TokenAuthentication`, because we set that in `core/settings.py`. But we haven't cleaned that up on all views yet, therefore I would say let's just add it and we work on that later on in another PR.

```py
class GroupDetailAPIView(GenericAPIView[Group]):
    queryset = Group.objects.all()
    serializer_class = GroupSerializer
    autentication_classes = (TokenAuthentication,)
    permissions_classes= (IsAuthenticatedOrReadOnly,)
```

@andrewtavis within the [Prototypes](https://www.figma.com/design/I9McFfaLu1RiiWp5IP3YjE/activist_public_designs?node-id=4482-32273&t=CpLTUsz4dOOPYgBi-4) the Organization is required but currently within the modles.py have set `null=True` on the `org` field.

If it is required I would recommend to create a seperate serializer for the `POST` method of `GroupAPIView`, similar to what we did in `EventListAPIView`.

andrewtavis:
All groups are just subsets of organizations, so an org is required, @to-sta :) I'll make the above changes now 😊

andrewtavis:
Tests will fail, but we're really close :)

andrewtavis:
Tests have been matched across all entity API views and are passing ✅ Thanks for the directions on how to finish this up, @to-sta! Can you review where it's at? :)

andrewtavis (approved):
Worked through the final needed changes in the sync just now with @to-sta :) Thanks so much for the great contribution here, @IWang20! Really a great first contribution to the project 😊 Looking forward to future collaborations with you!

Issue #879, opened 2024-05-19
Backend rate limit tests are failing


to-sta:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Our rate limit tests are currently failing, but the rate limiting functionality is working correctly. I have manually tested it using SwaggerUI.

to-sta:
cc @josueemartinezz

thesarfo:
is this being worked on? If not I could get this done

to-sta:
Would be great if you can tackle this @thesarfo

andrewtavis:
Thanks for the assignment, @to-sta, and thanks for offering to help out @thesarfo :)

to-sta:
@prajguru54 thanks for checking :). I removed the tests for a while in order to avoid failing gh actions. Forgot to add a comment here. The last change to the test was made in the commit ` 591bd07c243c279573cbf5b9eeb51e4ca5be89b0`. But I think it would be better to start from scratch 😄. 

We want to check the rate limits for signed in users and anom users.

andrewtavis:
@prajguru54 or @thesarfo, would either of you want to reimplement them?

thesarfo:
yes, I would want to do that. but @prajguru54 can take it if he feels up to it

andrewtavis:
You're assigned, @thesarfo, so by all means feel free to send something along :) Let us know if you'd like some ideas or support!

prajguru54:
@andrewtavis , I will be glad to take this. Already started on it yesterday. @thesarfo  Have you also started as well??

andrewtavis:
Hey all 👋 Checking in post a busy last month on my end. Hope all are well :) We're still looking for the tests that were removed to be re-implemented. @thesarfo or @prajguru54: Do either of you have interest?

thesarfo:
@andrewtavis I’ve been going through the codebase to locate the rate-limiting functionality but haven’t been able to pin it down yet. I’m keen to write the tests for it, could you point me in the right direction?

andrewtavis:
CC @to-sta who might have a bit better of an opinion on how to implement this? :)

to-sta:
[...]

PR #1229, opened 2025-04-25
Add rate limit tests

knika01:
### Contributor checklist

---

### Related issue

https://github.com/activist-org/activist/issues/879
- #ISSUE_NUMBER


backend/media/images/fa8de9da-6887-4a1d-b011-73b0c7f5454a.jpg
to-sta, line 1: I assume that slipped through 😄

andrewtavis:
Hey @knika01 👋 Could you fix the unneeded image as this was supposed to be git ignored? 

Aside from this, we have license headers that need to be added to the files and as seen [here](https://github.com/activist-org/activist/actions/runs/14655992007/job/41186545282?pr=1229) and there are some backend test failures as seen [here](https://github.com/activist-org/activist/actions/runs/14655991992/job/41186545295?pr=1229). Seems like for the backend tests there are some import errors?

We'd be happy to support more if there are questions! 😊

backend/tests/test_anon_throttle_swagger.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+from django.core.cache import cache
+from django.test import override_settings
+from rest_framework import status
+from rest_framework.test import APIClient
+
+
+@pytest.mark.django_db
```
jamshaidali, line 8: @knika01  I just wanted to highlight one thing here. Since we have set the throttle settings in settings.py, those settings will be used for the tests too. It's good that we have used override here in this particular test, but if we reach the endpoint limit during other tests, that could cause issues. We should have set the throttle settings for the tests in the conftest file. Let me know what you think.

jamshaidali (changes requested):

backend/tests/test_anon_throttle_swagger.py
andrewtavis: So you mean a different throttle level on a per app basis, @jamshaidali? I'd be ok with that :)

jamshaidali: @andrewtavis Yes, that's what I am trying to say. We can also set the throttle level for the tests so that our actual throttle settings do not break the tests when they hit the throttle limit as the number of views or APIs increases in the future.

andrewtavis: I'd say something like setting something for the tests makes sense for now 😊 Maybe you could send that change along after we merge this in, @jamshaidali? :)

jamshaidali: Sounds good, @andrewtavis. I will test the change to add throttle settings for tests and raise its PR once this PR is merged. I'll make sure to keep an eye on it to see if this one gets merged or not.

andrewtavis:
Hey @knika01 👋 Thanks much for the initial contribution here :) The branch is a bit out of date at this point and we wanted to finalize this, so the work for #879 was picked up in #1271. Really appreciate the effort here, and hope you contribute again 😊

Issue #1199, opened 2025-03-31
Investigate and potentially adding mypy django-stubs plugin


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

This issue would look into the benefits of adding in [typeddjango/django-stubs](https://github.com/typeddjango/django-stubs) to the mypy plugins used in activist. Once a report on the overall positives of adding it in has been finished, a decision would be made on whether it should be added. From there a PR could be raised to add the plugin and fix all the errors and warnings that it brings to our attention :)

viditi08:
Hi! I'd be happy to help with this issue.

andrewtavis:
Thanks for your willingness to help, @viditi08! Please let us know if you need some support. It'd be great to get the basic changes and benefits proposed in the issue after the investigation, and then we can move towards a PR :) 

Assigning now 😊

viditi08:
Thanks for assigning it to me! I’ll investigate the issue and propose the necessary changes and benefits. I'll keep you updated, and we can move forward with the PR once everything's ready! 😊

andrewtavis:
Sounds wonderful, @viditi08! 😊

andrewtavis:
All sounds good, @viditi08 :) Feel free to start working on the first PR 😊

viditi08:
Hey , When i comment out follow_imports = "skip"  , I am getting type erroes for factory , 
Do you want me to create custom  stub file with minimal type annotations to fix those type errors ? or just simply  ignore the errors ?

andrewtavis:
For now let's ignore them, but let's keep this in mind during the review :) Thanks for checking!

viditi08:
Thanks for the feedback! I've raised a PR request.

andrewtavis:
Perfect! Looking forward to the review :) :)

andrewtavis:
Sorry for the wait on bringing in the PR, @viditi08 :) What are you thinking for the next steps? Do you want to raise PRs on a file by file basis and then we can change `ignore_errors` to `false` once that's all done?

viditi08:
Yes, I can work on a file-by-file basis and remove those errors.

andrewtavis:
Fantastic, @viditi08 :) Great to have your support here!

andrewtavis:
[...]

PR #1239, opened 2025-04-29
Fix type errors, remove unused imports, apply Ruff formatting

qabilityp:
### Contributor checklist

---

### Related issue

- #1199

Summary of Changes:

Fixed Type Errors:

Resolved the issue with unsupported operand types in backend/events/serializers.py on line 127. Previously, datetime and None were being compared with > and < operators, causing errors. Added proper type checks to handle this.

Function Return Type Annotations:

Annotated the missing return types for functions in backend/events/views.py and backend/communities/groups/views.py (e.g., get_permissions, get_serializer_class).

Permission Class Fix:

Added new class attributes permission_classes_default and permission_classes_post to handle different permissions for different HTTP methods (GET and POST).

The get_permissions method was updated to apply the correct permission classes based on the request method.

Serializer Return Type Fix:

Updated the return type of get_serializer_class in backend/events/views.py to return either EventSerializer or EventPOSTSerializer, ensuring compatibility with the superclass GenericAPIView.

Corrected Serializer Callables:

Fixed issues where serializers like EventSerializer and GroupSerializer were mistakenly being used as callable objects (this caused errors on line 76 in backend/events/views.py and line 79 in backend/communities/groups/views.py).

Import Fix for OpenApiTypes:

The import for OpenApiTypes from drf_spectacular.utils was corrected to import from drf_spectacular.types.


andrewtavis:
Thanks for the fixes here, @qabilityp! Really appreciate the speed that you're getting these changes in 😊

Would you be able to take a look at this when you have time, @to-sta?

backend/communities/groups/views.py
```diff
@@ -36,18 +39,19 @@ class GroupAPIView(GenericAPIView[Group]):
     serializer_class = GroupSerializer
     pagination_class = CustomPagination
     authentication_classes = (TokenAuthentication,)
-    permission_classes = (IsAuthenticatedOrReadOnly,)
+    permission_classes_default = (IsAuthenticatedOrReadOnly,)
```
to-sta, line 42: We should stick to `permission_classes`, because the `drf-spectacular` package is looking for it.

The `permission_classes` is the default, and if there is a `get_permissions `method within the APIView, then that is the place to customize it and override permissions explicitly.

backend/communities/groups/views.py
```diff
@@ -36,18 +39,19 @@ class GroupAPIView(GenericAPIView[Group]):
...
+            self.permission_classes = self.permission_classes_post
 
         else:
-            self.permission_classes = (IsAuthenticatedOrReadOnly,)
+            self.permission_classes = self.permission_classes_default
 
         return super().get_permissions()
 
-    def get_serializer_class(self) -> GroupSerializer | GroupPOSTSerializer:
+    def get_serializer_class(self) -> Type[Union[GroupSerializer, GroupPOSTSerializer]]:
```
to-sta, line 54: Good catch! We should definitly use `Type` to indicate that we are returning the whole class and not an instance of the class.

I would still stick to the Union operator | because we don't have to support older python versions.

```py
def get_serializer_class(self) -> Type[GroupSerializer | GroupPOSTSerializer]:
      pass
```

backend/communities/groups/views.py
```diff
@@ -76,7 +80,7 @@ def get(self, request: Request) -> Response:
     )
     def post(self, request: Request) -> Response:
         serializer_class = self.get_serializer_class()
-        serializer: GroupPOSTSerializer = serializer_class(data=request.data)
+        serializer = serializer_class(data=request.data)
```
to-sta, line 83: I also try to avoid inline type hints at all cost but we are not getting intellisense here without it. Has that changed?

andrewtavis: I can't think of anything that would have removed intellisense, but there have been many edits to the pyproject.toml recently, so maybe something is conflicting? 🤔

PR #1266, opened 2025-05-16
fix: Issue in report title is not showing

Abhi-Bohora:
When a bot creates an issue mentioning what are the latest dependencies version that our project will need to be updated, the date in the issue title title was showing `$(date '+%B %d, %Y'` instead of real date, so this PR aims to fix that.

### Contributor checklist

---

### Related issue

- [#1262](https://github.com/activist-org/activist/issues/1262)


andrewtavis (approved):
Thanks for the quick fix here, @Abhi-Bohora! So great to have your support on the dependency update process 😊 Once the mypy work is done we should be good to go :) :)

Issue #1289, opened 2025-05-26
Add the ability to create FAQ entries to all FAQ pages


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

In #1286 we finalized editing FAQ entries via the new modal structure. The final thing that's needed now is the ability to add a new FAQ entry to to an organization, group or event via the FAQ page on each of these entities 😊 This would be done via the CTA button in the top left of each of the pages.
 
- Note: This issue finalizes the work for #818 and #815 
- Note: The WIP banners should be removed from all FAQ pages then :)

### Contribution

Happy to support as needed here and review a PR when one's up! 🚀

enrouxlibre:
Hey! I will look into it! I'll also see about making the FAQ match the correct language.

andrewtavis:
Closed by #1304 😊 Thanks for the great work here @enrouxlibre, @to-sta, @sh-ran and @nicki182!

PR #1294, opened 2025-05-27
[WIP] Refactor FAQ & Add FAQ functionality

enrouxlibre:
### Contributor checklist

- [X] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

So when I implemented the edit FAQ functionality, I simply copied what was done for Social Links. This is my attempt at coming up with a more "centralized" version so that we don't have the code in triplicate, since I don't see a reason why FAQ logic would change depending on whether it's for a Group, an Organization, or an Event. This will make editing the code faster in the future, especially if FAQ was added to other types of entities as well.

"Centralized" the logic to edit FAQ in the backend (new ViewSet) and the frontend (added a new store).

This is my first commit, in order to get early feedback. I haven't deleted the old code yet, although with these changes, the app isn't using them anymore.

I also want to look into removing the specific Faq types in the back end (GroupFaq, OrganizationFaq, EventFaq) in order to consolidate the code for tests/factories as well.

---

This also adds the functionality to add FAQs.

### Related issue

- #1289


enrouxlibre:
Following our conversation and since we want to split FAQs into separate tables, I think it's better to keep separate end points.

Issue #1307, opened 2025-06-04
Add functions to export .ics file with event information


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

This issue is to allow for the export of a `.ics` file about events in the following places:

- `/events` in the three dot menu of each event (export just the event)
- On an event page (export just the event)
- On the `organization/id/events page (export all organization events)
- On the organization/id/groups/id/events page (export all group events)

This allows those who are interested in events to easily save it to their calendar. The code that would need to have this functionality added to it was added in https://github.com/activist-org/activist/pull/1276.

### Contribution

There's a question of whether we need to involve the backend here in the first version of this functionality. Maybe it'd be enough to just export the `.ics` via a frontend dependency, but let's discuss before work is started and after the decision of where the `.ics` file is served from :)

andrewtavis:
@naveen28204280, all done making the new issue here :) Let me know what your thinking is as far as where this calendar file should be served from 😊

andrewtavis:
And also feel free to write in so I can assign :)

naveensrinivas282:
I'd like to take up this issue - please assign me
As for whether or not it should be in the backend - I personally think making backend endpoints would be better as it ensures better security and prevents the file from being manipulated and also makes syncing with Google Calendar much easier. I'll look into both options in any case and try to send the PR asap.

andrewtavis:
Thanks so much for taking this on and exploring the possibilities, @naveen28204280! All sounds great :) Please let us know if you need support 😊

andrewtavis:
Closed by #1313 🚀 Appreciate the hard work that you put into this, @naveen28204280! We've been super busy. Apologies that it took us a bit to get the PR in :)

PR #1276, opened 2025-05-19
Issue 716

Nishthajain7:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

This pull request adds a feature that allows exporting all events of an organization. I also adds a "Subscribe to event" button to the tooltip and about section of an event

### Related issue

- #716

PR #1313, opened 2025-06-09
Added export event as .ics file

naveensrinivas282:
### Contributor checklist

- [x] This pull request is on a [separate branch](g) and not the main branch

---

### Description

This feature is to allow for the export of a .ics file about events in the following places:

- /events in the three dot menu of each event (export just the event)
- On an event page (export just the event)
- On the `organization/id/events page (export all organization events)
- On the organization/id/groups/id/events page (export all group events)

This allows those who are interested in events to easily save it to their calendar. The code that would need to have this functionality added to it was added in https://github.com/activist-org/activist/pull/1276.

### Related issue

https://github.com/activist-org/activist/issues/1307

- #1307

### DIFF
--- a/backend/events/views.py
+++ b/backend/events/views.py
@@ -4,13 +4,14 @@ API views for event management.
 """
 
 import logging
+import re
 from typing import Any, Sequence, Type
 from uuid import UUID
 
 from django.db.utils import IntegrityError, OperationalError
 from django.http import HttpResponse
 from drf_spectacular.utils import OpenApiParameter, OpenApiResponse, extend_schema
-from icalendar import Calendar
+from icalendar import Calendar  # type: ignore
 from icalendar import Event as ICalEvent
 from rest_framework import status, viewsets
 from rest_framework.authentication import TokenAuthentication
@@ -494,6 +495,7 @@ class EventCalenderAPIView(APIView):
 
         try:
             event = self.queryset.get(id=event_id)
+
         except Event.DoesNotExist:
             return Response(status=status.HTTP_404_NOT_FOUND)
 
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -10,4 +10,7 @@
     ".vscode/tailwind.json"
   ],
   "vitest.rootConfig": "./frontend/vitest.config.mts",
+  "cSpell.words": [
```
andrewtavis: You can also add such words to your user settings, @naveen28204280 :) No need to add them to the VSCode settings as maybe others don't want this or use the extension :)

andrewtavis:
@naveen28204280, would you be able to look into the mypy errors in the [backend workflows](https://github.com/activist-org/activist/actions/runs/15539438431/job/43757458254?pr=1313)? And pinging @to-sta here: There's lots of backend tests that are failing for a relatively small change, which is really surprising 🤔 Do you have an idea of why all these tests might be failing?

backend/events/views.py
```diff
@@ -4,12 +4,15 @@
 """
 
 import json
-from typing import Any, Dict, List, Sequence, Type
+from typing import Any, Dict, List, Optional, Sequence, Type
 from uuid import UUID
 
+from backend import communities as c
```
to-sta: There are a lot of `No module named 'backend` errors within the backend workflow, caused by this import. The backend itself is not a python package.

andrewtavis: @naveen28204280, would you be able to fix the above import and then we can continue from here? Your feedback on the methods to export these `.ics` files would also be much appreciated, @to-sta!

to-sta: I haven't worked with iCal before, so I can't provide meaningful feedback without digging into it.

andrewtavis: No stress then :) We can figure this out a bit later in say a sync or something like this :)

naveensrinivas282: I'll look into the import today itself and correct it

.vscode/settings.json
naveensrinivas282: Oh sorry, I didn't see that. I'll be more careful from now on. Thanks for pointing it out :)

andrewtavis (approved):
Edits from @to-sta have us at an APIView now for the .ics export functionality, which as of now just works for a single event. I'll make another issue for getting it set up for organizations and groups :) Thanks for the great first contribution, @naveen28204280! 😊

PR #1328, opened 2025-06-22
1291 selector view events

nicki182:
### Contributor checklist

---

### Related issue

- 1292


andrewtavis:
All's looking really great here, @nicki182! One thing is that it looks like the left sidebar is now on top of another left sidebar. The same is the case for menu options and for the filter component on `/events` (see below):

Screenshot 2025-06-24 at 00 57 08

See in the above that there's a layer around the outside. Could we get a fix for that, and then I think we'll be good to go here! 😊

nicki182:
fixed, I didn't see it because I was using dark mode

andrewtavis:
Summary of what came along in 9e27a6d:

- Hope it's ok that I moved the filter sidebar work for events into `sidebar/left/filter/FilterEvents`
    - Felt this was easier to understand than `ContentEvents`
- We now use `SidebarLeftFilter` in the same way as `SidebarLeftContent`
    - I left the code that determines the filters per page in the `SidebarLeft` code so that we can reference it while we build out the filters for each page, but this should also be removed eventually
- Styling for the view selector so it matches the designs a bit better
- Switched all instances of `*_FILTER` in `SidebarType` to `*_PAGE`
- We have the orange filter icon when the sidebar is closed again

Hope the above sounds good! Happy to discuss any of it :)

PR #1331, opened 2025-06-23
backend dependency updates

sh-ran:
### Contributor checklist

---

### Description

Backend dependencies have been updated to the most recent releases.

### Related issue

- #1262


sh-ran:
Hello! @andrewtavis 

So I noticed that out github backend workflow uses postgres version 13 while our `docker-compose.yml` has version 15 installed.

Should I update our workflow or just leave it as is and keep the django version as same?

andrewtavis:
Let's definitely sync the versions and update the workflow. Thanks, @sh-ran!

andrewtavis:
And for the pytest errors, let's work to update as much as we can, but if you see that one is causing many tests to fail, then feel free to just do a minor update or something like that. We can focus in on updating tests in a future PR if you'd prefer. Just bringing this up so you know you have options here and don't need to fix everything in one go 😊

sh-ran:
There are no errors within the tests. The errors caused was coz django couldnt connect with postgres coz django needs atleast v14. Updating our workflow fixed it 😄 

PS: some of the packages have been updated to newer versions than on the report but everything is working good.

PR #1332, opened 2025-06-26
Fix/mypy content models

hericlibong:
### Contributor checklist

---

### Description

This pull request disables `ignore_errors = true` in the `pyproject.toml` overrides and fixes all `mypy` errors raised across models, views, and serializers in the following modules:

- `authentication`
- `content`
- `events`
- `communities`

### Types of errors fixed:

- Missing type annotations in views (`no-untyped-def`)
- Missing annotations for ManyToMany fields (`var-annotated`)
- Incompatible permission assignment (`assignment`)
- Unsafe date comparisons (`operator`)
- Wrong return types (`override`, `return-value`)
- Bad lookup on `filter(id=...)` (`misc`)
- Undeclared `OpenApiTypes` import (`attr-defined`)

All corrections are isolated, non-breaking, and focused on aligning the project with strict `mypy + django-stubs` typing.

See the file `resume.md` in this branch for a detailed mapping of errors and fixes.

---

### Related issue

- #1199


andrewtavis:
Looks like there are still some mypy and migration errors here, @hericlibong :) Could we ask that you check them out and send along further fixes?

hericlibong:
Hi, 
All CI checks are green now, migrations and tests pass locally.
Happy to have your feedback when you have time – thanks a lot for your support and review!

andrewtavis:
Thanks for the PR, @hericlibong! We'll work to bring it in in the coming days 😊

@to-sta and @sh-ran: I've cleaned this up as much as I could and tests are passing, but I did want to make sure that you two weight in on the changes to permissions class assignments as well as the use of `isinstance` for the dates and casting certain variables. Let us know what you're thinking here!

hericlibong:
Thanks for the detailed review and improvements! I checked your changes and everything looks great to me. Happy to see this contribution moving forward.

backend/communities/groups/views.py
```diff
@@ -50,18 +51,21 @@ class GroupAPIView(GenericAPIView[Group]):
...
+        Instantiates and returns the list of permissions that this view requires.
+        """
         if self.request.method in SAFE_METHODS:
             self.permission_classes = (IsAuthenticatedOrReadOnly,)
 
         else:
             self.permission_classes = (IsAuthenticated,)
 
-        return super().get_permissions()
+        return cast(List[BasePermission], super().get_permissions())
```
to-sta: I do not really understand this part here. Could you explain what this does @hericlibong?

hericlibong: This line is only for static analysis: cast tells mypy to treat the return value of super().get_permissions() as a List[BasePermission]. 
At runtime, cast has no effect—it’s a no-op. This is a common technique to resolve type ambiguities in Django, DRF, when using django-stubs.

andrewtavis: Maybe it would make sense to mypy ignore it as that's more understandable?

to-sta: Thanks for explaining it @hericlibong! 

This the only place where we have to be explicit? What's with the rest of the views?

If possible I think we should ignore this type of error and not just the line. Because this will come up again and again if this is a single line mypy ignore comment.

andrewtavis:
Ignored just the line from [this discussion](https://github.com/activist-org/activist/pull/1332#discussion_r2179980127) as the error in question is just a return value error, so there wasn't really anything fine grained that I could target and always skip. Let me know if there's anything else you'd change here, @to-sta! :)

hericlibong:
Thanks a lot for your follow-up and the details!

I've seen your comments and will take the time to review everything carefully.  
If I notice anything that still needs to be changed, I'll let you know here.

Thanks again for your support and the constructive feedback!

PR #1345, opened 2025-07-06
Refactor module import and pre-commit hook execution to fix mypy error

### LINKED ISSUE #1344: Refactor module imports and pre-commit hook execution for compatibility
### Terms

### Behavior

Running the `mypy-check` hook results in the following error:

Image

This is because the script uses:

```python
from utils import run_shell_command
```
While utils.py actually resides inside the same pre_commit_scripts package.

### Fixes Applied

Updated the import in prettier_eslint_checks.py:

```python
from .utils import run_shell_command
```

Updated the pre-commit hook entry in .pre-commit-config.yaml:

```python
entry: python -m pre_commit_scripts.prettier_eslint_checks
```

This ensures the script is executed as a module, allowing Python to resolve relative imports properly. It also aligns with Python’s recommended module execution patterns.

### Result

After applying the above changes:

- The mypy-check hook passes without error
- The prettier-eslint-checks hook runs correctly on both Windows and Linux

I'd be happy to open a PR with these fixes. Let me know!

ruthvik-mt:
### What this PR does
- Refactored import in prettier_eslint_checks.py to use relative import: from .utils import run_shell_command
- Updated pre-commit hook to run the script as a module using python -m
- Fixes mypy static type checking error caused by incorrect import resolution
- Ensures cross-platform compatibility with tools like pre-commit (Windows/Linux/macOS)

### Related issue
Closes #1344

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,6 +51,6 @@ repos:
       - id: prettier-eslint-checks
         name: run prettier and eslint checks
         files: ^frontend/
-        entry: python ./pre_commit_scripts/prettier_eslint_checks.py
+        entry: python -m pre_commit_scripts.prettier_eslint_checks
         language: python
         pass_filenames: false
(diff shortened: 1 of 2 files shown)


ruthvik-mt:
@andrewtavis! pl review!

andrewtavis (approved):
Thanks so much for the change here, @ruthvik-mt! Really appreciate you checking this out and making the issue 😊

Issue #1379, opened 2025-07-24
Refactor SocialLink Viewsets


sh-ran:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

This issue would rewrite the SocialLinks ViewSets for Groups, Organizations and Event entities. The current code needs to be re-written to adhere to the current way we are writing our views within the backend. It includes some changes on how the new social links are updated instead of deleting and creating a list within the backend.

For #1289 we'll be centralizing the logic for FAQs such that entities will have instantiated models from central FAQ functionality. We want to do the same for `social_links` such that organizations, events and groups would all have social link functionality that inherits from central functionality written in the content app.

andrewtavis:
Thanks for the issue and picking it up, @sh-ran!

MuhammadNizamani:
@sh-ran  are you going to solve this problem

andrewtavis:
Thanks for your interest in the project, @MuhammadNizamani! Would be best if @sh-ran takes this one as we've had lots of discussions on what needs to happen with this issue. Let us know if there's another issue you'd like to pick up though!

MuhammadNizamani:
> Thanks for your interest in the project, [@MuhammadNizamani](https://github.com/MuhammadNizamani)! Would be best if [@sh-ran](https://github.com/sh-ran) takes this one as we've had lots of discussions on what needs to happen with this issue. Let us know if there's another issue you'd like to pick up though!

@andrewtavis I'm an experienced Python backend developer. If you come across any issues that require Python, I’d love to help solve them.
I also have experience with Vue.js, but I prefer working with Python.

andrewtavis:
Would be really really great to work with you, @MuhammadNizamani! Also happy to do a call to onboard you at some point and discuss the project, if you'd like :) Feel free to reach out in our [community Matrix chats](https://matrix.to/#/#activist_community:matrix.org) (you'll see my account to start a direct message in the development and general channels 😊). As far as another issue, ones we could look into are:

- You could assist us finalizing the tests for views and serializes in #1132 (with @sh-ran) and #1133
- Once the authentication method is finalized there will be some work there to refresh it, but let's hold off on that for now
[...]

Issue #1133, opened 2025-02-27
Expand Serializers testing for backend


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

In https://github.com/activist-org/activist/issues/1121 and https://github.com/activist-org/activist/pull/1125 we added in a coverage report to the backend PR workflow. One thing that became apparent from that is that our testing for Serializers in the backend is quite low. Here's the current output for the Serializer files in the backend:

```
backend/authentication/serializers.py                 59      2    97%   111, 114
backend/communities/groups/serializers.py             46      7    85%   59-64, 67-72
backend/communities/organizations/serializers.py      50      7    86%   60-65, 68-73
backend/communities/serializers.py                    10      0   100%
backend/content/serializers.py                        57     20    65%   45-71, 92-106
backend/events/serializers.py                         49     12    76%   58-71, 74-79, 88-90
```

I'm expecting that the tests to expand the coverage of these files will be fairly similar, hence putting it all into a single issue. We can also break this up into file based PRs if that would be easier for the contributing :)

### Contribution

Happy to discuss the implementation here and review a PR once it's opened! I can also pick this up eventually if need be 🚀

narmadha-raghu:
Hi @andrewtavis 

I'd like to work on this issue! I'm looking to start contributing to open source projects, and improving test coverage seems like a great way to get familiar with the codebase. I have experience with backend development and writing tests, though I'm new to this specific project.

I understand this involves writing tests to improve coverage for the serializer files listed in the description. I'd like to start with one file (perhaps backend/authentication/serializers.py since it already has high coverage and might be a gentler introduction) and then continue with the others if that approach works well.

Could you provide guidance on,

The preferred testing approach for serializers in this project ?
Any specific testing patterns I should follow ?
Whether you'd prefer I tackle one file per PR or group them ?

I'll set up the development environment and familiarize myself with the existing tests. Thanks for the opportunity to contribute!

andrewtavis:
Hey @narmadha-raghu 👋 Welcome to the community! Great to have you here 😊

I think you're plan of starting with just `backend/authentication/serializers.py` makes total sense :) Thanks for sharing your thinking on how to tackle this issue!

Answers to your questions:
> The preferred testing approach for serializes in this project?
[...]

PR #1392, opened 2025-07-29
Refactor of views for Groups, Organizations and Events Social Links.

sh-ran:
### Contributor checklist

---

### Description

The models for the respective entity has been updated to have their own social links model rather than a shared social link model like before. The views have been refactored to handle deletion and updating of old links in a cleaner way. Tests include a 403 error to check if the user who is trying to update the links is either the creator of the group or is a staff/admin.

PS: Thank you @enrouxlibre. Your PR helped a ton while I was working on this issue.

### Related issue

- #1379


andrewtavis:
@sh-ran, would you be able to fix the merge conflicts here and then I'll bring this in next? :)

andrewtavis:
I need to look into social links now not updating on the frontend, but maybe this is a similar issue with admins not having the right to do changes or make new ones.

sh-ran:
I mean in the tests unless the user has a `is_staff` attribute the user wont be able to update the links. I look at the code once again. I haven't particularly ran the frontend. Maybe I should have. I will try to fix this ASAP.

sh-ran:
Hey @andrewtavis sorry for getting back to you about this so late. But I think I found why the links wont update in the frontend. So while looking at the logs i saw that while trying to update through the frontend, we are using the `group.id`. But now that we have added a new standalone model for each links, we now should be using something like `groups.socialLinks.id`. I also noticed a comment by you within the file `frontend/stores/group.ts` in the `updateSocialLinks` function saying that we should be using the id from social links rather than the group id. I tried to fix it but was unable to as I would have to change the interfaces for all the social links and uhhh in simple terms I sucked at it. I can work on this with a bit of guidance from you or Nicole. 
Basically we need to update the interfaces for social links to their own entities i.e., group, org and events and pass them in the url.

andrewtavis:
Hey @sh-ran 👋 @nicki182 and I are planning on looking in to this tomorrow sometime :) Might be a bit late on your end, but I'll let you know when we decide on a time!

sh-ran:
Sure let me know the timings @andrewtavis and Thank you!

andrewtavis:
Okkk, @sh-ran 😊 Our idea worked :) I've added create methods into the views and have written the modal such that it saves the original size of the links array, and if there are more links than at the start then the new ones are created :) Is working well. Good enough to bring this in and we can test further in the coming days.

andrewtavis:
~~Reducing the coverage threshold down to 80% for now. We should be able to get it back to 85 and even 90 once #1132 and #1133 are done 😊~~

Edit: Wasn't necessary in the end :)
Edit edit: Yes it was 😅

andrewtavis:
Put methods within the views have been switched to updates to match faqs as the tests were failing without it.

andrewtavis (approved):
Thanks so much for the amazing work here, @sh-ran! Note that on update the order of the social links is at times shifting, but we can figure this out later :) Great to have this level of standardization of the backend now! 🚀

andrewtavis:
Tests at 89% locally and 84% here... Will end up reducing the cover threshold.

PR #1395, opened 2025-07-30
Feature remove logs

OmkarSathish:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

This has 3 commits in it and the purpose of each commit is as follows -
1. Fixes the `i18n-check` bug/error.
2. Adds the `no-console` rule to the `eslint.config.mjs` .
3. Removes all the `console.log()` used in the `/frontend` directory.

### Related issue

- #1387

PR #1402, opened 2025-08-02
Fix password visibility issue in login form

amyanger:
Remove console.log statement that was logging user input including passwords to browser console during form interactions. This eliminates the security concern where passwords were briefly visible in developer tools.


andrewtavis:
Hey @amyanger 👋 Thanks for the quick PR :) This change is already included in #1395. What we're talking about here is actually the password field being visible on the page. It doesn't work with screen recordings, but if you go to `auth/sign-in` and log in with username `admin` and password `admin`, and then click `Sign in`, then you'll be able to see the password displayed for an instant before the page reroutes.

Can you let me know if you can see that?

amyanger:
> Hey @amyanger 👋 Thanks for the quick PR :) This change is already included in #1395. What we're talking about here is actually the password field being visible on the page. It doesn't work with screen recordings, but if you go to `auth/sign-in` and log in with username `admin` and password `admin`, and then click `Sign in`, then you'll be able to see the password displayed for an instant before the page reroutes.
> 
> Can you let me know if you can see that?

Hey @andrewtavis  my bad so sorry about that, I will have a look and see what you mean! Thanks for being patient with me!

amyanger:
Hey @andrewtavis thank you! Do you have any suggestions for other issues I could work on?

andrewtavis (approved):
Thanks for the fixes here, @amyanger! This does allow the `type` within `FormTextInputPassword` to propagate. Looks like we have an issue that we're now not able to show the password right now as for some reason the eye icon is sending the form? I'll make an issue for this and ping you in there!

Issue #682, opened 2024-02-02
Improvements of the User model


kurtisolo:
### Terms

- [X] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

`User` model in `backend/authentication/models.py` have some fields that should be discussed in my opinion.

Since the model is extending `AbstractUser` it is not recommended to have a `user_name` field, since there's already a default one. If I am not missing some specific reason for that I think we should then delete that field, or we'll have a useless double field.
Same with the `password` field: Django already provides both the field and the security stuff (hashing, salting, etc.). Also I don't really like to have this field unnecessairly hardcoded for security reasons.

If there are some reasons for the choices that were already made maybe link me the relative discussion/documentation and sorry for the useless issue :P

andrewtavis:
CC @to-sta for the discussion :)

andrewtavis:
Closed by #1425 :) Am going through old issues atm, @kurtisolo, and realized this had never been taken up. I agree that we can use the `AbstractUser` defaults! Switched it so that `username`, `password` and `email` are inherited now :)

Thanks for the issue, and hope you're well! 😊

PR #1425, opened 2025-08-11
#682 Update user model in backend to not recreate default field

andrewtavis:
### Contributor checklist

---

### Description

Removes three fields from the user model in the backend as they're in the default `AbstractUser` class that we use from Django. Fixed the test for sign up as that was passing a name rather than a username, which was accepted by our original custom field, but not by the default (we were accepting spaces for a full name as a username). Also includes a minor frontend fix from a refactor I did.

Will merge when tests pass :)

### Related issue

- #682


andrewtavis:
@to-sta and @sh-ran: Requested review just so you both are aware of the changes. I do agree that we should use the defaults where possible for the user model as suggested in #682, so this is sending those changes 😊 Will bring this in :)

Issue #1433, opened 2025-08-15
Color of default organization images is changing (specifically to dark mode)


andrewtavis:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

An issue that's happening within the image carousel is that the default images for organizations - the fists from the landing page - are changing colors. If you're in light mode and you go to an organization page you'll see the carousel like this:

<img width="529" height="348" alt="Image" src="https://github.com/user-attachments/assets/ca3defc0-7200-4a94-a015-8c519810a02e" />

Then if you reload the page you'll see that the fists have changed to white, which is the dark mode images:

<img width="527" height="349" alt="Image" src="https://github.com/user-attachments/assets/7d629960-bf63-48a5-9792-6d138592c6a3" />

Note: My default color mode is dark mode, so it could be switching based on this.

This issue would be to find out what's happening and fix the issue so that the light mode images are always shown and not switched to dark mode.

OmkarSathish:
@andrewtavis i'd like to work on this

andrewtavis:
Thanks for all of your efforts here, @OmkarSathish! Really do appreciate all the work you're putting in :)

andrewtavis:
Closed by #1438 🚀 Great to have you in the community, @OmkarSathish! 😊

andrewtavis:
Please let us know if there's another issue of interest :)

PR #1438, opened 2025-08-16
bugfix: handled the auto img mode switch in the organization page

OmkarSathish:
### Contributor checklist

---

### Description

From what i found out - it was a hydration issue.
The solution does the following - 
- i completely avoided the server side handling of the logic of the mode
- made sure it waits for proper DOM mounting before setting images
- it also handles real-time color mode changes through watchers

### Related issue

- #1433


frontend/composables/useFileManager.ts
```diff
@@ -4,10 +4,11 @@ import type { ContentImage, FileImageMix } from "~/types/content/file";
 
 import { UploadableFile } from "~/types/content/file";
 
-const colorMode = useColorMode();
 const { token } = useAuth();
+
 const defaultImageUrls = computed(() => {
-  const imageColor = colorMode?.preference == "light" ? "light" : "dark";
+  const colorMode = useColorMode();
```
andrewtavis, line 10: These two lines were ultimately what was needed here, @OmkarSathish :) We need to use the `.value` as you found out, and the color mode call needs to be in the function. No need for the new composable. Specifically we were facing an issue where it was causing the image carousel position dots to malfunction. Not really sure what was going on, but we would scroll to the end of the images, and then we'd scroll to past the end to the first image, but the last dot would be highlighted indicating that we were still viewing the last image.

andrewtavis (approved):
Thanks for the fix here, @OmkarSathish! Really has been something that I've been looking at for a while, and it's great to finally have this fixed 😊

frontend/composables/useFileManager.ts
OmkarSathish: oh, i see. I unnecessarily complicated it 🫥 rookie mistake

andrewtavis: No stress! It happens :)

Issue #1458, opened 2025-08-24
Check that all entity instances of content models are inheriting from their base models


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

Something that we're realizing is that it's normal in the backend now that entity models that are versions of content models are not inheriting from the base content models. Instances of this include FAQ, social links, and text models.

This issue would make sure that all entity models for groups, organizations and event are inheriting from their base content models.

The resource models have been written correctly just now, and can be used as a reference for how to do this.

OmkarSathish:
@andrewtavis i'll pick this up

andrewtavis:
Would be great to get your help here, @OmkarSathish! Please let us know if you have any questions :)

andrewtavis:
Closed by #1464! 😊 Thanks so much for all your hard work across the projects, @OmkarSathish!

PR #1464, opened 2025-08-26
updated entities to inherit from their base content models

OmkarSathish:
### Contributor checklist

---

### Description

- updated all entity models to inherit from their base class
    - `EventFaq`, `GroupFaq`, `OrganizationFaq`  inherits from`Faq`
    - `EventSocialLink`, `GroupSocialLink`, `OrganizationSocialLink`  inherits from `SocialLink`

### Related issue

- #1458

### DIFF
--- a/backend/authentication/admin.py
+++ b/backend/authentication/admin.py
@@ -215,7 +215,7 @@ class UserFlagAdmin(admin.ModelAdmin[UserFlag]):
     Admin table for displaying User flags.
     """
 
-    list_display = ["user", "created_by", "created_on"]
+    list_display = ["user", "created_by", "creation_date"]
 
     def save_model(
         self, request: HttpRequest, obj: UserFlag, form: ModelForm[Any], change: bool
--- a/backend/authentication/factories.py
+++ b/backend/authentication/factories.py
@@ -133,6 +133,6 @@ class UserFlagFactory(factory.django.DjangoModelFactory):
 
     user = factory.SubFactory("authentication.factories.UserFactory")
     created_by = factory.SubFactory("authentication.factories.UserFactory")
-    created_on = factory.LazyFunction(
+    creation_date = factory.LazyFunction(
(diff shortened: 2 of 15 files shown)


andrewtavis:
Thanks so much for the work here, @OmkarSathish! Would we be able to create a base `Text` content model and then inherit this into the text models for organizations, groups and events? Organizations and groups have a `donate_prompt` field that would also need to be added to the final inheriting model, but aside from that it looks the same :)

OmkarSathish:
I'll work on it now
- Text content model
- donate_prompt

andrewtavis:
Thanks, @OmkarSathish! I think we'll be good with this PR once those changes come through 😊 Do you also want to check the other models in the backend and make a new issue that details some others that could be combined in the way that you're doing it here? You could then work on that one next?

OmkarSathish:
Yes, I'll check on what can be better written and raise an issue. For this one, I'll send the PR within the next hour or 2

OmkarSathish:
@andrewtavis I have a doubt. In the `Flag` Model Timestamp Field Naming - `EventFlag`, `GroupFlag`, `ResourceFlag`, `UserFlag` all use `created_on` while `OrganizationFlag` uses `created_at`. Shall i raise an issue (since its not exactly related to inheriting from the base model) and fix this or just add another commit to the same PR?

andrewtavis:
We can just remove the `created_at` and do the necessary fixes :) Exactly why we're doing this, as that's an error :) I can do this or you can, @OmkarSathish 😊

OmkarSathish:
@andrewtavis on it

andrewtavis (approved):
Thanks for the great work here, @OmkarSathish, and also for bringing our attention to the `created_at` situation. I switched the codebase to just use `creation_date` for everything. I think that'll be best for consistency 😊

Feel free to do the exploration and open the issue with the models that you find can be inherited into their respective entities! :)

Issue #1375, opened 2025-07-22
Define session model in backend


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

As of now we're using a mock session model in the backend for user authentication. Ideally we'd have a fully defined model that has a table in the database. We'd also have a get session service to return the current session of the user.

### Contribution

Happy to support and help review! 😊

to-sta:
<ins>Just for context:</ins> At the moment we use basic authentication tokens that are provided with [django-restframework](https://www.django-rest-framework.org/api-guide/authentication/). I would recommand, we are switching to [jwt](https://datatracker.ietf.org/doc/html/rfc7519) `access` and `refresh` tokens.

Klath123:
Hey @andrewtavis I would like to work in this issue

andrewtavis:
Hey @Klath123 👋 Thanks for your interest in the project! We've been talking with @MuhammadNizamani about picking this issue up. @MuhammadNizamani, do you still have interest in this? If so, then maybe we can find another issue for you to pick up, @Klath123 :)

Klath123:
No worries , @andrewtavis I would love to work on some other issue 👍

MuhammadNizamani:
@andrewtavis  I'm interested in this issue and have been going through the backend code to get familiar and comfortable with it. Again thanks for your guidance

andrewtavis:
Thanks for picking this up, @MuhammadNizamani! Are you still considering working on one of the testing issues in #1132 or #1133 first? Maybe @Klath123 could work on the other one? :)

MuhammadNizamani:
@andrewtavis ok I will first work on tests so give this issue to @Klath123

andrewtavis:
You can certainly take on this issue, @MuhammadNizamani :) Would prefer it if you did this one, but maybe you can work on #1133 first so that you have an issue to get into the codebase with before jumping into this one? @Klath123 could then help @sh-ran a bit with #1132 😊

Klath123:
Ok! on it

sh-ran:
IF this issue is not being worked on as @MuhammadNizamani is working on the serializer tests and @Klath123 will be working on the view tests, would it be ok if I can take this one up and send in a PR in the next couple of days? @andrewtavis 

[...]

PR #1473, opened 2025-08-31
feat: Moving from token authentication to JWT authentication. Added Session GET views

sh-ran:
### Contributor checklist

---

### Description

This PR upgrades our current authentication system from `TokenAuthentication` to `JWTAuthentication` using `rest_framework_simpleJWT` module. I have set the `JWT_SETTINGS` mostly to defaults only changing a one of two here and there as the default seemed good enough.

Session GET View is working well but the same cannot be said for `POST` and `DELETE`. I will work on these in the coming days.

PS: A whole lot of files had to be changed because of the authentication getting changed, most of them being tests.

### Related issue

- #1375

### DIFF
--- a/backend/authentication/admin.py
+++ b/backend/authentication/admin.py
@@ -29,7 +29,7 @@ admin.site.register(SupportEntityType)
 # MARK: User Creation
 
 
-class UserCreationForm(forms.ModelForm[UserModel]):
+class UserCreationForm(forms.ModelForm):
     """
     A form for creating new users.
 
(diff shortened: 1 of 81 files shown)


andrewtavis:
Thanks for the PR, @sh-ran! Really great to have this! 😊

sh-ran:
Regarding the `mypy` errors, I had to use `class Meta` to avoid getting these errors when trying to start django:  `Model is not subscriptable` error.

sh-ran:
`POST` and `DELETE` methods are now working for the session models. Tests have been included for all the methods. Everything looks good apart from the `mypy` errors. I get a `django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.` error which I think is related to the `JWT settings` with in the `core/settings.py` file. I just cant figure out what's wrong tho. Will be working on fixing these errors now.

sh-ran:
I have chose to add ignore tags to ignore the mypy errors as fixing it by doing `ModelAdmin[UserModel]` would cause django not start which gives the error `ModelAdmin not subscriptable` error.

PR #1494, opened 2025-09-09
fix: replace empty arrow function with nullable guard to satisfy lint

jayamrutiya:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://github.com/jayamrutiya/activist/tree/code-smells/unexpected-empty-arrow-function) and not the main branch

---

#### What changed
- Replaced removeGuard = () => {} with a nullable function reference let removeGuard: (() => void) | null = null.
- Updated cleanup logic to use safe invocation removeGuard?.() on unmount.
- This resolves the SonarLint rule typescript:S1186 (Unexpected empty arrow function).

#### Why
- An empty arrow function was only used as a placeholder, which is misleading.
- Nullable references make it clear that the function may not always be assigned.
- Using optional chaining (removeGuard?.()) is a safer and cleaner pattern.

#### Benefits
- Code now complies with linting rules.
- Improves clarity and maintainability.
- Prevents potential misuse of empty no-op functions.


andrewtavis:
Hey @jayamrutiya 👋 Wanted to check in with you on this particular PR here as it seems that you're using Sonar based linting, which isn't something that we're using on the project. I understand that ESLint and Soner have different uses, but before we start bringing in code that's based on Sonar I think it'd be good that we understand what some of the conflicts that we might experience could be.

One way forward I could see here is that you'd open an issue that suggests that the project add in an ESLint plugin for Sonar or add Sonar based checks as a dependency. In there we could discuss the benefits and if we have the capacity right now to do the changes to get us up to this new set of standards. Would you be willing to write something up for us and go through the pros and cons in an issue?

jayamrutiya:
Hi @andrewtavis Thanks for raising this. You’re right that the project doesn’t use Sonar right now. Let’s go ahead and merge this PR since the change itself is safe and improves clarity, and I’ll open a separate issue to discuss whether we want to adopt Sonar-based checks in the future.

Issue #1500, opened 2025-09-11
Hardcoded user ID in event creation (frontend/stores/event.ts) causes incorrect attribution


monu808:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

Description Event creation currently sends a hardcoded created_by UUID, ignoring the authenticated user and risking all events being attributed to the same account. This lives in [frontend/stores/event.ts]

Impact

Incorrect authorship and audit trails
Possible security/abuse vector (spoofed ownership)
Harder to test permission logic on backend
Breaks multi-user scenarios
Steps to Reproduce

Sign in as any user (not the hardcoded one)
Create an event via UI
Inspect resulting event (e.g. via network tab or backend admin)
Observe created_by does not match the authenticated user

Expected Backend derives creator from the auth token/session or the frontend supplies the actual authenticated user ID dynamically.

Actual Static UUID always sent.

Proposed Fix

Remove hardcoded created_by
Let backend infer user from Authorization header OR inject current user id from user store
Add error handling for failed requests
Add a test asserting attribution correctness

andrewtavis:
Closed by #1501 🚀 Thanks for opening this issue, the conversation around the solution and the contribution to improve these methods, @monu808! 😊 Hope we can work with you again :)

PR #1501, opened 2025-09-11
Remove hardcoded created_by UUID from create functions

monu808:
Contributor checklist
 This pull request is on a separate branch and not the main branch
 I have run the tests for the backend and frontend depending on what's needed for my changes as described in the testing section of the contributing guide
Description
Removes the hardcoded created_by UUID previously sent in entity create requests (events, groups, organizations, resources). The stores/pages now:

Derive the authenticated user id (if available) from auth/user state
Omit created_by when unauthenticated, allowing the backend to infer from the session/token
Prevent incorrect attribution and potential spoofing
Tested by:

Logging in and creating an event/group/org/resource: verified network payload includes my user id.
Logging out and attempting creation: request rejected (401) with no created_by in payload.
Confirmed created objects show correct creator in API/admin.
Related issue
#1500

### DIFF
--- a/frontend/pages/events/create.vue
+++ b/frontend/pages/events/create.vue
@@ -143,14 +143,13 @@
               />
             </div>
           </div>
-          <!-- TOOD: add events to CardTopicSelection -->
+          <!-- TODO: Add events to CardTopicSelection. -->
           <CardTopicSelection
             v-model="formData.topics"
             class="mt-5"
             pageType="event"
           />
           <div class="py-3"></div>
-          <!-- for spacing -->
           <div
             class="card-style mx-14 flex w-full justify-between gap-6 px-5 py-6"
           >
@@ -158,8 +157,7 @@
               <h3 for="roles" class="block font-medium">
                 {{ $t("i18n.pages.events.create.roles") }}*
               </h3>
-              <!-- TODO: replace this input with something that lets you make
-                roles and such -->
+              <!-- TODO: replace this input with something that lets you make roles. -->
               <input
                 v-model="formData.roles"
                 id="roles"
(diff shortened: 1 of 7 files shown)


andrewtavis:
Hey @monu808 👋 We'll take a look at this PR in the coming days/week as we look into this functionality. Big thing is that we've yet to make a decision on how the user will be determined in creation methods. We have the option of sending as you've worked on here, or it would be determined by the session in the token that's being sent. As of now it's being determined by the token in the backend, but we didn't have access to session management until recently.

CC @to-sta and @sh-ran for your two's input here: Would you two prefer to derive the user from the token as is being done now, or would you prefer to receive the `user_id` directly? Note that the second option would require a new issue to do the needed refactoring for this.

andrewtavis:
There are also security concerns to consider here, as sending the `user_id` in requests rather than inferring it from our backend processes might be bad practice. If we can decide on this, then maybe the contribution here @monu808 would be that you can simply remove the `user_id` and the requirement for it from the requests (if there is one in the backend).

monu808:
> There are also security concerns to consider here, as sending the `user_id` in requests rather than inferring it from our backend processes might be bad practice. If we can decide on this, then maybe the contribution here @monu808 would be that you can simply remove the `user_id` and the requirement for it from the requests (if there is one in the backend).

Agreed. Best practice is to not send user_id/created_by from the frontend and let the backend infer it from the authenticated request.

andrewtavis:
So should we proceed with the refactor and remove `user_id` from the creation methods then, @monu808? Feel free to send along another commit to update this and we'll finalize from there 😊

monu808:
> So should we proceed with the refactor and remove `user_id` from the creation methods then, @monu808? Feel free to send along another commit to update this and we'll finalize from there 😊

Yeah sure just give me couple of minutes to think if there exist a different path to do it or otherwise I'll make a commit and remove 'user_id'

andrewtavis:
Thanks, @monu808! We'll take a look at this now while we're reviewing things :)

Issue #1531, opened 2025-09-21
Resources are sometimes reordered on edit


andrewtavis:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

An issue that we've been noticing in the platform is that editing resources at times causes the order of the resources to shift, with the most likely bug being seen being that the edited resource goes to the end of the list of resources. This can be reproduced via the following:

- Load up the project via `docker compose --env-file .env.dev up --build`
- Sign in to the platform with the admin account
    - Username: admin
    - Password: admin
- Navigate to view the organizations and enter an organization page (you'll be on About)
- Navigate to Resources via the left sidebar
- Click the edit icon to edit a resource that's not the last one in the list
- Submit the edit
- There's a chance that the resource has now moved to the end of the list, or if not then try it again a few times

Getting some support with this would be very appreciated! Happy to support and review a PR 😊

Ethan-new:
I will work on this! :)

andrewtavis:
Thanks for the interest in the project and support, @Ethan-new! Please let us know if you need any support :)

Ethan-new:
Ok I fixed it pr above

Ethan-new:
ok new pr above fixed :) #1536

andrewtavis:
Closed by #1536 :) Thanks for the great first contribution and keeping at it with another PR, @Ethan-new 😊

PR #1533, opened 2025-09-21
fix: resolve resources list ordering and use existing CardResource co…

Ethan-new:
- Remove unnecessary CardSearchResultResource component
- Use existing CardResource component for consistent resource display
- Add proper sorting by order field with fallbacks to creation date and ID
- Implement reactive data fetching with refresh capability
- Add window focus listener to refresh list when returning to page
- Add EntityType import for CardResource component
- Fixes issue where edited resources would appear at bottom of list

### Contributor checklist

- [✅ ] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [✅ ] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

Fix: Resolve resources list ordering issue 
Problem: Resources were moving to the bottom of the list after being edited

**Root Cause**
- The main resources list page used static useFetch without reactivity
- When resources were edited in organization/event/group contexts, those stores would refresh, but the main list wouldn't update
- The backend likely returns resources in modification order, causing edited items to appear at the bottom
- Missing component (CardSearchResultResource) was causing runtime errors

**Solution**
- Removed unnecessary component: Deleted CardSearchResultResource.vue and used the existing CardResource component
- Implemented proper sorting: Added hierarchical sorting by order field → creation date → ID
- Added reactivity: Implemented refresh capability and window focus listener to catch updates
- Maintained consistency: Resources now maintain their position based on the order field

I verified the fix by checking that the CardResource component renders properly and the sorting logic works as expected with different resource configurations.

### Related issue

- #1531


andrewtavis:
Thanks, @Ethan-new! We'll get to the review in the coming days :)

nicki182:
The above doesn't fix the issue. The files that needs to be changed are the modal edits for resources since it's reorder after edit + the file you changed it's not used today.

nicki182:
If you have questions of how to replicate the bug you need to go either group/organization/event and within does go to resources and then edit a resource and see if it's reordered after editing

Ethan-new:
I hope I found the solution you were looking for. See new PR
#1536

Issue #919, opened 2024-06-30
Update to Nuxt 4


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One thing that the community should get to sooner rather than later is update underlying frameworks early in the process so we don't need to think about this later when the platform grows. The second of these is Nuxt, which will hit version 4.0 soon at the time this issue is being made. This isn't needed for the first release, but should be gotten to by the end of 2024 :)

There's currently no official upgrade guide, but I'll add it to this issue when it's up!

- Blocks #1148

### Contribution

I'll be working on this once the new version of Nuxt is released!

CreativeStudio16:
setting

aakashkrsingh1:
Hey @andrewtavis , a new version released 2 weeks ago. Can I pick this up, please? 
_New to open source._

supraaxdd:
Hi, is this issue still blocked? If not, I'd be able to work on it.

andrewtavis:
Thanks for your interest, @supraaxdd! Looks like Nuxt 4 is out. Could you send along a message before the PR that outlines the changes that need to be made? Will help us to plan reviews and what needs to come in first 😊

supraaxdd:
> Thanks for your interest, [@supraaxdd](https://github.com/supraaxdd)! Looks like Nuxt 4 is out. Could you send along a message before the PR that outlines the changes that need to be made? Will help us to plan reviews and what needs to come in first 😊

I'll take a look at what changes are in the docs vs. what's in the source code and see what needs to be changed. Once I get my head around the whole show, I think I could make a quick summary :)

supraaxdd:
Here's my little update (perhaps a summary) of what I found.

Firstly, I'd like to say that all of the change notes can be found [here](https://nuxt.com/docs/4.x/getting-started/upgrade).

Here is the summary of the changes:
One of the big changes is the directory structure, with the `app` directory being used as the default `srcDir`. 
An excerpt from the migration guide can be found below:
 - the new Nuxt default `srcDir` is `app/` by default, and most things are resolved from there.
  - serverDir now defaults to `<rootDir>/server` rather than `<srcDir>/server`
  - layers/, modules/ and public/ are resolved relative to `<rootDir>` by default
[...]

PR #1547, opened 2025-09-28
#919 Update frontend to Nuxt 4 + other updates

andrewtavis:
### Contributor checklist

    - Tests will need some work

---

### Description

This PR starts the process of updating the frontend dependencies. We need to update to Nuxt 4, and with that lots of other dependencies are needed. Goal is that this will set a new standard for the frontend and then we can maintain it from there :)

### Related issue

- #919
- #1262


andrewtavis:
OK @nicki182, to be expected we have failing TS checks and also some unit tests are failing. The unit tests seem to mostly be imports. TS will take a lot more time.

I'd suggest that we do some more work on this when we have some time and get it to a point that all functionality is there and all tests are passing. We can then update the following:

- [x] Prettier
- [x] ESLint and all plugins
- [x] TypeScript and all plugins
- [x] Tailwind and all plugins
- [x] Pinia and all plugins
- [x] Node (needs to be in `package.json` and in the `Dockerfile`)

From there we can start with the other dependencies and get them to a higher level as well. We have a lot of `tiptap` that's out of date, and other dependencies as well.

nicki182:
Ok, I will see about it in the next few days

andrewtavis:
Discussing this with @nicki182 now :) There are a few bugs now related to Pinia, but we'll be making new issues once this is merged.

PR #1558, opened 2025-10-05
1557 fix hydration frontend

### LINKED ISSUE #1557: Update data fetching
### Terms

### Behavior

Sometimes hydration is not working when updating or creating.

nicki182:
### Contributor checklist

---

### Description

Updating data would sometimes not work, since for entities the usual cycle to update is to call fetchById that uses useAsyncData but with the new update, this uses caching stored from the initial fetch so I added a new argument in fetchById called refreshData to refresh the data that calls refresh from useAsyncData
### Related issue

- Closes #1557

### DIFF
--- a/frontend/app/pages/events/index.vue
+++ b/frontend/app/pages/events/index.vue
@@ -27,6 +27,9 @@
 <script setup lang="ts">
 import { ViewType } from "~/types/view-types";
 
+const eventStore = useEventStore();
+const { events } = storeToRefs(eventStore);
+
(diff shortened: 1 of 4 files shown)


andrewtavis (approved):
All's working great here as well, @nicki182! 🎃 Noting what you just mentioned: we're using the default `fetchById` without `refreshData` to use the cache of the `useAsyncData` unless its necessary because we're are updating data :) Just a note if we're wondering about it at another time.

Issue #1573, opened 2025-10-11
Accessibility/UX: Drag handles invisible in dark mode (poor contrast)


aasimsyed:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description
The drag handles for reordering FAQ entries and Resources are difficult or impossible to see in dark mode due to insufficient color contrast. The handles appear to be dark/black on a dark background, making them nearly invisible to users.

### Steps to Reproduce
1. Enable dark mode in the application
2. Navigate to any page with draggable items:
   - Organization FAQ page (`/organizations/{orgId}/faq`)
   - Organization Resources page (`/organizations/{orgId}/resources`)
   - Group FAQ page (`/organizations/{orgId}/groups/{groupId}/faq`)
   - Group Resources page (`/organizations/{orgId}/groups/{groupId}/resources`)
   - Event FAQ/Resources pages
3. Try to locate the drag handles (grip icon) on the left side of each card
4. Observe that the handles are barely visible or completely invisible

### Expected Behavior
- Drag handles should be clearly visible in both light and dark modes
- Color contrast should meet **WCAG 2.1 Level AA** standards (minimum 4.5:1 ratio for normal text, 3:1 for large text/icons)
- Users should easily identify which items are draggable
- Handles should be distinguishable from the card background

### Actual Behavior
- Drag handles are dark/black in dark mode
- Handles blend into the dark card background
- Users cannot easily find or interact with drag handles
- Poor visual affordance for drag-and-drop functionality

Nishant-Manocha:
I'd like to take this on! Please assign this issue to me.

andrewtavis:
Sure thing, @Nishant-Manocha! Please let us know if you need support with the implementation. We'd be happy to help 😊

Nishant-Manocha:
Thanks, @andrewtavis, that's great to hear! I'm starting on the
implementation now and will reach out if I hit any roadblocks or need a
second set of eyes.

Nishant-Manocha:
Hi @andrewtavis,

I've finished making the changes for #1573 (Drag handles invisible in dark mode).

I'm ready to push the branch and open a Pull Request. Please confirm if now is a good time, or if there's anything else I should do first. Thanks!

andrewtavis:
[...]

PR #1577, opened 2025-10-13
Fix Drag Bug

Nishant-Manocha:
Fix the Bug Accessibility/UX: Drag handles invisible in dark mode (poor contrast) #1573

### DIFF
--- a/frontend/app/assets/css/tailwind.css
+++ b/frontend/app/assets/css/tailwind.css
@@ -292,6 +292,14 @@ MARK: Base
   .style-warn {
     @apply border border-primary-text bg-warn-yellow text-primary-text focus-brand hover:bg-warn-yellow/80 active:bg-warn-yellow dark:border-warn-yellow dark:bg-warn-yellow/10 dark:text-warn-yellow dark:hover:bg-warn-yellow/25 dark:active:bg-warn-yellow/10;
   }
+  /* Draggable handle styling intentionally left to component-level classes. */
+/* Improve drag handle affordance: visible color, subtle background and focus ring */
+.drag-handle {
+  @apply text-primary-text inline-flex items-center justify-center p-1 rounded-md;
+  }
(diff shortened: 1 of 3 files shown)


andrewtavis:
@aasimsyed, could you assist here with whether the change to the list variable rather than the entity props is appropriate for the `v-if`. 

As far as the other change, personally I'd prefer that we can an `IconDraggable` component here that contains the style as the global styles really should be things that we're using across multiple components. Applying styles to HTML elements like p tags and headers to me is inline with Tailwind, but we shouldn't be expanding out the global styles file with CSS that is could be contained within a repeatable component.

nicki182:
> @andrewtavis @Nishant-Manocha
> 
> Which issue is this change related to? I noticed that the PR references this issue, but only the CSS change seems in scope:
> 
> #1573
> 
> Quick question though: I noticed we're keeping faqList (ref) and organization.faqEntries (store) in sync with a watcher. Would a computed property be simpler here?
> 
> ```ts
> const faqList = computed({
>   get: () => orgStore.organization.faqEntries || [],
>   set: (value) => { orgStore.organization.faqEntries = value; }
> });
> ```
> 
> This would eliminate the dual state and watcher. Am I missing something, or is there a reason we need the ref + watch pattern? Just curious!
> 
> @andrewtavis there was discussion on making the draggable component a few days back. Do we want to wait for that instead of making the changes here?

I am learning as I go with vue and I'm from the world of react so I'm used to have a useEffect with a useState, not ideal in a lot of cases. I can change it towards this, but @aasimsyed I am doing refactor and this will be changed. We will no longer get the data from the store, we will get it from a composable that will check if in the store or in cache data within Nuxt, if not it will get it from the server.

andrewtavis:
Closing this as #1610 sent along the needed changes here :) Thanks for your efforts, @Nishant-Manocha! Hope we can work with you again 😊

Issue #1608, opened 2025-10-20
Bug: Join URL field not persisting in Group About page edit form


aasimsyed:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

## Description
When editing the "About" section of a group page, the join URL field is not being saved/persisted after form submission. The field accepts the new value during editing but reverts to the default URL after the modal closes.

## Steps to Reproduce
1. Navigate to a group's About page
2. Click the edit icon to open the edit modal
3. Fill in the join URL field with a custom URL (e.g., `https://test.activist.org/join-group?run=123456`)
4. Verify the field contains the new value
5. Submit the form
6. Wait for modal to close
7. Check the join button's href attribute

## Expected Behavior
The join button should have the custom URL that was entered in the form.

## Actual Behavior
The join button still shows the default URL (`https://activist.org/`) instead of the custom URL that was submitted.

aasimsyed:
I opened a PR for this, so I'll assign to myself.

aasimsyed:
this was fixed in https://github.com/activist-org/activist/issues/1603

andrewtavis:
Thanks for the triage, @aasimsyed! 😊

Issue #1603, opened 2025-10-19
Move `get_invovled_url` to `text` tables


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

As discussed in the most recent sync, it might be more efficient if we save the `get_involved_url` to the `text` tables for organizations, groups and events. The reason for this is that this URL is edited along with the text that accompanies it within the given modals, and having it in a separate place has meant that A) We've never actually updated it; and B) We'd need to do multiple calls for one submit of the modal rather than to just one table. Moving the field to the `text` table would solve these issues.

CC @nicki182 and @sh-ran

Happy to support with this! Maybe @sh-ran and I can finalize this in a call soon?

gabrielsrs:
Hi @andrewtavis, I would like to try this issue!!

andrewtavis:
Thanks for your interest here, @gabrielsrs! Note that there might be a fix for this already in #1611 :) You can take this up if that PR doesn't close this 😊

PR #1611, opened 2025-10-20
Fix join URL persistence issue in About section editing

aasimsyed:
### Contributor checklist

---

## Problem
Custom join URLs were not persisting after editing the About section. E2e tests failed because join buttons reverted to default URLs instead of maintaining custom URLs.

## Root Cause
`getInvolvedUrl` field is stored on `Organization`/`Group` models, not text models. Frontend was trying to update it through text endpoints that didn't handle this field.

## Solution
Modified backend text endpoints to handle `getInvolvedUrl` directly in a single API call:

**Backend**: Extract `getInvolvedUrl` from request, update text fields via serializer, then update organization/group `getInvolvedUrl` field directly.

**Frontend**: Send `getInvolvedUrl` in text request payload, maintain simple single API call logic.

Tested using e2e tests and manually.

### Related issue

- https://github.com/activist-org/activist/issues/1608

### DIFF
--- a/backend/communities/groups/views.py
+++ b/backend/communities/groups/views.py
@@ -617,7 +617,7 @@ class GroupTextViewSet(GenericAPIView[GroupText]):
             )
 
         # Extract getInvolvedUrl from request data if present.
-        get_involved_url = request.data.pop("getInvolvedUrl", None)
+        get_involved_url = request.data.pop("get_involved_url", None)
 
         # Update the text fields.
         serializer = self.serializer_class(group_text, data=request.data, partial=True)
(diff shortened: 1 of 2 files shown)


andrewtavis:
Thanks for the PR, @aasimsyed! 🎉

question (non-blocking): Does this need to be done for events as well, and if so, would you like to add it to this PR?

note/question: This does a fix for #1603 such that we wouldn't need to do more than one call as the backend can extract the field. Might be best to do it this way so that the text tables are just what's directly seen?

aasimsyed:
@andrewtavis I didn't notice that issue earlier before I made this PR. I agree that it's the better solution. should we close this PR and wait for the #1603 implementation instead?

andrewtavis:
praise/note: Thanks so much for the note here, @aasimsyed! Talking about this with @nicki182 right now, we're realizing that there is a definite use for having this URL in the text table because it opens up the ability for a user to set different get involved URLs based on the localization of the texts - i.e. for those viewing Spanish content, please redirect them to the Spanish join form.

Closing this given the discussion above 😊

CC @sh-ran 👋

PR #1613, opened 2025-10-21
fixed datetime and frontend docker compose

### LINKED ISSUE #1600: [Bug] Event can be created with zero duration (Start time equals End time) (Admin Panel)
### Terms

### Behavior

The date/time validation in the admin panel (/admin/) allows an event to be created with an "End time" that is identical to the "Start time". This allows the creation of events with zero duration, which is logically inconsistent.

Steps to Reproduce:

    Go to the /admin/ panel to add a new event.

    Fill in all mandatory fields.

    Set Start time to a future date, for example: October 20, 2025, 06:00.

    Set End time to the exact same date and time: October 20, 2025, 06:00.

    Save the event.

Actual Behavior: The event is created successfully.

Expected Behavior: The form should display a validation error, such as "End time must be after start time", as the requirement implies it must be strictly posterior.

Soumadip-Mishra:
Fix #1600  
i) Fixed the datetime logic error.
ii) Added node_modules in the docker compose to ignore , because it was giving EAccess error in windows , because node_modules made a different cache folder on each re-render .

### DIFF
--- a/backend/events/tests/test_event_api.py
+++ b/backend/events/tests/test_event_api.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from typing import Any, TypedDict
 
 import pytest
-from dateutil import relativedelta
+from dateutil.relativedelta import relativedelta
 from rest_framework.test import APIClient
 
 from authentication.factories import UserFactory


Soumadip-Mishra:
@andrewtavis

andrewtavis:
Thanks so much for the PR, @Soumadip-Mishra :) We'll get to the review in the coming days 😊

@sh-ran, could you take a first look here and assess the needs for Windows?

backend/events/serializers.py
```diff
@@ -347,17 +347,20 @@ def _invalid_dates(self, start: Union[int, str], end: Union[int, str]) -> bool:
...
         # Convert to datetime if they're strings.
         start_dt = parse_datetime(start) if isinstance(start, str) else start
         end_dt = parse_datetime(end) if isinstance(end, str) else end
+        
 
         return (
             isinstance(start_dt, datetime)
             and isinstance(end_dt, datetime)
-            and start_dt > end_dt
+            and (start_dt >= end_dt or start_dt<curr_dt)
```
sh-ran: thought: I think we can remove `=` operator as the `start_dt` can be equal to `end_dt` for an event, so we can just check if the `start_dt` is ahead of `end_dt`.

docker-compose.yml
```diff
@@ -72,6 +72,7 @@ services:
       fi"
     volumes:
       - ./frontend:/app
+      - /app/node_modules
```
sh-ran: @andrewtavis This should not have caused any issues on windows or any machines for that matter. Unless they are running it locally. I suggest we revert this file.

andrewtavis: Reverted :)

backend/events/serializers.py
andrewtavis: Changing this gives us:

```
FAILED events/tests/test_event_api.py::test_EventDetailAPIView - TypeError: can't compare offset-naive and offset-aware datetimes
```

Maybe we can make an issue to look into this?

andrewtavis:
@sh-ran, let's fix this and bring it in in the sync tomorrow :)

andrewtavis:
If you could figure out the fix, that'd be great!

andrewtavis:
Likely is that the current date isn't the correct type, so if we wanted to go back to a non-equal comparison and fix how that's being done then we should be good to merge here :)

Soumadip-Mishra:
> praise: Thanks for the changes here, @Soumadip-Mishra! Really great to have you as a contributor 😊
> 
> note: We didn't merge in your change to the Docker Compose file. If you continue to experience issues with Docker, please let @sh-ran and I know and we'll try to assist!
> 
> note: I kept the `=` in the date comparison as the test was failing without it. Might be good to make an issue to fix these comparisons and make sure that dates are consistent throughout the backend?


I learnt that node modules make a different cache file for each instance , to not conflict with local file , we need to unmap the node modules folder with the local machine , don't know why it is not giving error in your case

sh-ran:
Hey @andrewtavis,
You were right, there was an mismatch between the datetime types and I have fixed it. Will be sending along the changes as soon as all the tests pass. Apologies for not running the tests before hand last night.

@Soumadip-Mishra could you please attach a screenshot for the error? Maybe I can take a look at it.

backend/events/serializers.py
```diff
@@ -347,17 +349,18 @@ def _invalid_dates(self, start: Union[int, str], end: Union[int, str]) -> bool:
         Returns
         -------
         bool
-            True if the start is after the end (invalid).
+            True if the start is after the end time, or start is before current time (invalid).
             False otherwise (valid).
         """
+        curr_dt = datetime.now()
```
sh-ran: @andrewtavis seems like I cannot push the changes to the PR as I do not have permissions.

Basically, you'd have to change this line to `curr_dt = datetime.now(timezone.utc)`

And within the tests `events/tests/test_event_api.py`,

You have to add these lines: 

Image

That would make the test dates to be set to 2 years from now so that the serilizers can pass.

We could also just disable the days in the frontend so that users cannot select days that are in the past. So we can change the serializer to just check for the end_date being set behind the start_date.

andrewtavis (approved):
Approving again :)

praise: Thanks @sh-ran for the help finalizing this! 😊

PR #1542, opened 2025-09-26
Image upload e2e tests in Organization Page

UnknownSean8:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [x] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description
This PR implement e2e tests for image uploads in the image carousel modal component.

Happy to have a sync on this soon to make the page/epic ready for Hacktoberfest.

### Related issue

- #1175

PR #1665, opened 2025-11-02
Image upload e2e tests

UnknownSean8:
### Contributor checklist

---

### Description

e2e tests for image uploads.

Created a new patch for rebasing and optimization from PR #1542.

One thing to take note of, when a user deletes an image and closes the modal, the image is not updated immediately on the image carousel and only updates when the page is reloaded.

### Related issue

- #1542


andrewtavis:
praise: Thanks for opening a new PR, @UnknownSean8! Appreciate you making reviews easy for us 😊

@aasimsyed, would you be able to do a first review here? :)

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      // Reload the page for the change to take effect.
+      await page.reload();
+
+      // Verify the number of image in the carousel matches the number of files in the modal.
+      // It is 3 since the default number of images in the Image Carousel is 3 images.
+      // TODO: Need to count the initial number of images in the Image Carousel instead
+      await expect(
+        organizationPage.aboutPage.getImageCarouselImages
+      ).toHaveCount(3);
```
aasimsyed: might be better if you calculated the initial count dynamically, like:

```ts
const initialCarouselCount = await organizationPage.aboutPage
  .getImageCarouselImages
  .count();
// ... perform actions ...
await expect(
  organizationPage.aboutPage.getImageCarouselImages
).toHaveCount(initialCarouselCount - 1); // after removal
```

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+        .closeButton(organizationPage.uploadImageModal.modal)
+        .click();
+
+      // Wait for the modal to close and page to update.
+      await expect(organizationPage.uploadImageModal.modal).not.toBeVisible({
+        timeout: 10000,
+      });
+
+      // Reload the page for the change to take effect.
+      await page.reload();
```
aasimsyed: is this bypassing a bug in the app? shouldn't the change be reflected immediately instead of requiring a reload? this seems to be bypassing a valid issue.

aasimsyed:
suggestion: you may also want to upload multiple images to check the carousel pagination as well instead of just a single image

andrewtavis:
Quick check in @UnknownSean8: Would you be able to finalize the comments from @aasimsyed so we can bring this in? :)

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
UnknownSean8: Hi @aasimsyed, this is currently bypassing a bug in the app where the changes (removal of images) are not reflected when the modal is closed. Afaik, the issue is due to the SSR issue, it used to happen during image upload as well but that is fixed.

suggestion: should we open a issue for this or is there already an issue for this?

UnknownSean8:
> Quick check in @UnknownSean8: Would you be able to finalize the comments from @aasimsyed so we can bring this in? :)

Sure, would be able to work on this as well as the other ticket by this week.

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
aasimsyed: Hi @UnknownSean8, yes I think opening a new bug issue and then update this test to only the expected behavior will be best. That way the test should fail until the bug is fixed. Would you like to open the issue?

UnknownSean8: That sounds good to me. Yup, I can open the issue and fix the test. Thanks!

UnknownSean8:
> suggestion: you may also want to upload multiple images to check the carousel pagination as well instead of just a single image

I am able to create the test for uploading multiple images but I'm not sure how to check carousel pagination. Do you have any idea how I can do this?

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
UnknownSean8: Issue created under #1791

andrewtavis:
Quick check in here, @UnknownSean8: Do you have a moment to look into this and finalize? Happy to hop on a call and support if needed :)

UnknownSean8:
Hi @andrewtavis, yes, I'll finish this PR by this week. Thanks!

andrewtavis:
Great to hear, @UnknownSean8! 😊

UnknownSean8:
Oops sorry, I'll have to delay this for another week.

UnknownSean8:
Hi @andrewtavis, the test is working fine with some additions to test multiple image uploads in the carousel. However, the tests are not passing now as the images uploaded are still not updated on the page without a refresh.

frontend/package.json
```diff
@@ -10,8 +10,8 @@
     "preview": "nuxi preview",
     "postinstall": "nuxi prepare",
     "test": "vitest run --run",
-    "test:local": "rm -rf blob-report && yarn test:local:desktop || true && yarn test:local:mobile || true && yarn test:local:merge",
-    "test:local:desktop": "cross-env TEST_ENV=local PLAYWRIGHT_BLOB_OUTPUT_FILE=./blob-report/desktop.zip playwright test --project='Desktop Chrome' --reporter=blob",
+    "test:local": "rm -rf blob-report && yarn test:local:desktop || true && gbyarn test:local:mobile || true && yarn test:local:merge",
```
aasimsyed: is `gbyarn` a typo here?

frontend/package.json
```diff
@@ -10,8 +10,8 @@
...
+    "test:local:desktop": "cross-env TEST_ENV=local PLAYWRIGHT_BLOB_OUTPUT_FILE=./blob-report/desktop.zip playwright test --project='Desktop Chrome' --reporter=blob --ui",
```
aasimsyed: --ui on test:local:desktop opens the Playwright interactive UI; it is incompatible with --reporter=blob and breaks test:local:merge.

frontend/package.json
aasimsyed:
```suggestion
    "test:local": "rm -rf blob-report && yarn test:local:desktop || true && yarn test:local:mobile || true && yarn test:local:merge",
```

frontend/package.json
aasimsyed:
```suggestion
    "test:local:desktop": "cross-env TEST_ENV=local PLAYWRIGHT_BLOB_OUTPUT_FILE=./blob-report/desktop.zip playwright test --project='Desktop Chrome' --reporter=blob",
```

frontend/test-e2e/page-objects/organization/about/OrganizationAboutPage.ts
```diff
@@ -61,4 +61,9 @@ export const newOrganizationAboutPage = (page: Page) => ({
   // MARK: Media
 
   imageCarousel: page.getByTestId("image-carousel"),
+  imageCarouselEditIcon: page
+    .getByTestId("image-carousel")
+    .getByTestId("icon-edit"),
+  getImageCarouselImages: page.getByTestId("image-carousel").getByRole("img"),
```
aasimsyed: blocker: this locator is unstable because `MediaImageCarousel.vue` sets `:loop="true"` on `<swiper-container>`, which clones slides for infinite scrolling — so `getByRole('img')` inside `image-carousel` returns real slides *plus* clones. That's the root cause of the CI failure `Expected: 1, Received: 3` at the spec's line 111.

Two options, I'd prefer (a):

(a) Add a stable test-id on the real `` inside `<swiper-slide>` in `frontend/app/components/media/image-carousel/MediaImageCarousel.vue` (e.g. `data-testid="image-carousel-image"`) and change this locator to:

    getImageCarouselImages: page.getByTestId("image-carousel-image"),

(b) Exclude clones at the locator:

    getImageCarouselImages: page
      .getByTestId("image-carousel")
      .locator("swiper-slide:not(.swiper-slide-duplicate) img"),

Option (a) is more maintainable and matches the `data-testid` pattern you're already using on the modal buttons.

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      // Wait for the modal to close and page to update.
+      await expect(organizationPage.uploadImageModal.modal).not.toBeVisible({
+        timeout: 10000,
+      });
+
+      // Verify the number of image in the carousel matches the number of files in the modal.
+      await expect(
+        organizationPage.aboutPage.getImageCarouselImages
+      ).toHaveCount(existingUploadEntriesCount + 1);
```
aasimsyed: blocker: this is asserting against the *modal's* image count (`existingUploadEntriesCount` from line 78) but checking the *carousel* locator. You already captured `initialCarouselCount` on line 49 — that's the right variable here, and line 141 already uses it correctly. Once the locator in `OrganizationAboutPage.ts` is fixed (see my other comment), this becomes a straight substitution.

change to: 

```suggestion
      // Verify the carousel grew by exactly one image.
      await expect(
        organizationPage.aboutPage.getImageCarouselImages
      ).toHaveCount(initialCarouselCount + 1);
```

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      // Wait for edit icon to be available (auth state should be loaded).
+      await expect(
+        organizationPage.aboutPage.imageCarouselEditIcon
+      ).toBeVisible({
+        timeout: 10000,
+      });
+
+      // Click the edit icon to open the image upload modal.
+      await organizationPage.aboutPage.imageCarouselEditIcon.click({
+        force: true,
```
aasimsyed: suggestion: `click({ force: true })` bypasses Playwright's actionability checks and is usually a smell — it hides overlay/timing issues that'll bite later. Can we replace with `await expect(...).toBeVisible(); await ...click();`? If `force: true` is genuinely required (e.g. because Swiper paints an overlay on top), please leave a one-line comment explaining why so future maintainers don't have to rediscover it.

also see lines: 114–116, 168–170

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+import { navigateToFirstOrganization } from "~/test-e2e/actions/navigation";
+import { expect, test } from "~/test-e2e/global-fixtures";
+import { newOrganizationPage } from "~/test-e2e/page-objects/organization/OrganizationPage";
+
+test.beforeEach(async ({ page }) => {
+  // Already authenticated via global storageState.
+  await navigateToFirstOrganization(page);
+
+  // Wait for auth state to be fully loaded.
+  await page.waitForLoadState("domcontentloaded");
```
aasimsyed: nit: the combination of `waitForLoadState("domcontentloaded")` + an `expect(async () => { document.readyState === "complete" }).toPass(...)` poll is much heavier than the rest of the suite, which just uses a single `waitForLoadState("networkidle"|"load")`. Suggest aligning:

```ts
test.beforeEach(async ({ page }) => {
  await navigateToFirstOrganization(page);
  await page.waitForLoadState("networkidle");
});

File: frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts [hunk 9]
```diff
@@ -0,0 +1,239 @@
...
+        );
+      await expect(imageUploadInput).toBeEnabled();
+
+      // Count initial number of files uploaded in the modal.
+      const existingUploadEntriesCount = await organizationPage.uploadImageModal
+        .getUploadedImages(organizationPage.uploadImageModal.modal)
+        .count();
+
+      // Set image input.
+      const createPngFile = (name: string) => ({
```
aasimsyed: duplicated PNG buffer
**Line:** 81–88 and 188–195

aasimsyed: suggestion: the base64 PNG is duplicated between the two tests. Extract it to a module-level const (or test-e2e/fixtures/images.ts) so we have one place to swap in a real fixture image later:
```suggestion
const TINY_PNG_BASE64 =
  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVR4nGP4z8AAAAMBAQDJ/pLvAAAAAElFTkSuQmCC";

const tinyPng = (name: string) => ({
  name,
  mimeType: "image/png",
  buffer: Buffer.from(TINY_PNG_BASE64, "base64"),
});```

File: frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts [hunk 10]
```diff
@@ -0,0 +1,239 @@
...
+        organizationPage.uploadImageModal.imageUploadInput(
+          organizationPage.uploadImageModal.modal
+        );
+      await expect(imageUploadInput).toBeEnabled();
+      await expect(imageUploadInput).toBeEditable();
+
+      // Count initial number of files uploaded in the modal.
+      const existingUploadEntries = await organizationPage.uploadImageModal
+        .getUploadedImages(organizationPage.uploadImageModal.modal)
+        .all();
```
aasimsyed: nit: .all() snapshots ElementHandles that are only used for .length. Prefer .count() — same value, cheaper, no dangling handles — and it matches how test 2 does it on line 183:
```suggestion
const existingUploadEntriesCount = await organizationPage.uploadImageModal
  .getUploadedImages(organizationPage.uploadImageModal.modal)
  .count();
```

aasimsyed:
praise: Thanks for pushing this through the rebase, @UnknownSean8. The test-id additions on the real components (`upload-image-upload-button`, `upload-image-remove-${index}`) and the page-object/component-object split are clean and match the project's patterns.

Requesting changes on three items:

blocker (1): `frontend/package.json` has two accidental regressions — `yarn` → `gbyarn` and a stray `--ui` on `test:local:desktop`. Both break `yarn test:local` for every developer. Please revert; `package.json` shouldn't change in this PR.

blocker (2): the first test fails in CI (`Expected: 1, Received: 3` at line 111) because (a) it asserts against `existingUploadEntriesCount` — the *modal* count — instead of `initialCarouselCount`, and (b) `getByTestId('image-carousel').getByRole('img')` counts Swiper's loop-mode clones, not just real slides. Fix both by adding a stable `data-testid="image-carousel-image"` on the real `` in `MediaImageCarousel.vue` and updating `getImageCarouselImages` to use it, then using `initialCarouselCount` for carousel assertions.

blocker (3): #1791 is closed but PR #1822 only fixed *upload* on the group modal, not *delete* on the org-about path this PR tests. Per our earlier thread: let's either reopen #1791 (or open a narrower bug) and let the delete assertion fail red with a TODO link, or `test.fixme` it until the app is fixed.

Smaller nits in inline comments.

frontend/package.json
UnknownSean8: ah yes! I forgot to remove it. Tho, I might suggest putting in documentation on how to test individual files.

frontend/test-e2e/page-objects/organization/about/OrganizationAboutPage.ts
UnknownSean8: Ahhhh, no wonder it keeps failing! Thanks for pointing this out. I do have one question on this as well, but I'll keep it below. I agree that option A is cleaner and more maintainable.

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
UnknownSean8: Ah yes, since this is the carousel check, which is already outside of the modal, I should be using the `initialCarouselCount` that is already captured before that is for the number of images outside of the modal, aka the carousel.

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
UnknownSean8: I rmb I copied the boilerplate beforeEach from some other tests, I will do more research on what it means. But I've already made the change to the commit.

UnknownSean8:
Praise (appreciation): Thank you so much, @aasimsyed, for the in-depth review. I don't have much experience writing e2e tests, and this helps a lot. I really learned a lot through these code reviews.

question (1): As part of the image fixtures, do you mean a playwright fixture like in the `global-fixtures.ts` file, or just a regular function export would be good enough?

question (2): I am a bit confused about the img clones, since it's img clones, wouldn't putting it inside the real `` tags also capture the clones as well? When I try to inspect the carousel, it seems that they are reorganising the img element instead.

question (3): The image seems to persist after the test is run. Is this an expected behaviour?

todo (1): open up #1791 or create a narrower bug to fix the error where _delete_ on images in the modal does not take effect on the carousel after the modal closes.

Thanks!

aasimsyed:
@UnknownSean8 I opened this PR related to the refresh issue and the swiper issue: https://github.com/activist-org/activist/pull/2109

aasimsyed:
> question (2): I am a bit confused about the img clones, since it's img clones, wouldn't putting it inside the real `` tags also capture the clones as well? When I try to inspect the carousel, it seems that they are reorganising the img element instead.

> question (2): I am a bit confused about the img clones...

> question (2): I am a bit confused about the img clones...

Hmm, you have a point. I had to re-check this after you brought it up. You're right that if the testid were on the `` itself, it wouldn't really help: Swiper's `loop` mode clones the whole slide subtree (children included), so the testid would just ride along to the duplicates too.

Poking around the DOM a bit more, I think two things are happening at the same time, which is probably why inspecting it felt weird:

1. Swiper adds *extra* slide copies at the head/tail (it tags them `.swiper-slide-duplicate`) so the loop feels seamless.
2. On every transition it also shuffles the live slides around — that's likely the "reorganising" you were seeing, not an actual replacement.

I initially thought about a test-side workaround of filtering out `.swiper-slide-duplicate`, but thinking about it more that felt a bit fragile: it couples our tests to Swiper's internal class names, so a library update could quietly break things. So I went with a slightly more robust approach I added to #2109 instead: the carousel now exposes a `:data-slide-count="imageUrls?.length ?? 0"` attribute on its container, which reads straight from the component's own prop and skips the clone/reorder noise entirely.

Once #2109 gets merged, the assertion becomes something like:

```ts
await expect(
  page.getByTestId("image-carousel-main")
).toHaveAttribute("data-slide-count", String(initialCarouselCount + 1));
```

aasimsyed:
todo: @UnknownSean8 heads-up on something you'll want to handle before merge... not a problem with your test logic, more a shared-state thing.

Your specs upload images to the seed `activist` org but nothing cleans them up between runs, so after ~10 consecutive `yarn test:local` runs the org hits `uploadLimit` and every upload assertion starts failing for an unrelated reason (fixture saturation, not a regression in your code). I hit this over and over while debugging #2108 and the only "fix" was purging rows via a Django shell, which obviously isn't the workflow we want.

The minimum needed is a small `beforeEach` inside `organization-about-image-upload.spec.ts` that purges images on the test org before each test. That keeps the cleanup contained inside this PR and doesn't couple it to anything else.

For context: I'd briefly opened #2110 to track this but closed it as premature since the symptom only manifests on this branch. Once this merges perhaps could ship a reusable helper as a small follow-up PR so the inline `beforeEach` becomes a one-liner.

UnknownSean8:
@aasimsyed thanks for the review, I git the `uploadLimit` over and over as well during tests and just delete them manually in the modal. But that definitely should not be the intended behaviour, will fix it by tmr.

Thanks for the refactor on the carousel. I will also attempt to create the reusable helper as well, but not in this patch.

UnknownSean8:
@aasimsyed Hi, I changed the `initialCarouselCount` to be 0 as the initial number of images in the carousel without any images is 3, which will cause problems later in the tests.

question (1): However, it can also be more explicit, where I remain the same `getImageCarouselImages.count()` (will get 3) and just adjust the numbers during the assertion.

question (2): Regarding the image purging before the test run, is there a function call to do this? Otherwise, writing it through a playwright test format doesn't seem robust. I was thinking of creating a `purgeImages` composable in `useOrganizationImageMutations`. Would that work, wdyt?

blocker (1): It seems like the fix on #2109 invalidates the carousel cache, but not in the modal. Images are not updated in the modal after file upload. However, this is not the case in file deletion. I did a quick check on `refreshOrganizationImagesData()`, and it seems to be in place.

I think the problem might be with the `draggable` in `ImageMultipleFileDropZone.vue`, not been updated. But you probably know more about this.

aasimsyed:
> @aasimsyed Hi, I changed the `initialCarouselCount` to be 0 as the initial number of images in the carousel without any images is 3, which will cause problems later in the tests.
> 
> question (1): However, it can also be more explicit, where I remain the same `getImageCarouselImages.count()` (will get 3) and just adjust the numbers during the assertion.
> 
> question (2): Regarding the image purging before the test run, is there a function call to do this? Otherwise, writing it through a playwright test format doesn't seem robust. I was thinking of creating a `purgeImages` composable in `useOrganizationImageMutations`. Would that work, wdyt?
> 
> blocker (1): It seems like the fix on #2109 invalidates the carousel cache, but not in the modal. Images are not updated in the modal after file upload. However, this is not the case in file deletion. I did a quick check on `refreshOrganizationImagesData()`, and it seems to be in place.
> 
> I think the problem might be with the `draggable` in `ImageMultipleFileDropZone.vue`, not been updated. But you probably know more about this.

question (1): If I understand what you're asking correctly, I would suggest neither hardcoding 0 nor adjusting the math around the clone count. Use the `data-slide-count` attribute that was added to the carousel container in #2109  it reads straight from the component prop and ignores Swiper's ghost slides completely. Read it right after the `beforeEach` purge so you always have the correct baseline regardless of leftover state:

```ts
const initialCarouselCount = parseInt(
  (await page
    .getByTestId("image-carousel-main")
    .getAttribute("data-slide-count")) ?? "0"
);
// after upload:
await expect(page.getByTestId("image-carousel-main")).toHaveAttribute(
  "data-slide-count",
  String(existingUploadEntriesCount + 1)
);
// after delete:
await expect(page.getByTestId("image-carousel-main")).toHaveAttribute(
  "data-slide-count",
  String(initialCarouselCount)
);
```

question (2): A Vue composable won't work here... composables run in the browser, and Playwright tests run in Node.js, so they can't talk to each other. It would also mean adding a "delete all images" function to the real app just for test cleanup, which isn't a good idea.

The robust way to do this I think is in Playwright is `page.request`, which makes HTTP calls from the test process using the same auth session the browser already has. No UI interaction, no new app code needed:

```ts
test.beforeEach(async ({ page }) => {
  const { organizationId } = await navigateToFirstOrganization(page);
  await page.waitForLoadState("networkidle");

  const res = await page.request.get(
    `/api/public/communities/organization/${organizationId}/images`
  );
  const images: { id: string }[] = await res.json();
  await Promise.all(
    images.map((img) =>
      page.request.delete(`/api/auth/content/images/${img.id}`)
    )
  );

  // Reload so the carousel reflects the purged state before tests read initialCarouselCount.
  await page.reload({ waitUntil: "networkidle" });
});
```

Note the routes are `/api/public/...` and `/api/auth/...` (not `/api/v2/...`) those are the Nuxt proxy paths.

blocker (1): You're right that `ImageMultipleFileDropZone.vue` is part of it, but there are actually three bugs combining to cause this. I've fixed them in #2118 since they're independent of this PR and cleaner to review.

UnknownSean8:
> question (1): If I understand what you're asking correctly, I would suggest neither hardcoding 0 nor adjusting the math around the clone count. Use the `data-slide-count` attribute that was added to the carousel container in #2109 it reads straight from the component prop and ignores Swiper's ghost slides completely. Read it right after the `beforeEach` purge so you always have the correct baseline regardless of leftover state:

Ah, apologies for the confusion, it's not about clones. If we go to the page in the initial state without any images uploaded, there are 3 placeholder images that are there. Once a user uploads an/multiple images, the 3 placeholder images will be gone.

```ts
// initial state = 3 (due to the presence of 3 image placeholders)
      const initialCarouselCount = parseInt(
        (await page
          .getByTestId("image-carousel-main")
          .getAttribute("data-slide-count")) ?? "0"
      );

// After an image uploads, the placeholders are gone, and the only image is the one we uploaded.
// So, expects 1 instead of initialCarouselCount + 1
      await expect(page.getByTestId("image-carousel-main")).toHaveAttribute(
        "data-slide-count",
        String(1)
      );

// After image deletion, our uploaded image is gone, and the 3 image placeholders return.
      await expect(page.getByTestId("image-carousel-main")).toHaveAttribute(
        "data-slide-count",
        String(initialCarouselCount)
      );
```

question (1): I've put in the `beforeEach`, it works nicely thanks. Is there a difference in using `beforeEach` in this case? I would think that `afterEach` is better as it isolates these tests in `organization-about-image-upload.spec.ts` from other tests.

aasimsyed:
@UnknownSean8 you were right about the placeholder count I checked the source and see that
`MediaImageCarouselFull.vue` falls back to `defaultImageUrls` (3 placeholder images) when `images.length === 0`,
and `data-slide-count` is bound directly to that `imageUrls` array length. So after a purge the attribute reads
`"3"`, not `"0"`. Your logic was correct:

- After purge: `initialCarouselCount = 3`
- After uploading 1 real image, placeholders are replaced: assert `"1"`
- After deleting that image, placeholders return: assert `String(initialCarouselCount)` -> `"3"`

Sorry for the confusion in my earlier reply.

**On the CI failure (`Expected: "1", Received: "3"` at line 105)**

The carousel stays at `"3"` because the upload never actually goes through — the modal shows an empty file list
after the file is staged, so nothing gets sent to the server. This is the exact app bug fixed in #2118 (stale
`orgId` key + Vue watch getter + concat deduplication). Your test logic is correct; it's blocked on #2118
merging first.

**On `beforeEach` vs `afterEach`**

`beforeEach` is the right choice here. `afterEach` only cleans up *after* a test, so a crashed run, a retry, or
any external session that left images behind would still dirty the org before the next test reads
`initialCarouselCount`. `beforeEach` guarantees a known empty baseline at the *start* of every run, which is
what stable count assertions require. You can add `afterEach` on top if you want to leave things tidy for
manual inspection, but it doesn't replace `beforeEach`.

UnknownSean8:
Got it @aasimsyed, thanks! I rebased and also refactored to include `afterEach`.

frontend/test-e2e/component-objects/UploadImageModal.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const newUploadImageModal = (page: Page) => ({
+  modal: page.locator("#modal").first(),
+  closeButton: (modal: Locator) => modal.getByTestId("modal-close-button"),
+  imageUploadInput: (modal: Locator) => modal.locator("input").first(),
+  uploadButton: (modal: Locator) =>
+    modal.getByTestId("upload-image-upload-button"),
+  removeButton: (modal: Locator, index: number) =>
+    modal.getByTestId(`upload-image-remove-${index}`),
+  getUploadedImages: (modal: Locator) => modal.getByRole("img"),
```
aasimsyed:
```ts
// I think this is wrong: counts all <img> in the modal including thumbnails rendered from localFiles
// before update:modelValue has propagated to files.value
getUploadedImages: (modal: Locator) => modal.getByRole("img"),

// I think this is correct: matches only the remove buttons that are rendered in the same
// synchronous cycle as emitFiles(), guaranteeing files.value is updated
getUploadedImages: (modal: Locator) =>
  modal.locator('[data-testid^="upload-image-remove-"]'),
```

frontend/test-e2e/page-objects/organization/about/OrganizationAboutPage.ts
```diff
@@ -61,4 +61,9 @@ export const newOrganizationAboutPage = (page: Page) => ({
   // MARK: Media
 
   imageCarousel: page.getByTestId("image-carousel"),
+  imageCarouselEditIcon: page
+    .getByTestId("image-carousel")
+    .getByTestId("icon-edit"),
+  getImageCarouselImages: page.getByTestId("image-carousel-image"),
+  getImageCarouselBullets: page.locator("swiper-pagination-bullet"),
```
aasimsyed:
```ts
// I think this is wrong: "swiper-pagination-bullet" is an element selector, matches nothing
getImageCarouselBullets: page.locator("swiper-pagination-bullet"),

// Correct: ".swiper-pagination-bullet" is a CSS class selector
getImageCarouselBullets: page.locator(".swiper-pagination-bullet"),
```

aasimsyed:
I think there might also be a race condition that can cause test to be flaky. you could fix here: 

`frontend/app/components/image/ImageMultipleFileDropZone.vue`

make handleRemove `async` and `await` `removeFile`:

```ts
async function handleRemoveFile(file: FileUploadMix) {
  await removeFile(localFiles.value, file.data);
  emitFiles();
  emit("file-deleted", file);
}
```

aasimsyed:
Hey @UnknownSean8, figured out why the uploads were failing in CI. I tested this locally on my fork and found an issue.

The `filescan` service was never started in either E2E workflow - only `backend` and `db` were. Django's upload view calls `scan_uploads_and_rewind` before saving any file, which POSTs to `http://filescan:9101/scan`. With that container absent, every upload returned HTTP 400 (`"The file could not be scanned."`). The frontend catches that error silently and closes the modal anyway, so the test only showed the symptom (`data-slide-count` stuck at 3) rather than the actual cause.

The reason it passes locally is that your full local stack has the `filescan` service running alongside the rest of the containers. In CI only `backend` and `db` were ever started, so `http://filescan:9101` was unreachable from the moment the test ran.

I've opened a fix in #2124. Once that merges you should be able to rebase your branch on top of it and (I think) CI should pass without any further code changes on your end.

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      await imageUploadInput.setInputFiles(tinyPng("file.png"));
+
+      // New entry appears in the modal.
+      await expect(
+        organizationPage.uploadImageModal.getUploadedImages(
+          organizationPage.uploadImageModal.modal
+        )
+      ).toHaveCount(existingUploadEntriesCount + 1);
+
+      // Upload image.
```
aasimsyed, line 104: if you replace 

```ts
await organizationPage.uploadImageModal
  .uploadButton(organizationPage.uploadImageModal.modal)
  .click();
```

with the following could be useful (this is what helped diagnose the issue in the first place, so it's a useful permanent guard to have):

```ts
// Register the listener before clicking so it captures the response
// even if the upload completes before the await below.
const uploadResponsePromise = page.waitForResponse(
  (res) =>
    res.request().method() === "POST" &&
    res.url().includes("/content/images"),
  { timeout: 15000 }
);
await organizationPage.uploadImageModal
  .uploadButton(organizationPage.uploadImageModal.modal)
  .click();

// uploadImages() catches server errors internally and closes the modal
// regardless, so without this check a failed upload is invisible and the
// test would only fail later at the carousel assertion with no useful context.
const uploadResponse = await uploadResponsePromise;
const uploadStatus = uploadResponse.status();
const uploadBody = await uploadResponse.text().catch(() => "<unreadable>");
expect(
  uploadResponse.ok(),
  `Upload failed: HTTP ${uploadStatus} — ${uploadBody}`
).toBeTruthy();
```

UnknownSean8:
> Hey @UnknownSean8, figured out why the uploads were failing in CI. I tested this locally on my fork and found an issue.

Thanks, I was actually looking at it and can't find the issue as it's passing everything on my end.

frontend/test-e2e/specs/all/organizations/organization-about/organization-about-image-upload.spec.ts
UnknownSean8: I assume filescan is not limited to just images, maybe this guard could be a helper function later on?

UnknownSean8: Maybe `test-e2e/utils/file-upload-helpers.ts` is somewhere I can put it in?

aasimsyed: Yes, you're right. The filescan service is invoked for any file type, so the guard is reusable wherever the tests do uploads, and `test-e2e/utils/file-upload-helpers.ts` is the right place; it fits the pattern of `faq-helpers.ts` and `modal-helpers.ts` that already live there.

andrewtavis:
Ping @UnknownSean8 👋 #2124 has been brought in, so the hope would be that we're good to go here now. Assuming tests pass, we'll do a final review and merge this in 😊

andrewtavis:
@aasimsyed, @UnknownSean8: Looks like we still are having the same issues on the Organization About page. Thoughts on next steps to get them passing?

aasimsyed:
@andrewtavis I think it's making progress, since it's not failing on the CI for the filescan issue anymore. Looking into current fails.

aasimsyed:
Hey @UnknownSean8, need to walk back an earlier suggestion and clarify what I think is actually failing. Before PR #2124, every upload in CI was silently returning HTTP 400 (missing `filescan` service), the modal was closing anyway because `uploadImages()` swallows errors, and the carousel was stuck at the placeholder count of 3. This masked what was really broken. Now that uploads succeed, the actual issue is clearer.

**correction: drop `uploadResponsePromise`  this is the only real blocker in test 1**

Looking at the current code, `uploadResponsePromise` is called *after* `await expect(modal).not.toBeVisible()`. The upload response has already fired by then, so the listener is registered too late and it times out every time at 15 s. That timeout is why test 1 fails - the carousel assertion (`String(1)`) never even runs.

The modal is the right gate: `handleUpload` only calls `closeModal()` after `uploadImages` has awaited both the server POST and the data refresh, so modal closure = server ack. Just drop the helper entirely:

```ts
await organizationPage.uploadImageModal
  .uploadButton(organizationPage.uploadImageModal.modal)
  .click();
await expect(organizationPage.uploadImageModal.modal).not.toBeVisible({
  timeout: 10000,
});
```

Delete the import and `file-upload-helpers.ts` (no other callers). 

**on the carousel assertion in test 1**

The hard-coded `String(1)` on line 120 works today because `purgeImages` guarantees a clean slate, but it's brittle - replace it with `String(existingUploadEntriesCount + 1)` so it's explicit about what it's asserting. Test 2's `String(existingUploadEntriesCount + filePng.length)` is already correct and should be passing after the filescan fix.

One note on `initialCarouselCount` (my earlier suggestion): I confirmed locally that after `purgeImages`, `data-slide-count` reads **3** not 0, because `MediaImageCarouselFull.vue` falls back to placeholder slides when there are no real images. So don't use `initialCarouselCount` for post-upload assertions - keep it only for the post-delete restore assertion in test 1 where it correctly reflects the pre-upload state.

UnknownSean8:
Sounds good, I'll be back tonight so I will try to rebase ltr today or tmr. Thanks!

andrewtavis:
Safe travels, @UnknownSean8!

aasimsyed (approved):

UnknownSean8:
Phew, such a long PR hahaha, thanks @aasimsyed along the way. @andrewtavis I think this is ready to merge. Thanks!

andrewtavis (approved):
praise: Thanks for the persistence here not just in implementation, @UnknownSean8, but also in reviewing, @aasimsyed! Glad to have this finished up :)

PR #1705, opened 2025-11-11
Add new tsconfig and update import statements

andrewtavis:
### Contributor checklist

---

### Description

Various fixes for Vitest imports based on the work that's been done for Playwright.

### Related issue

- No issue

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -1,2 +1,3 @@
 tsconfig.playwright.tsbuildinfo
+tsconfig.vitest.tsbuildinfo
 coverage/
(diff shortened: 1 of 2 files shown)


andrewtavis:
Looking into this with @nicki182 right now :) All's working as far as imports are concerned, but we have lots of TS errors in the Vitest files now. It'd be great if we could figure out how to initialize `globalThis` such that we don't have as many warnings/errors in those file :)

CC @aasimsyed for the conversation here. Would be great if we made an issue and looked into this 😊

aasimsyed:
@andrewtavis I was still getting "cannot find module" on the imports with this merge, so I am adding back the includes globs (different than the ones you had earlier) which seems to fix it.

nicki182:
@aasimsyed wait a second, did you delete node_modules and .nuxt and did a yarn install?

Issue #1739, opened 2025-11-22
Improve auto importing within Vitest suite


andrewtavis:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

As of now the setup for importing global functions and classes to the Vitest suite is not working properly such that we have lots of errors. Each time we add something new we also need to add lots of mocks and types to make sure that individual tests will run rather than a reactive solution to changes that are happening in the frontend.

This issue is to investigate this part of the testing infrastructure and improve it :)

andrewtavis:
CC @nicki182 and @aasimsyed 👋 @aasimsyed, would you have interest in working on this? If you need more context on this issue @nicki182 would be happy to support :)

aasimsyed:
hey @andrewtavis, thanks, I'm interested in working on this. @nicki182 would love any context you can offer to get started.

PR #1751, opened 2025-11-24
Fix/improve vitest auto importing

aasimsyed:
### Contributor checklist

---

### Related issue

- closes https://github.com/activist-org/activist/issues/1739

### DIFF
--- a/frontend/test/auto-imports.ts
+++ b/frontend/test/auto-imports.ts
@@ -6,35 +6,33 @@
  * for all use* composables that aren't already mocked.
  */
 
-/// <reference types="node" />
-
-import { readFileSync, existsSync } from "fs";
-import { join, dirname } from "path";
+import { existsSync, readFileSync } from "fs";
+import { dirname, join } from "path";
 import { fileURLToPath } from "url";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 /**
- * Extracts all use* composable names from .nuxt/imports.d.ts
+ * Extracts all use* composable names from .nuxt/imports.d.ts.
  */
 function extractComposableNames(importsContent: string): Set<string> {
   const composables = new Set<string>();
 
-  // Match export statements: export { useSomething, useOther } from '...'
-  // Also matches: export { useSomething as alias } from '...'
+  // Match export statements: export { useSomething, useOther } from '...'.
+  // Also matches: export { useSomething as alias } from '...'.
   const exportPattern = /export\s+{\s*([^}]+)\s*}\s+from/g;
   let match;
 
   while ((match = exportPattern.exec(importsContent)) !== null) {
     const exports = match[1];
-    // Split by comma and extract function names
+    // Split by comma and extract function names.
     const functions = exports
       .split(",")
       .map((f) => f.trim())
-      // Filter to only use* functions (use followed by uppercase letter)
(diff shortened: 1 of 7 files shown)


andrewtavis:
Thanks for this, @aasimsyed! Let's try to bring this in soon :)

nicki182:
@aasimsyed can you do an example refactor of a file with your changes implemented? because I see the setup but I want to understand how the implementation would be with the composables.

aasimsyed:
@nicki182 

Refactored existing test files to demonstrate all five patterns for using
auto-import mocks in Vitest tests. Each pattern is documented with
comments explaining when and why to use it.

Pattern 1: Use auto-mock without override (ModalBase.spec.ts)
- Use the default auto-mock from setupAutoImportMocks() as-is
- No manual mocking needed - auto-mock provides sensible defaults
- Example: useRoute() returns { params: {}, query: {} }

Pattern 2: Override with custom behavior (SidebarLeft.spec.ts)
- Override auto-mock with test-specific return values
- Custom mocks take precedence over auto-generated defaults
- Example: useRouter() needs specific currentRoute.name for tests

Pattern 3: Override for call tracking (BtnRouteInternal.spec.ts)
- Override auto-mock with spy function (vi.fn()) to track calls
- Verify composable was called with correct arguments
- Example: useLocalePath() needs call verification in tests

Pattern 4: Use mockImplementation() on setup spy (useColor.spec.ts)
- Change return values per test using mockImplementation()
- Composable must be mocked as spy function (vi.fn()) in setup.ts
- Example: useColorMode() returns different values per test

Pattern 5: Override in beforeEach with defaults, then per-test (http.spec.ts)
- Set default mock in beforeEach() for most tests
- Override within individual tests for specific scenarios
- Example: useAuth() has default in beforeEach, overridden for error cases

nicki182:
Ok, then can you remove the rest of the setup that default auto imports controls? Since by doing so we reduce just have this to make auto mocks for compostables. @aasimsyed If I understand right

aasimsyed:
@nicki182 The manual mocks provide test-specific behavior (real i18n instances, spy functions, proper data structures) that the auto-import utility can't handle: it only provides default empty object mocks. The two systems work together: manual mocks = custom behavior, auto-import = preventing `ReferenceError` errors and covering the ~296 other composables that don't need custom behavior.

We could refactor this into a cleaner factory-based system (centralize mock definitions in config files, extract factories for reusable patterns), which would reduce the setup.ts boilerplate and accomplish what you're asking about. That's a larger architectural change I think that would be better suited for its own focused PR rather than mixing it with the current auto-import improvement (if I am to understand the original issue intent). What do you think? Should we create a separate issue/PR for that or just do it in this PR?

nicki182:
@aasimsyed ok, I see. I was thinking of using more so the 4 pattern as a standard and refactor the tests so there is barely any setup in the setup file and for each test add the structure needed there. So it's localized for where it's needed. We could do your suggestion after this PR. What do you think @andrewtavis?

aasimsyed:
@nicki182 I created this issue based on the conversation. can you please take a look:  https://github.com/activist-org/activist/issues/1753

nicki182:
> @nicki182 I created this issue based on the conversation. can you please take a look: #1753

Looks fine to me

andrewtavis:
Plan above sounds good to me. Looking forward to checking out the setup once these issues are finalized 😊

PR #1754, opened 2025-11-25
test: Add comprehensive unit tests for all Pinia store entities

### LINKED ISSUE #1749: Unit tests: Test stores entities
### Terms

### Issue

Create unit tests for all stores for data entities. 
- event
- organization
- group
- topics

aasimsyed:
### Contributor checklist

---

### Description
This PR adds comprehensive unit tests for all Pinia stores that manage data entities.

### New Test Files
- `frontend/test/stores/helpers.ts` - Shared mock data factory functions
- `frontend/test/stores/event.spec.ts` - Tests for `useEventStore` (23 tests)
- `frontend/test/stores/organization.spec.ts` - Tests for `useOrganizationStore` (26 tests)
- `frontend/test/stores/group.spec.ts` - Tests for `useGroupStore` (26 tests)
- `frontend/test/stores/topics.spec.ts` - Tests for `useTopics` (9 tests)

### Related issue

- Closes #1749

### DIFF
--- a/frontend/test/stores/event.spec.ts
+++ b/frontend/test/stores/event.spec.ts
@@ -9,11 +9,12 @@ import { createMockEvent, createMockEventFilters } from "../mocks/factories";
 
 describe("useEventStore", () => {
   beforeEach(() => {
-    // Create fresh Pinia instance for each test to ensure isolation
+    // Create fresh Pinia instance for each test to ensure isolation.
     setActivePinia(createPinia());
   });
 
   // MARK: Initial State
+
   describe("Initial State", () => {
     it("initializes with null event", () => {
       const store = useEventStore();
@@ -37,6 +38,7 @@ describe("useEventStore", () => {
   });
 
   // MARK: Getter Actions
+
   describe("Getter Actions", () => {
     it("getEvent returns current event", () => {
       const store = useEventStore();
@@ -71,6 +73,7 @@ describe("useEventStore", () => {
   });
 
   // MARK: Setter Actions
+
   describe("Setter Actions", () => {
     it("setEvent updates event state", () => {
       const store = useEventStore();
@@ -110,6 +113,7 @@ describe("useEventStore", () => {
   });
 
   // MARK: Integration Tests
+
   describe("Integration Tests", () => {
     it("setting event then getting it returns the same event", () => {
       const store = useEventStore();
@@ -183,7 +187,7 @@ describe("useEventStore", () => {
       store.setPage(5);
       store.setFilters(mockFilters);
 
-      // Verify all state remains intact
+      // Verify all state remains intact.
       expect(store.getEvent()).toEqual(mockEvent);
       expect(store.getEvents()).toEqual(mockEvents);
       expect(store.getPage()).toBe(5);
@@ -192,6 +196,7 @@ describe("useEventStore", () => {
   });
 
   // MARK: Edge Cases
+
   describe("Edge Cases", () => {
     it("handles setting empty events array", () => {
       const store = useEventStore();
--- a/frontend/shared/types/map.d.ts
+++ b/frontend/shared/types/map.d.ts
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-
 export interface RouteProfile {
   profile: string;
   api: string;
--- a/frontend/test/mocks/data.ts
+++ b/frontend/test/mocks/data.ts
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-import { TopicEnum } from "../../shared/types/topics";
-
 /**
  * Default mock data constants for test factories.
  * These values are used as defaults when creating mock entities in tests.
  */
+import { TopicEnum } from "../../shared/types/topics";
 
 // MARK: ContentImage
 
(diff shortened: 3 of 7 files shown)


andrewtavis:
+1 that I'm always in favor of "utils" over "helpers".

aasimsyed:
Thanks for the feedback, @nicki182!

I agree about making the mock factories more discoverable. The `helpers.ts` file contains factory functions for creating mock data (e.g., `createMockEvent`, `createMockTopic`, `createMockOrganization`, etc.) that would be useful for other tests beyond just the store tests.

I've reorganized the structure, let me know if this is OK:

### New Structure
```
frontend/test/
  mocks/
    factories.ts
  stores/
    event.spec.ts 
    organization.spec.ts
    group.spec.ts
    topics.spec.ts
```

nicki182:
@aasimsyed I like that. The only change I would make is that I would move  the default data mocks to a different file called data.ts within mocks so you have the factories and the data on different files since I have a feeling this might get bigger with time and as initial format that would be good. what do you think?

aasimsyed:
Thanks for the suggestion, @nicki182! I've separated the default mock data into a `data.ts` file as you recommended.

## Changes Made

- **Created `frontend/test/mocks/data.ts`**: Contains all default mock data constants organized by entity type (ContentImage, User, PhysicalLocation, SocialLink, Event, Organization, Group, Topic)

- **Updated `frontend/test/mocks/factories.ts`**: Factory functions now import and use the default data from `data.ts`, keeping the factory logic separate from the data values

All existing tests continue to pass, and the API remains unchanged . Test files still import factory functions from `factories.ts`.

Let me know if this structure looks good!

nicki182:
@aasimsyed looks good to me. We can bring it in.

nicki182:
@aasimsyed the prettier is breaking and I don't understand the why. Since that file was not touched , can  you run your prettier locally and see if that happens to you? because once that is fixed we can bring this in

aasimsyed:
@nicki182 it passes for me locally.

aasimsyed:
@nicki182 the issue was a Prettier version mismatch. CI runs `sudo yarn add prettier`, which installs the latest (3.7.1) instead of the version in package.json. This creates a version mismatch. Not sure if we need that step since Prettier is already in package.json, so `yarn install` installs it.

Prettier 3.7.1 formats dynamic imports differently than Prettier 3.6.2:

3.6.2:
```ts
const { newEventPage } = await import(
  "~/test-e2e/page-objects/event/EventPage"
);
```

3.7.1:
```ts
const { newEventPage } =
  await import("~/test-e2e/page-objects/event/EventPage");
```

I upgraded Prettier to 3.7.1 and now it passes, but maybe the better solution would be to remove the redundant workflow step that installs latest Prettier.

nicki182:
@andrewtavis what do you think? I was thinking something similar yesterday when I saw the install of prettier separately. But I did not connect about prettier versions. @aasimsyed thank you for finding the problem. Maybe it's something we discuss tomorrow better in the dev sync and arrive to a solution

andrewtavis:
I'd say that in the future we just take it as an opportunity to upgrade Prettier? Same for Ruff in the backend? Makes sense to me that at least for formatters we do the latest version. Not necessarily for linters, but formatters update away :)

andrewtavis:
Updating it in the PR that a fail is occurring in as is done here makes total sense to me as well. No need to open something new :)

nicki182:
@andrewtavis so let's merge this in? Since I think with this PR fixes the issue

andrewtavis:
Will merge it in momentarily :)

frontend/test/stores/event.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    it("can set page to 0", () => {
+      const store = useEventStore();
+      store.setPage(5);
+      store.setPage(0);
+      expect(store.page).toBe(0);
+      expect(store.getPage()).toBe(0);
+    });
+
+    it("can set page to negative number", () => {
```
andrewtavis, line 224: question: Just checking my understanding here: Why are we checking `-1` as a page? Would this be if we have something like a `last` option at some point?

Merging regardless and we can remove if need be :)

andrewtavis:
Updated commit description to close the issue per the discussion above :)

andrewtavis (approved):
praise: Thanks for the great work, @aasimsyed, and thanks to you and @nicki182 for the great discussion on structure here 😊

frontend/test/stores/event.spec.ts
aasimsyed: @andrewtavis it's not for a "last" option - this is just an edge case test documenting the store's current behavior. The `setPage()` method accepts any number without validation, including negative values.

In practice, pages start at 1 (as seen in `useGetEvents` and `useGetOrganizations`), but the store itself doesn't enforce this. The test is documenting that invalid inputs like `-1` are currently accepted without any validation.

This could be something to improve later by either:
1. Adding validation in the store (e.g., `setPage` should only accept `>= 1`)
2. Removing this test if we decide to add validation

For now, it's just documenting the current behavior rather than testing intentional functionality. Happy to remove it or we can add validation as a follow-up.

andrewtavis: Maybe making the issue for adding the validation and removing those tests would make sense, @aasimsyed :) Would you want to make the issue?

andrewtavis: Seeing now that you already made and closed the issue, @aasimsyed :) Thanks so much!

Issue #1758, opened 2025-11-26
Bug: Topic filters disappear from URL after selection


aasimsyed:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

## Description

When filtering by topics on the Events or Organizations pages, the selected topics do not persist in the URL. After selecting a topic (or multiple topics), the URL briefly shows the topic parameter but then immediately reverts, removing all topic filters.

This affects:
- **Events page**: Both the header topics combobox and the sidebar topics filter
- **Organizations page**: The header topics combobox

Both pages use the same `ComboboxTopics` component in their headers, which is where the primary bug originates.

**Note:** This issue was primarily observed and reproducible through automated Playwright end-to-end tests. The timing-sensitive nature of the bug (URL briefly showing the correct value before reverting) made it difficult to observe consistently during manual testing, but automated tests reliably caught the issue.

### On Events Page:
1. Navigate to the Events page (`/events`)
2. Open the left sidebar if it's not already open
3. Click on the "Topics" filter dropdown in the sidebar
4. Select a topic (e.g., "Environment")
5. Observe the URL bar

### On Organizations Page:
1. Navigate to the Organizations page (`/organizations`)
2. Click on the "Topics" filter dropdown in the header
3. Select a topic (e.g., "Environment")
4. Observe the URL bar

**Note:** The same issue occurs on both pages because they share the same header `ComboboxTopics` component. The Events page also has the issue in the sidebar filter.

## Expected Behavior

- After selecting a topic, the URL should update to include the topic parameter (e.g., `?topics=ENVIRONMENT&view=map`)
- The topic parameter should remain in the URL and not disappear
- The selected topic should persist when:
  - Navigating between pages
  - Refreshing the page
  - Selecting additional topics
- Multiple topic selections should all appear in the URL (e.g., `?topics=ENVIRONMENT&topics=HEALTH&view=map`)

## Actual Behavior

- The URL briefly shows the topic parameter after selection
- The URL then immediately reverts, removing the topic parameter
- The URL ends up as just `?view=map` without any topics
- This happens consistently every time a topic is selected
- The issue affects both single and multiple topic selections

## Additional Issues

- Sometimes the topics combobox dropdown does not open when clicking on it, especially after a page refresh

## Environment

- Browser: All browsers
[...]

PR #1761, opened 2025-11-26
Fix: Topic filtering URL persistence, query parameter handling, and combobox compatibility

aasimsyed:
### Contributor checklist

---
## Problem

Topic filters were not persisting in the URL on Events and Organizations pages. After selecting a topic, the URL would briefly show the topic parameter but then immediately revert, removing all topic filters.

Additionally, refreshing a page with a topic filter resulted in "No events found" and an error: `s.forEach is not a function`.

## Root Cause

1. **Header ComboboxTopics circular update**: The header `ComboboxTopics` component was emitting `update:selectedTopics` whenever `selectedTopics` changed, including when changes came from prop updates (route changes). This caused it to emit an empty array, which triggered route updates that cleared topics from the URL.

2. **Vue Router query parameter inconsistency**: Vue Router returns `route.query.topics` as a string when there's a single topic (e.g., `?topics=ENVIRONMENT`), but as an array when there are multiple topics (e.g., `?topics=ENVIRONMENT&topics=HEALTH`). The API service functions expect `topics` to always be an array and call `.forEach()` on it, which fails if `topics` is a string.

3. **FormSelectorCombobox custom input wrapper issues**:
   - Missing visible `ComboboxButton` to open the dropdown
   - `setSelectionRange` errors when using `as="div"` wrapper with custom input component

## Solution

1. **Header ComboboxTopics fix**: Added `isUpdatingFromProps` flag to prevent emitting updates during prop changes. The component now only emits when the user actually interacts with it.

2. **Topics query parameter normalization**: Added normalization logic in both `events/index.vue` and `organizations/index.vue` to ensure `route.query.topics` is always converted to an array. This handles Vue Router's inconsistent behavior:
   - If `topics` is already an array → use as-is
   - If `topics` is a string → wrap it in an array `[topics]`
   - This ensures the API service can always call `.forEach()` without errors

3. **FormSelectorCombobox fix**:
   - Added visible `ComboboxButton` with chevron icon for programmatic dropdown control
   - Added `setSelectionRange` forwarding workaround (Headless UI calls native input methods on wrapper div)
   - Fixed dropdown positioning with proper absolute positioning

### Related issue

- closes https://github.com/activist-org/activist/issues/1758

### DIFF
--- a/backend/authentication/tests/flag/test_user_flag_list.py
+++ b/backend/authentication/tests/flag/test_user_flag_list.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 import logging
+from unittest.mock import patch
 
 import pytest
 
@@ -19,3 +20,25 @@ def test_user_flag_list(authenticated_client):
 
     assert response.status_code == 200
     logger.info("User flag list test completed successfully")
+
+
+def test_user_flag_list_no_pagination(authenticated_client):
+    """
+    Test list all user flags in case of no pagination.
+    """
+
+    client, user = authenticated_client
+
+    with patch(
+        "authentication.views.UserFlagAPIView.paginate_queryset"
+    ) as mock_paginate:
+        mock_paginate.return_value = None
+
+        logger.debug("Making request to user flag endpoint")
+        response = client.get(path="/v1/auth/user_flags")
+
+        assert response.status_code == 200
+
+        # Verify that paginate_queryset was called.
+        mock_paginate.assert_called_once()
+        logger.info("User flag list test completed successfully")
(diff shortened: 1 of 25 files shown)


nicki182:
@andrewtavis and @aasimsyed  I'll review this tomorrow

nicki182:
note: I made some changes to simplify the solution, I prefer to go towards simpler solutions and changed the naming convention of a function you added towards a more generic name since I see it helps with url queries that has arrays within it. And I saw a disables es-lint rule for an any, if you think there is no other solution than using it then ok.  You can review the changes I made too. 
praise: this is pretty good, I think we are done and we can merge in.

aasimsyed:
Thanks for simplifying the solution, @nicki182! The `normalizeArrayFromURLQuery` name is much more generic and reusable. 

I've reviewed your changes and added a sync guard to prevent infinite loops between watchers, ensuring empty arrays are properly emitted to clear URL query params when users deselect all options.

Do you think it still looks good to merge? 🚀

nicki182:
praise:@aasimsyed I see what you mean and this looks good to me. So, yeah for me it's good to merge.

nicki182:
note: @andrewtavis I brought this in since it's ok for merging

Issue #1712, opened 2025-11-11
Bug: “Days ahead” filter only shows events on the exact day


aasimsyed:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

## Summary
When I pick a “Days ahead” filter (1/7/30) on the events page, it only shows events happening exactly on that future day. I expect it to show everything coming up between now and that day instead.

## Steps to Reproduce
1. Visit the events list and open the sidebar filters.
2. Choose the “7 days ahead” option.
3. Notice that events happening tomorrow or three days out disappear, and only events occurring seven days from now remain.

## Actual Behaviour
- “Days ahead” filters behave like single-day lookups; the list only shows events on that exact future date.

## Expected Behaviour
- “7 days ahead” should include all upcoming events from now through day 7 inclusive (same for 1 and 30 days).

## Acceptance Criteria
- [ ] Update the “Days ahead” filter to treat the selected value as an upper-bound range starting from today.
- [ ] Verify that 1/7/30 selections include all events between now and the chosen day.
- [ ] Add regression tests so the filter keeps working as a range.

Chuksll:
Hello @andrewtavis my team and I would like to work on this issue.
Here's the rest of my team:
- mfan59
- a-estrada
- lhazard434

andrewtavis:
Sure thing, @Chuksll! One thing to note is that each of you needs to write into this issue and #1711 so I can assign. You can't assign people without triage rights to the repo unless they write into the issue.

Please let us know if you need any support!

mfan59:
Hi! I'm @Chuksll's teammate, and I would like to be assigned to this issue, please! Thanks :)

a-estrada:
Hi, adding to above, I am a teammate of @Chuksll and @mfan59 and would like to be assigned to this issue as well. Thanks!

SuperHappycat:
Can I join too? I think I know how to fix it.
```python
#SPDX-License-Identifier: AGPL-3.0-or-later
"""
A class for filtering events based on user defined properties.
"""

from datetime import date, datetime, timedelta
from typing import Any, Union

import django_filters
from django.db.models.query import QuerySet

from content.models import Topic
from events.models import Event


class EventFilters(django_filters.FilterSet):  # type: ignore[misc]
    """
[...]

PR #1773, opened 2025-11-30
fix: implement proper days_ahead range filtering

a-estrada:
### Contributor checklist

---

#### Summary

Replace the `active_on` date-based event filter with a `days_ahead` range filter, add backend tests for its behavior, and align the frontend filters, types, error handling, and end-to-end tests with the new semantics.

##### Backend

- Replace `active_on` with `days_ahead` on `EventFilters`:
  - Introduce a `filter_days_ahead` helper that:
    - Uses `timezone.now()` as the lower bound.
    - Interprets the upper bound as the end of the day `days` days from now (inclusive).
    - Safely handles non-integer values by returning an empty queryset.
  - Remove the old `active_on` filter and associated method to avoid overlapping or confusing filter semantics.
  - Expose `days_ahead` in `Meta.fields` so it is available via query params.

- Add focused filter tests for `days_ahead`:
  - Events starting in the past are excluded, and events beyond the configured window are excluded.
  - `days_ahead` works in combination with other filters (`type`, `setting`).
  - Non-positive values (`0` or negative) are accepted and behave predictably (only events starting at the current instant are included for `0`).

##### Frontend

- Update event filter sidebar wiring:
  - Map the `"days"` form control directly to a `days_ahead` query parameter.
  - Remove the client-side construction of an `active_on` timestamp and rely on the backend’s `days_ahead` range logic instead.

- Align event service with shared error handling:
  - Import and use the existing shared `errorHandler` from `#shared/utils/errorHandler` in `listEvents`.
  - Type the caught error as `unknown` so the handler is responsible for normalization.

- Extend and simplify event filter types:
  - Remove `active_on?: string` from `EventFilters`.
  - Add `days_ahead?: string` so the new filter is consistently typed across the frontend.

- Improve error messaging for invalid pages:
  - In the shared `errorHandler`, detect a `404` with `detail: "Invalid page."` and surface it as `"Invalid page number."`.
  - Leave other HTTP errors and payload shapes handled by the existing normalization logic.

- Update desktop end-to-end tests for events filtering:
  - Assert that the URL now contains `days_ahead=7` instead of an `active_on` date when applying the “7 days ahead” filter together with type and location filters.
  - Ensure state persistence checks also validate `days_ahead=7` after a page reload.
  - Remove client-side expectations tied to a specific `active_on` date, since date-range semantics now live entirely on the backend.

### Files changed

- `backend/events/filters.py`
  - Replace the `active_on` filter with a `days_ahead` filter and include it in `Meta.fields`.
  - Implement `filter_days_ahead` using `timezone.now()` and an inclusive upper bound at the end of the target day.

- `backend/events/tests/test_event_filters.py`
  - New test module covering `days_ahead` window behavior, interactions with `type` and `setting`, and non-positive inputs.

- `frontend/app/components/sidebar/left/filter/SidebarLeftFilterEvents.vue`
  - Wire the “Days ahead” option to send `days_ahead` instead of constructing an `active_on` timestamp on the client.

- `frontend/app/services/event/event.ts`
  - Import the shared `errorHandler` utility and delegate `listEvents` error normalization to it.

- `frontend/shared/types/event.d.ts`
  - Replace `active_on?: string` with `days_ahead?: string` on `EventFilters`.

- `frontend/shared/utils/errorHandler.ts`
  - Map backend `404` errors with `detail: "Invalid page."` to a clearer `"Invalid page number."` message while preserving the existing cause and details.

- `frontend/test-e2e/specs/desktop/events-filter.spec.ts`
  - Update expectations to assert `days_ahead=7` in the URL for combined filter scenarios and on refresh, removing `active_on`-specific checks.

---

### Related issue

- closes https://github.com/activist-org/activist/issues/1712

### DIFF
--- a/backend/events/filters.py
+++ b/backend/events/filters.py
@@ -46,10 +46,10 @@ class EventFilters(django_filters.FilterSet):  # type: ignore[misc]
     )
 
     def filter_days_ahead(
-        self, queryset: QuerySet[Any, Any], name: str, days_ahead: int
+        self, queryset: QuerySet[Any, Any], name: str, days: int
     ) -> QuerySet[Any, Any]:
         """
-        Filter events occurring within the next ``days_ahead`` days as a rolling window.
+        Filter events occurring within the next ``days`` days as a rolling window.
 
         Parameters
         ----------
@@ -57,22 +57,23 @@ class EventFilters(django_filters.FilterSet):  # type: ignore[misc]
             Base queryset.
 
         name : str
-            Filter field name (``days_ahead``).
+            Filter field name (``days``).
 
-        days_ahead : int
+        days : int
             Number of days into the future.
 
         Returns
         -------
         QuerySet[Any, Any]
-            Events starting between ``now`` and ``now + days_ahead`` (inclusive).
+            Events starting between ``now`` and ``now + days`` (inclusive).
         """
         now = timezone.now()
 
-        if not days_ahead:
-            return queryset.none()
+        try:
+            days_ahead_int = int(days)
 
-        days_ahead_int = int(days_ahead)
+        except Exception:
+            return queryset.none()
 
         if days_ahead_int < 0:
             return queryset.none()
--- a/backend/events/tests/test_event_filters.py
+++ b/backend/events/tests/test_event_filters.py
@@ -15,7 +15,7 @@ from events.factories import EventFactory
 pytestmark = pytest.mark.django_db
 
 EVENTS_URL = "/v1/events/events"
-FIXED_NOW = datetime(2025, 11, 20, 12, 0, tzinfo=dt_timezone.utc)
+FIXED_NOW = datetime(1970, 1, 1, 0, 0, tzinfo=dt_timezone.utc)
 
 
 @patch("django.utils.timezone.now", return_value=FIXED_NOW)


frontend/app/services/event/event.ts
```diff
@@ -65,7 +65,21 @@ export async function listEvents(
       { withoutAuth: true }
     );
     return { data: res.results.map(mapEvent), isLastPage: !res.next };
-  } catch (e) {
+  } catch (e: unknown) {
+    const err = e as {
+      response?: { status?: number; data?: { detail?: string } };
+      statusCode?: number;
```
nicki182: note:I would rather you move this code to the error handler

frontend/app/services/event/event.ts
```diff
@@ -65,7 +65,21 @@ export async function listEvents(
...
+      status?: number;
+      data?: { detail?: string };
+    };
+
+    const status = err.response?.status ?? err.statusCode ?? err.status ?? null;
+    const detail = err.response?.data?.detail ?? err.data?.detail ?? undefined;
+
```
nicki182: note: I prefer if it broke since it means we are doing something wrong, and just show the error invalid page number, something like this

nicki182 (commented):
note: I like the change and it makes sense to call days_ahead but then remove everything regarding active_on because its confusing. If we are renaming a and changing a filter then you need to remove the other one since it will create confusion since it exists.
praise: thank you for your contribution. This is a very good one since you added the testing too from the backend of the filtering. I hope you implement the last thing and I will check if everything works ok. If it does then I will merge it in.

nicki182:
@andrewtavis anything else you want to add?

andrewtavis:
I agree with the comments above and don't have anything else to suggest at the current stage :) @a-estrada, would you be able to address @nicki182's comments above?

frontend/app/services/event/event.ts
a-estrada: Done – I moved the `404 "Invalid page."` handling into the shared `errorHandler` in `frontend/shared/utils/errorHandler.ts` and removed the special case from `listEvents`.

`listEvents` now just calls `errorHandler(e)` and the centralized handler turns that specific backend error into an `AppError`.

frontend/app/services/event/event.ts
a-estrada: Updated as suggested: we no longer swallow the error in `listEvents`.
The shared `errorHandler` now detects `status === 404` with `detail: "Invalid page."` and sets the message to `"Invalid page number."`, but still throws an `AppError`.

nicki182:
note: @a-estrada I just checked and it seems there is an issue with calculating the days ahead, when I put one day it gives me 2 days ahead. Can you revise and fix it?

a-estrada:
@nicki182 Thanks for flagging this.

The previous implementation effectively behaved like “from now until the end of day N,” but I’ve updated the filter to treat `days_ahead` as a rolling window starting at `timezone.now()` and ending at `now + N days`, and added a dedicated test to pin down that boundary.

This should better match the intuitive expectation for days ahead.

nicki182:
@andrewtavis what do you think Andrew? Because to me I put one day ahead and for example it's Thursday, it gives me until Saturday, not Friday.

PR #1819, opened 2026-01-04
fix(frontend): use floating-vue for tooltips #1818

### LINKED ISSUE #1818: TooltipBase instances are not appearing on hover
### Terms

### Behavior

The way that `.has-tooltip` is being applied in combination with `TooltipBase` in the frontend is now not working, potentially due to changes from Tailwind updates. We used to get a tooltip over the `/` and `^k` icons in the search bar based on the classes in `tailwind.css`, but this is now not working :)

aasimsyed:
This seems to fix the tooltip issues.

- Add floating-vue Nuxt plugin to enable v-tooltip directive
- Replace broken .has-tooltip + TooltipBase pattern with v-tooltip in:
  - SearchBar.vue (3 tooltips)
  - PageFilter.vue (3 tooltips)
  - IconOrganizationStatus.vue (2 tooltips)
- Remove unused .has-tooltip CSS rule from tailwind.css

I think the CSS-based tooltip approach broke in Tailwind v4 due to layer specificity changes (utilities > components). Additionally, tooltips were trapped behind stacking contexts from overflow and z-index.

floating-vue renders tooltips in a portal at body level, bypassing both CSS specificity and stacking context issues.

### Contributor checklist

---

### Related issue

- Closes #1818

### DIFF
--- a/frontend/test/composables/queries/helpers/useAsyncDataMock.ts
+++ b/frontend/test/composables/queries/helpers/useAsyncDataMock.ts
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+/**
+ * Helper utilities for mocking useAsyncData in integration tests.
+ * Allows capturing and testing the handler and getCachedData functions.
+ */
+import { vi } from "vitest";
+import { ref, type Ref } from "vue";
+
+/**
+ * Result type for captured useAsyncData call.
+ */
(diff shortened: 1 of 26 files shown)


nicki182:
@aasimsyed works perfectly and I like the solution since its a more elegant way. I was thinking of having to attach the component to be using the tooltip instead of the css since the original solution is more fickle so its great you did so already.

PR #1828, opened 2026-01-06
fix(frontend): prevent query param persistence on route change

### LINKED ISSUE #1738: Route queries should be reset when the user navigates to a new route
### Terms

### Behavior

When the user filters on the `/events` or `/organizations` routes and then switches to the other route via the menu option at the top of the left sidebar, then the filters are still being set within the URL. For instance:

- User filters by a topic using the topics combobox at the top right of the page on events or organizations
- The URL now has a filter for the topic selected
- The user switches to the other route via the sidebar route options below the search bar
- The same route is still being set

Ideally when switching a base route the filters would be reset.

aasimsyed:
Add path-aware watchers to filter components that only sync query params when on the correct route. This prevents stale query params from persisting when navigating between /events and /organizations.

- SidebarLeftFilterOrganization: only sync when path === "/organizations"
- SidebarLeftFilterEvents: only sync when path === "/events"
- Add 26 unit tests covering issue #1738 scenarios

Fixes #1738

### Contributor checklist

---

### Related issue

- #1738

### DIFF
--- a/frontend/app/middleware/clearQueryOnPathChange.global.ts
+++ b/frontend/app/middleware/clearQueryOnPathChange.global.ts
@@ -3,17 +3,17 @@
  * Clear query params when the route path changes.
  * Prevents stale filter params from persisting when navigating between
  * different sections (e.g. /events -> /organizations).
- * Fix for: https://github.com/activist-org/activist/issues/1738
+ * @see https://github.com/activist-org/activist/issues/1738
  */
 export default defineNuxtRouteMiddleware((to, from) => {
-  // Only run on client-side navigation (skip initial load/refresh)
+  // Only run on client-side navigation (skip initial load/refresh).
   if (!from.name) {
     return;
   }
 
   // Path changed and destination has query - clear to prevent cross-route pollution.
-  // Must check to.query (not from.query) to avoid infinite redirect: after we redirect,
-  // the next run has to.query empty, so we stop.
+  // Must check to.query (not from.query) to avoid infinite redirect.
+  // After we redirect, the next run has to.query empty, so we stop.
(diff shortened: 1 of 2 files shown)


nicki182:
question: @aasimsyed I imagine you checked more general solutions which where the url should automatically should flush all queries when its changing urls. This works perfectly but I wanted to double check if there is a more general solution, and if you did so, which ones did you check?

andrewtavis:
Quick note, @aasimsyed: It looks like your emails for Git and GitHub are out of sync :)

aasimsyed:
praise: @nicki182 Thanks for raising this. I hadn’t considered a more general approach at first, so your question helped me to think on that.

I think the right fix is a global middleware (`clearQueryOnPathChange.global.ts`) that clears all query params whenever the path changes (e.g. `/events` -> `/organizations`), instead of path-specific watchers in each filter component. That should be easier to maintain and will cover any future pages that use query params. I’m open to adjusting this if you have a different direction in mind.

aasimsyed:
> Quick note, @aasimsyed: It looks like your emails for Git and GitHub are out of sync :)

argh... new computer. how do i fix this again? I just edited my global config now, so hopefully that fixed?

andrewtavis:
Looks good now, @aasimsyed! Are we ready for a final review here? :)

aasimsyed:
yes, @andrewtavis this is ready now

andrewtavis:
From the review with @aasimsyed, since merging in #1922 there are now some issues with the functionality here. The URL parameters are being maintained now. We need the changes from the other PR to be synced with the functionality over here :)

Note from the conversation with @nicki182: We are fine with topics being removed when the route is changed 😊

Issue #1843, opened 2026-01-18
Switch auth in frontend from `nuxt-auth` to `nuxt-auth-utils`


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

We're having a persistent problem with [nuxt-auth](https://github.com/sidebase/nuxt-auth) as a dependency that `next-auth` is a requirement and is currently set to a version with vulnerabilities. In looking at this with @nicki182, the general feeling is that it would be better to attempt a migration to [nuxt-auth-utils](https://github.com/atinux/nuxt-auth-utils).

> [!NOTE]
> Closing this issue would close #1646.

nicki182:
will be doing so by using a proxy server with the backend which will help later for SSR implementation later on 😉

andrewtavis:
Updated what was easy as far as frontend dependencies, @nicki182. I'm hoping that the switch here will get rid of a bunch of old ones from the tree and then we'll be able to close most of the dependabot ones :)

nicki182:
update:
- login and logout work. 
- server has been set up
to do:
- sign up
- refresh token logic 
- all services pass through proxy server where the token will be retriven (once this is done there is going to be a lot of fixing to be done on the tests)
@andrewtavis

andrewtavis:
Great to hear the above, @nicki182! Really looking forward to the review and bringing in the changes :)

AadityaAnand:
Hi, is anyone working on this, or it still needs to be done?

andrewtavis:
Hi @AadityaAnand 👋 @nicki182 is assigned and is working on this :) Thanks for your interest in the project! Feel free to reach out in other issues that interest you if you see that there haven't been recent messages or if it's unassigned.

nicki182:
@andrewtavis a couple of services are working now, get topics and events and added new middleware logic to get token.
by tomorrow most services should work with new logic and with this in the long term it could work to add SSR since now we will fully have nuxt backend working mostly as a proxy server

nicki182:
its done in an sense. There is some things to be corrected. I wanted to ask for you to test out the functionality and tell me if everything works properly.  I need some help since this affects the whole app for services and we can have a call later on explaining what has been done 
@andrewtavis

andrewtavis:
[...]

PR #1846, opened 2026-01-18
Update dependencies that are able to pass eslint/typecheck

andrewtavis:
### Contributor checklist

---

### Description

This PR updates multiple dependencies now that we're mostly through updating the dev dependencies. Note that we're still not able to directly close any dependabot PRs as the versions are still too low for some subdependencies. Closing those PRs may need to wait till work on #1843 is finished.

### Related issue

- No related issue


frontend/package.json
andrewtavis, line 121: `vue-i18n` has higher versions, but we once again run into the `$t` TS errors when we update.

andrewtavis:
I'm going to test this a bit in the dev env and make sure that things like the discussion component can still work, and then I think we can merge this as well. From here, I think that we just need to hope that frontend dependencies get a bit cleaner after the work in #1843 is done.

andrewtavis:
We're having some issues with the discussion component, but this is because it's not tested/planned for the mvp. This is also the case on main. Under the circumstances we can bring this in, and we'll look into the issues with the discussion component when there's capacity.

Issue #1753, opened 2025-11-24
Refactor tests to use pattern-based mocks with minimal setup.ts


aasimsyed:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

# Refactor tests to use pattern-based mocks with minimal setup.ts

## Summary

Refactor tests so there is barely any setup in `test/setup.ts` and for each test add the structure needed there. Remove default mock implementations from `setup.ts` and have tests explicitly define their mock needs using reusable mock factories and the 5 established patterns. This directly addresses the reviewer's question: "can you remove the rest of the setup that default auto imports controls?"

## Current State

- ~90 lines of manual mock definitions in `test/setup.ts`
- Tests rely on hidden global defaults (e.g., `useRoute()` returns `{ params: {}, query: {} }` by default)
- Mock behavior is not explicit in tests - hard to see what each test actually needs
- Mixing of infrastructure setup and mock definitions

## Desired Behavior

**The Goal:** Minimal `setup.ts` with only infrastructure-dependent mocks, and tests explicitly define their mock needs using reusable mock factories.

**To Do:**

1. **Centralize mock factories** - Create reusable mock factory functions in `test/mocks.ts`
   - Example: `createUseRouteMock(params, query)`, `createUseAuthMock(user)`, etc.
   - Common mock patterns available as reusable functions
   - Easy to update behavior in one place

2. **Minimal setup.ts** - Only infrastructure-dependent mocks (like `useI18n` which needs i18n instance) and spy initialization where needed. No default implementations applied automatically.

3. **Tests explicitly import and use** - Each test explicitly imports mock factories and sets up what it needs:
   - Pattern 1: Use auto-mock as-is (if default `{}` from auto-import works)
   - Pattern 2: Import factory and use it (e.g., `globalThis.useRoute = createUseRouteMock({ id: "123" })`)
   - Pattern 3: Override with spy for call tracking
   - Pattern 4: Use `mockImplementation()` on spy functions
   - Pattern 5: Set defaults in `beforeEach` using factories, override per-test

4. **No automatic global defaults** - Tests must explicitly set up mocks using factories or inline implementations

## Example

**Mock Factory:**
```typescript
// test/mocks.ts
export const createUseRouteMock = (params = {}, query = {}) =>
[...]

PR #1903, opened 2026-02-17
issue 1783 mutation unit tests

### LINKED ISSUE #1783: Unit tests for mutations
### Terms

## Summary
Implement comprehensive unit tests for mutation composables in folder frontend/app/composables/mutations. 

Both use direct service calls and refresh Nuxt async/query data after mutations.

## Requirements
- Test each mutation handler: `createFAQ`, `updateFAQ`, `reorderFAQs`, `deleteFAQ`, `uploadIconImage`, `refreshEventData`, ext
- Assert correct mutation behavior, service call invocations, state updates, and error propagation.
- Mock all underlying service calls and cache refresh (`refreshNuxtData`). 
- Test error scenarios (service throws/returns error), check calls to `showToastError` and the correct reactive error state.
- Verify loading transitions (true/false states before/after mutation runs).
- Test edge cases: missing eventId or bad FAQ data, unexpected errors, handling for delete/add order, cache invalidation/refetch, ect
- Test that `refreshEventData` always calls refreshNuxtData with the correct key  after a mutation.
- Cover reactivity (readonly states) for `loading` and `error` throughout all mutation methods.

## Paths & code references`
- Service helpers. 

## Implementation guidelines
- Use Vitest/Jest and Vue/Nuxt test utils to mock service/composable dependencies and async behaviors
- Place unit tests under `frontend/testst/composables/mutations`
- Document any additional edge cases, helpers, or mutation patterns discovered during implementation
- Link to parent and related subissues for review

aasimsyed:
### Contributor checklist

---

## Unit tests for mutation composables (#1783)

Adds unit tests for all 15 mutation composables (Event, Group, Organization) and fixes two related bugs.

### Tests
- **230 tests** across FAQ, Image, Resources, SocialLinks, and Texts composables
- Success paths: service calls, `refreshNuxtData` with correct keys, return values
- Error paths: `error.value`, `showToastError`, no refresh on failure
- Edge cases: empty IDs, invalid input (bad FAQ/resource/link data)
- Loading transitions and readonly `loading`/`error` state

### Bug fixes
- **Event text service:** Use `event` instead of `orgId` in the request body
- **Mutation composables:** Set `error.value = appError` in catch blocks for consistent error handling

### References
- Uses #1753 mock patterns

Description of the proposed change...

### Related issue

- Closes #1783

### DIFF
--- a/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
+++ b/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
@@ -92,7 +92,7 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
 
       return true;
     } catch (err) {
-      const appError = errorHandler(err);
+      const appError = err as AppError;
       error.value = appError;
       showToastError(appError.message);
       return false;
--- a/frontend/app/composables/mutations/useEventImageIconMutations.ts
+++ b/frontend/app/composables/mutations/useEventImageIconMutations.ts
@@ -23,7 +23,7 @@ export function useEventImageIconMutations(eventId: MaybeRef<string>) {
 
       return true;
     } catch (err) {
-      const appError = errorHandler(err);
+      const appError = err as AppError;
       error.value = appError;
       showToastError(appError.message);
       return false;
--- a/frontend/app/composables/mutations/useEventResourcesMutations.ts
+++ b/frontend/app/composables/mutations/useEventResourcesMutations.ts
@@ -25,7 +25,7 @@ export function useEventResourcesMutations(eventId: MaybeRef<string>) {
 
       return true;
     } catch (err) {
-      const appError = errorHandler(err);
+      const appError = err as AppError;
       error.value = appError;
       showToastError(appError.message);
       return false;
--- a/frontend/app/composables/mutations/useEventSocialLinksMutations.ts
+++ b/frontend/app/composables/mutations/useEventSocialLinksMutations.ts
@@ -29,7 +29,7 @@ export function useEventSocialLinksMutations(eventId: MaybeRef<string>) {
 
       return true;
     } catch (err) {
-      const appError = errorHandler(err);
+      const appError = err as AppError;
       error.value = appError;
       showToastError(appError.message);
       return false;
--- a/frontend/app/composables/mutations/useEventTextsMutations.ts
+++ b/frontend/app/composables/mutations/useEventTextsMutations.ts
@@ -26,7 +26,7 @@ export function useEventTextsMutations(eventId: MaybeRef<string>) {
       await refreshEventData();
       return true;
     } catch (err) {
-      const appError = errorHandler(err);
+      const appError = err as AppError;
       error.value = appError;
       showToastError(appError.message);
       return false;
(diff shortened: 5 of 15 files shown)


nicki182 (commented):
@aasimsyed the tests looks great on the tests but my only issue is that in all mutations composables files it has:
const appError = errorHandler(err);
      error.value = appError;
      showToastError(appError.message);
added which within the services the error is already handled within. Now if you want to debate about it we can, but then you will or I will have to move all error handler to the composable layer and no in the services layer

aasimsyed:
Thanks for the review, @nicki182. I want to make sure I’ve understood your feedback correctly.

Forgive me if I’m wrong, but I had understood the flow as:

1. **Services** catch the raw error, normalize it with `errorHandler`, and rethrow it. They don’t show UI or set state.
2. **Composables** catch that thrown error and handle it for the user (show toast, set `error.value`, return false).

If that’s right, I added `error.value = appError` because some composables weren’t setting it and the tests expect it. I also called `errorHandler(err)` in the composable for safety, in case something other than an `AppError` is thrown. But I may be misreading the intended split of responsibilities.

If the right approach is to centralize error handling in the composable layer, I’m happy to refactor that way. To avoid changing services in this PR, could we instead simplify the composables to something like:

```typescript
} catch (err) {
  error.value = err as AppError;
  showToastError((err as AppError).message);
  return false;
}
```

nicki182:
No, its correct. Because if you look at the service it is normalizing it. But if you want to add it as a safety measure, sure leave it. @aasimsyed

nicki182:
> Thanks for the review, @nicki182. I want to make sure I’ve understood your feedback correctly.
> 
> Forgive me if I’m wrong, but I had understood the flow as:
> 
> 1. **Services** catch the raw error, normalize it with `errorHandler`, and rethrow it. They don’t show UI or set state.
> 2. **Composables** catch that thrown error and handle it for the user (show toast, set `error.value`, return false).
> 
> If that’s right, I added `error.value = appError` because some composables weren’t setting it and the tests expect it. I also called `errorHandler(err)` in the composable for safety, in case something other than an `AppError` is thrown. But I may be misreading the intended split of responsibilities.
> 
> If the right approach is to centralize error handling in the composable layer, I’m happy to refactor that way. To avoid changing services in this PR, could we instead simplify the composables to something like:
> 
> ```ts
> } catch (err) {
>   error.value = err as AppError;
>   showToastError((err as AppError).message);
>   return false;
> }
> ```

I just reread again what you said and the above sounds good. Maybe we can have if instanceof AppError to make it safer that we know how the error will be. @aasimsyed

nicki182:
@aasimsyed I ended up doing your suggestion. Thank you for making this PR, its great. 
Now we have all the services aspect of the application unit tested. :fireworks:

aasimsyed:
praise: @nicki182 thanks for doing that! sorry I couldn't get to it earlier, but have been a bit preoccupied.

nicki182:
> praise: @nicki182 thanks for doing that! sorry I couldn't get to it earlier, but have been a bit preoccupied.

@aasimsyed  all good, I just wanted to bring it in to have the services layer tested. Besides you did the bulk of it, comparatively I did something small :wink:

PR #1923, opened 2026-02-23
fix: fix calendar issue with first load when creating event and movin…

nicki182:
…g loading of global css

### Contributor checklist

---

### Description

The bug happens because vite when first building the page recognizes and loads all libraries first. But since the calendar component which contains a library is not loaded beforehand, because of that when the component loads in vite reloads and build itself again. From what I understand this shouldn't happen in production. 

As a solution when the store loads which is when the page loads,the component is called which makes the library load in to dependencies. 

I moved some css to be global setup since they are static. 
### Related issue

- #1924

### DIFF
--- a/frontend/app/stores/machines/create-event.ts
+++ b/frontend/app/stores/machines/create-event.ts
@@ -5,15 +5,15 @@ const EventDetailsStep = () =>
   import("../../components/machine/steps/createEvent/MachineStepsCreateEventDetails.vue");
 const EventTypeStep = () =>
   import("../../components/machine/steps/createEvent/MachineStepsCreateEventType.vue");
-const TimeStep = () =>
+const EventScheduleStep = () =>
   import("../../components/machine/steps/createEvent/MachineStepsCreateEventTime.vue");
-const LocationStep = () =>
+const EventLocationStep = () =>
   import("../../components/machine/steps/createEvent/MachineStepsCreateEventLocation.vue");
-const OnlineLink = () =>
+const EventOnlineLink = () =>
   import("../../components/machine/steps/createEvent/MachineStepsCreateEventLinkOnline.vue");
 
 if (import.meta.env.DEV) {
-  TimeStep();
+  EventScheduleStep();
 }
 
 const { create } = useEventMutations();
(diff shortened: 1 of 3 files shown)


nicki182:
@kmje405 or @kolin-newby either of you can look at my solution and maybe suggest better ideas, I will make a proper description of the issue later.

frontend/app/stores/machines/create-event.ts
```diff
@@ -12,6 +12,10 @@ const LocationStep = () =>
 const OnlineLink = () =>
   import("../../components/machine/steps/createEvent/MachineStepsCreateEventLinkOnline.vue");
 
+if (import.meta.env.DEV) {
+  TimeStep();
```
kolin-newby: The naming of this confused me a little. Is the idea that this is the "time selection step" of some process, from looking at the value that's what I'm guessing, but it could be more explicit to the component being imported: `MachineStepsCreateEventTime`.

kolin-newby: I think this confusion comes from the common 'TimeStep' name meaning something along the lines of incrementing some time counter or dev variable. IOW: "one step of time" vs. "the time step of a process"

nicki182: @kolin-newby you mean the name of the variable to make it more explicit? I didn't think it was that necessary since this the state machine and I'm just bringing the step components. I thought if I named as you said it would be more confusing since the machine itself is the create-event machine but I can change it to maybe ScheduleStep since it's step to make give the schedule of the event maybe....

andrewtavis: @kolin-newby: We did a rename to `EventScheduleStep` to match some of the other variables that we have for this. The next verbose step for naming the variables would be `CreateEventScheduleStep`, and beyond that I don't think we need to do :) Let us know if you think it would be more clear if we're naming the steps explicitly within these files!

andrewtavis: We'll likely merge here, but this can be a PR in itself as we'll need to do the change for groups and orgs as well :)

PR #1934, opened 2026-02-27
Adds Event ID Filter and Applies Filter After Event Creation

### LINKED ISSUE #1928: Creating an event or multiple events should redirect to the `/events` route and filter for all created events
### Terms

### Description

#1927 looks into where the user should be navigated after organizations and groups are created. This issue would do the same for events, where the idea we had was that the user could be navigated to `/events` with the filter set to only show the events that the user just created. This would allow us to show the same response regardless of if the user created one event or multiple.

I'm open to discussing if this redirect should only happen if the user uses the option to create more events at the end of the event creation flow, and if not then they're navigated to the event's about page. This might be the expected behavior for those creating one event, with the navigation to `/events` being learned behavior for those users who choose to make more than one event at once.

### Contribution

Happy to discuss the work to be done here and then support with implementation! 🚀

kolin-newby:
### Contributor checklist

---

### Description

WIP

I have implemented the new event ID filter and the tests for it on the backend, and wanted to open a PR so someone can review it while I work on the frontend changes.

The event ID filter operates on the logic that any invalid or nonexistent IDs will be ignored, and any valid and existent IDs passed along side the former will still be applied. If you can think of a better logic for this, please feel free to bring it up!

### Related issue

- closes #1928

### DIFF
--- a/frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
+++ b/frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
@@ -151,12 +151,9 @@
 </template>
 
 <script setup lang="ts">
-import type { CreateEventInput } from "~~/shared/types/event";
-
 import { addDays, differenceInCalendarDays, format, isSameDay } from "date-fns";
 import { z } from "zod";
 
-const { create } = useEventMutations();
 const flow = inject<FlowControls>("flow");
 const isCreating = ref(false);
 
@@ -269,47 +266,9 @@ const handleSubmit = async (values: Record<string, unknown>) => {
     all_day: t.allDayLong || false,
   }));
 
-  const nodeData = flow.context.value
-    .nodeData as unknown as ContextCreateEventData;
-
-  const createdEventIds: string[] =
-    nodeData[CreateEventSteps.Time]?.createdEventIds ?? [];
-
-  let nextCreatedEventIds: string[] = [...createdEventIds];
-
-  if (createAnother) {
-    isCreating.value = true;
-    try {
-      const data = Object.values({
-        ...nodeData,
-        [CreateEventSteps.Time]: {
-          times: mappedTimes,
-          createAnother,
-          createdEventIds,
-        },
-      }).reduce(
-        (acc, data) => ({
-          ...(acc as Record<string, unknown>),
-          ...(data as Record<string, unknown>),
-        }),
-        {}
-      );
-      const event = (await create(
-        data as CreateEventInput
-      )) as unknown as EventResponse;
-
-      if (event) {
-        nextCreatedEventIds = [...nextCreatedEventIds, event.id];
-      }
-    } finally {
-      isCreating.value = false;
-    }
-  }
-
   flow.next({
     times: mappedTimes,
     createAnother,
-    createdEventIds: nextCreatedEventIds,
   });
 };
 </script>
--- a/frontend/app/components/modal/create/ModalCreateEvent.vue
+++ b/frontend/app/components/modal/create/ModalCreateEvent.vue
@@ -60,7 +60,7 @@ async function handleCreatedEventRouting(createdEventIds: string[]) {
     return;
   }
 
-  const viewQueryValue = route.query.view;
+  const viewQueryValue = route.query.view || ViewType.LIST; // default to 'list' if no view query param is present
 
   // Preserve the next query in case we are navigating to a new path.
   const preserveNextQuery = useState("preserveNextQuery", () => false);
--- a/frontend/app/composables/queries/useGetOrganizationByUser.ts
+++ b/frontend/app/composables/queries/useGetOrganizationByUser.ts
@@ -16,37 +16,58 @@ export function useGetOrganizationsByUser(
   const page = ref(1);
   const userIdRef = computed(() => userId);
   const filtersRef = computed(() => unref(filters));
-
+  const oldFilters = ref<OrganizationFilters | undefined>(filtersRef.value);
+  const isFinished = ref(true);
   // UseAsyncData for SSR, hydration, and cache.
   const { data, pending, error, refresh } = useAsyncData<Organization[]>(
     () => getKeyForGetOrganizationsByUser(userId, page.value, filtersRef.value),
     async () => {
       try {
-        if (!userIdRef.value || userIdRef.value === "") {
-          return [];
+        if (
+          !userIdRef.value ||
+          (userIdRef.value === "" && !isLastPageRef.value)
+        ) {
+          return organizations.value as Organization[];
         }
+        if (
+          JSON.stringify(oldFilters.value) !== JSON.stringify(filtersRef.value)
+        ) {
+          oldFilters.value = filtersRef.value;
+          page.value = 1;
+          organizations.value = [];
+        }
+        isFinished.value = false;
         const paginatedOrganizations = await listOrganizationsByUserId(
           userIdRef.value,
           page.value,
           filtersRef.value
         );
+        isFinished.value = true;
         isLastPageRef.value = paginatedOrganizations.isLastPage;
-        return [
-          ...organizations.value,
-          ...paginatedOrganizations.data,
-        ] as Organization[];
+        const newOrgs = paginatedOrganizations.data.filter(
+          (newOrg) =>
+            !organizations.value.some(
+              (existingOrg) => existingOrg.id === newOrg.id
+            )
+        );
+        organizations.value = organizations.value.concat(newOrgs);
+        return organizations.value as Organization[];
       } catch (error) {
         showToastError((error as AppError).message);
         throw error;
       }
     },
     {
-      watch: [page, userIdRef, filtersRef],
+      watch: [userIdRef, filtersRef],
       default: () => [],
     }
   );
   const getMore = async () => {
-    if (isLastPageRef.value) {
+    if (
+      isLastPageRef.value ||
+      JSON.stringify(oldFilters.value) !== JSON.stringify(filtersRef.value) ||
+      !isFinished.value
+    ) {
       return;
     }
     page.value += 1;
--- a/frontend/app/pages/events/index.vue
+++ b/frontend/app/pages/events/index.vue
@@ -39,8 +39,8 @@
 </template>
 
 <script setup lang="ts">
-const viewType = ref<ViewType>(ViewType.MAP);
 const route = useRoute();
+const viewType = ref<ViewType>((route.query.view as ViewType) || ViewType.MAP);
 const router = useRouter();
 const loadingFetchMore = ref(false);
 


frontend/app/components/modal/create/ModalCreateEvent.vue
```diff
@@ -22,6 +22,7 @@ const { create } = useEventMutations();
 async function handleSubmission(data: unknown) {
   await create(data as CreateEventInput);
   handleCloseModal();
+  console.log("TESTING"); //eslint-disable-line no-console
```
kolin-newby: This marks the place I am thinking of implementing the frontend changes.

backend/events/filters.py
```diff
@@ -45,6 +46,56 @@ class EventFilters(django_filters.FilterSet):  # type: ignore[misc]
...
+        uuids = []
+        for v in raw_values:
+            try:
+                # UUID value validation
+                uuids.append(uuid.UUID(str(v).strip()))
+            except ValueError:
+                # If a single UUID is invalid it should still filter for the valid ones
+                continue
+
+        return queryset.filter(id__in=uuids) if uuids else queryset.none()
```
kolin-newby, line 97: I went the filter function route because implementing a ModelMultipleChoiceFilter, like what we have for topics, had an issue validating UUID values due to some weird parameter passing behavior.

For my direct testing (on Firefox), the values were collected as a list with one string of all the UUIDs separated by commas: ["uuid_1,uuid_2,uuid_3,uuid_4"]

But for the test cases that I made, the list was as expected: ["uuid_1", "uuid_2", "uuid_3" ,"uuid_4"]

Thus the need for some custom parsing. Let me know what you think.

andrewtavis:
Thanks so much for the PR, @kolin-newby! We'll try to get to the review in the coming days :)

kolin-newby:
@nicki182 @andrewtavis Just pushed the working frontend changes. I'm sure there's some tightening up to do on this yet, but I wanted to get it updated so you could have a look. Let me know what you think and I will try to add some comments to explain my thinking.

nicki182:
@nicki182 I just had a glanced on it and I have my thoughts... I will do my proper review about it later. I will ask questions on the review too because I want to understand what you understood from the code to make it more readable.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
```diff
@@ -257,6 +264,48 @@ const handleSubmit = async (values: Record<string, unknown>) => {
...
+        data as CreateEventInput
+      )) as unknown as EventResponse;
+
+      if (event) {
+        nextCreatedEventIds = [...nextCreatedEventIds, event.id];
+      }
+    } finally {
+      isCreating.value = false;
+    }
+  }
```
kolin-newby: I moved the create logic to this step file so it can be handled asynchronously, allowing us to access the return value of the create function to get the IDs of the created events.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
```diff
@@ -150,10 +150,14 @@
...
 
 <script setup lang="ts">
+import type { CreateEventInput } from "~~/shared/types/event";
+
 import { addDays, differenceInCalendarDays, format, isSameDay } from "date-fns";
 import { z } from "zod";
 
+const { create } = useEventMutations();
 const flow = inject<FlowControls>("flow");
+const isCreating = ref(false);
```
kolin-newby, line 158: A TODO could maybe be to map the submit button to this, so users can see that the system is working if there is any delay.

frontend/app/components/modal/create/ModalCreateEvent.vue
```diff
@@ -20,8 +20,16 @@ const { create } = useEventMutations();
...
+  // collect 'createdEventIds' from the flow context on submission and add final event ID
+  const typedData = data as CreateEventInput;
+  let createdEventIds: string[] = typedData.createdEventIds;
+  const res = (await create(typedData)) as unknown as EventResponse;
+  if (res) createdEventIds = [...createdEventIds, res.id];
+
   handleCloseModal();
+
+  // return list of created event IDs to be handled by watcher in 'useFlowScreens.ts'
+  return createdEventIds;
```
kolin-newby: This is just here to collect the ID of the event created by the flow's `onSubmit` function and pass the list of IDs to the useFlowScreen watch statement.

frontend/app/composables/useFlowScreens.ts
```diff
@@ -67,10 +70,15 @@ export function useFlowScreens(
...
+    async (finished) => {
+      let createdEventIds: string[] = [];
       if (finished && options.onSubmit) {
-        options.onSubmit(store.saveResult);
+        createdEventIds = (await options.onSubmit(
+          store.saveResult
+        )) as unknown as string[];
       }
+
+      await handleCreatedEventRouting(createdEventIds);
```
kolin-newby: I'm hoping this is the right place to add this. It made sense to me to handle the routing in the flow logic, and in this watch specifically, because it fires the final `onSubmit` function call as the flow exits.

frontend/app/stores/machines/create-event.ts
```diff
@@ -80,18 +78,8 @@ export const useCreateEventStore = createFlowStore({
...
-                ...(data as Record<string, unknown>),
-              }),
-              {}
-            );
-            create(data);
-            return CreateEventSteps.EventDetails;
-          }
-          return "end";
+
+          return createAnother ? CreateEventSteps.EventDetails : "end";
```
kolin-newby: See my comment on moving this logic [here](https://github.com/activist-org/activist/pull/1934/changes#r2875577493).

kolin-newby:
@nicki182 Looking forward to the review! I just fixed the test that was failing and commented most of my thinking in the code.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
```diff
@@ -257,6 +264,48 @@ const handleSubmit = async (values: Record<string, unknown>) => {
...
+      }).reduce(
+        (acc, data) => ({
+          ...(acc as Record<string, unknown>),
+          ...(data as Record<string, unknown>),
+        }),
+        {}
+      );
+      const event = (await create(
+        data as CreateEventInput
+      )) as unknown as EventResponse;
```
kolin-newby: Is there a reason we seem to prefer the `as unknown as <SpecificType>` over using the `<SpecificType>` from the beginning?

For example, why not have the `create()` function (and the `createEvent()` call it makes) return an `EventResponse` when it's called instead of type casting after we get the value back?

I know I added this code, but the casting I'm asking about seems to be a common pattern in the code base. In case it's not clear, I'm genuinely curious; I'm not trying to suggest this is wrong.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
nicki182: I was thinking about it but it shouldn't take that much time. My original idea is when you go through the loops you should create in the background because it will take some time for the user to finish the whole process again if they create another event. We can add tooltip each time one is created maybe.  And only when you don't decide to loop the submit button loads and takes you to the event or the list filtered by events. Which will only create one event .

frontend/app/composables/useFlowScreens.ts
nicki182: I would re think this. useFlowScreen should not have code that is specific for a case it should always be generic you can maybe add a callback or add rerouting logic since there is multiple cases that is needed. This is managing the flow of screens for the machines that where created.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
nicki182: Because we can't use any, I first use unknown for generics. What happens is that afterwards when I want to use it with a specific type I first I have to use unknown and the the correct type to make the typescript work. If you have a better suggestion I'm all ears to hear about it. If it can make it more understandable the code I'm always happy to hear it.

frontend/app/components/modal/create/ModalCreateEvent.vue
nicki182: I think its better to reroute here directly and not in the useFlowScreens.

frontend/app/stores/machines/create-event.ts
nicki182: I would rethink this because I prefer if it wasn't in a step the create event because this could easily change and I prefer all services call above the steps layer. The steps should be as simple as possible since it should only contain its own logic. I do understand your thinking behind this, specially if you want to add that the submit button has loading but I would try to abstract and see of solution that is within the machine context layer to trigger possibly some loading effect.

nicki182 (commented):
@kolin-newby I have my thoughts as I said and would like to hear your thought process of what I just said. And see if we can come with a solution we are both satisfied. I want to maintain a infrastructure of the code so that each layer has its own role. Thank you so much for your contribution and opening a discussion possibly of the structure of the code maybe. 

@andrewtavis any thoughts do you have?

frontend/app/components/modal/create/ModalCreateEvent.vue
kolin-newby: I think that would be fine. I ran into some issues getting the routing to take effect from there, but I think I should be able to sort that out if that's the better place for it to be.

frontend/app/composables/useFlowScreens.ts
kolin-newby: Makes sense, thanks for explaining!

frontend/app/stores/machines/create-event.ts
kolin-newby: Okay, I can take another look at this as well. The main issue is that I need to await the `create()` call so that I can use its return value. The machine steps, as far as I understand, should all be synchronous, so I would need to add the list parsing somewhere else if this is the case.

kolin-newby: @nicki182 Is it feasible to make the `.CreateMoreEventsOrNot` step inside `create-event.ts` async to allow handling the create and list parsing there?

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
kolin-newby: My only thought would be a sweep of the system to implement strict custom typing across all these scenarios. Unfortunately, in my experience with TypeScript, it's all or nothing. That would be an undertaking, but not necessarily out of the question in the future!

kolin-newby:
@nicki182 ready for another review!

nicki182:
@kolin-newby if you don't mind I cleaned up some code that was not necessary and I fixed some issues regarding a composable that I made since I was there and everything looks good. So, I'll bring it in

nicki182:
@kolin-newby thank you for your work. This is great :smile:

Issue #1938, opened 2026-03-03
fix(a11y): Map tooltip images missing alt text and CardDonate edit toggle not keyboard accessible


Solus90:
### Terms

- [x] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description

Two accessibility issues were found in the frontend:

1. Map tooltip images missing alt text
   In `MediaMapEvent.vue` and `MediaMapEvents.vue`, icon images inside map popup
   tooltips are built via innerHTML with no `alt` attribute, violating WCAG 1.1.1
   (Non-text Content). Each image sits next to a `<span>` that already provides
   the label, so the correct fix is `alt=""` to mark them as decorative.

2. CardDonate edit toggle is not keyboard accessible
   In `CardDonate.vue`, the edit/cancel toggle is implemented as a bare `<Icon>`
   with an `@click` handler inside a `<div>`. This element has no `role`, no
   `tabindex`, and no `aria-label`, making it completely inaccessible to keyboard
   and screen reader users. It should be a `<button>` with a dynamic aria-label
   reflecting the current state (edit / cancel).

### Contribution

I'd love to help resolve this

philorien:
Feel free to assign this to me and I can get a fix in

nicki182:
@Solus90 all yours! 😉

andrewtavis:
Thanks for the issue here, @Solus90! Quick note that @aasimsyed works a lot on a11y and general usability for the platform :) Maybe you two can discuss what could be explored for potential improvements in these regards 😊

PR #1939, opened 2026-03-03
fix(a11y): add alt text to map tooltip images and make CardDonate edit toggle keyboard accessible

Solus90:
### Contributor checklist

---

### Description

Fixes two accessibility issues found during a frontend a11y audit:

**1. Map tooltip images missing `alt` attributes (`MediaMapEvent.vue`, `MediaMapEvents.vue`)**

Icon images inside map popup tooltips were built via `innerHTML` with no `alt` attribute, violating WCAG 1.1.1 (Non-text Content). Each image sits next to a `<span>` that already provides the label, so `alt=""` is correct — marking them as decorative and avoiding redundant screen reader announcements.

**2. CardDonate edit toggle not keyboard accessible (`CardDonate.vue`)**

The edit/cancel toggle was a bare `<Icon>` with an `@click` handler inside a plain `<div>` — no `role`, no `tabindex`, no `aria-label`. Keyboard and screen reader users had no way to reach or understand it. Replaced with a `<button type="button">` with `focus-brand` styling and a dynamic `aria-label` that updates to reflect the current state (edit vs cancel), using new i18n keys added to `en-US.json`.

### Related issue

- #1938

### DIFF
--- a/frontend/app/components/card/CardDonate.vue
+++ b/frontend/app/components/card/CardDonate.vue
@@ -7,21 +7,19 @@
       </h3>
       <button
         v-if="userIsAdmin"
-        type="button"
-        :aria-label="editModeEnabled ? $t('i18n.components.card_donate.cancel_edit_donation_info_alt_text') : $t('i18n.components.card_donate.edit_donation_info_alt_text')"
         @click="toggleEditMode"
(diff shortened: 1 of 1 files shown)


andrewtavis:
Thanks for the PR, @Solus90! We're checking this out in the dev sync that we have right now. We'll fix the errors and then bring this in!

PR #1943, opened 2026-03-04
Fix: Ensure UTC date handling in date utilities

### LINKED ISSUE #1937: Date Formatting Tests Fail in Frontend For Certain TimeZones at Certain Times
### Terms

### Summary:
The tests in `date.spec.ts` that use the calls in the format `const date = new Date("2024-06-15");` fail during certain times, in certain timezones due to the custom `formatDate(date);` function calling the `*.getDate()`, which uses local time, while the `new Date("<date string>")` call assumes UTC-0 time.

There may also be other tests in other files that are affected by this.

### Reproduce:
This is difficult due to the nature of timezones...

Be in a negative UTC offset timezone during a time that is a number of hours (equal to or less than your offset from UTC-0) away from midnight. (ex: UTC-7 during or after 1700 local time)

If you can set your system time manually you may be able to run the tests locally and reproduce.

### Fix:
Using a format like `const date = new Date(2024, 5, 15);` over a string date should solve the issue as the numeric value input seems to also use local time.

rohan-pandeyy:
closes #1937

### Contributor checklist

---

### Description

There was a mismatch between UTC construction and local-time accessors. So when local timezone is behind UTC (negative offset), UTC midnight rolls back to the previous day in local time.

image

#### Fix: use UTC accessors in `getAllDaysInRange` and `formatDate` in `date.ts`

image

and all the 16 tests pass :)

### DIFF
--- a/frontend/shared/utils/date.ts
+++ b/frontend/shared/utils/date.ts
@@ -7,13 +7,13 @@ export const getAllDaysInRange = (range: {
   const currentDate = new Date(range.start);
   while (currentDate <= range.end) {
     dates.push(new Date(currentDate));
-    currentDate.setDate(currentDate.getDate() + 1);
+    currentDate.setUTCDate(currentDate.getUTCDate() + 1);
   }
   return dates;
 };
 export const formatDate = (date: Date): string => {
-  const day = String(date.getDate()).padStart(2, "0");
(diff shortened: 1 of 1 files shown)


nicki182:
@kolin-newby you can take a look at this? and if its good I can do a final review to check if it works and merge it.

PR #1948, opened 2026-03-05
Update machine to fit new requirements and make it more rebust

nicki182:
…ed data is added to add data bewtween nodes and added laoding for submit

### Contributor checklist

---

#### Update of machine:
- The action node: An invisible node (like logic) designed specifically to pause the machine, delegate a side-effect (like an API call) back to the UI component, and then automatically continue. (useFlowScreens,flow and create-event are the files affected by this change)
- Machine Validation: A new defineFlowMachine within utils folder has been added. This wrapper that automatically validates step configurations during runtime. (create-group, create-event, create-organization,flow, machine-type are the files affected by this change).
- New loading management: When its saving you can access this through the context of the provider when a submit between steps happens you will see this.  
- State Management: sharedData has been added on the store so that data shared by the nodes there. After an action node, it automatically saves the callback's return value to store.sharedData.__lastActionResult. And there os setSharedData action is passed through context. (flow and useFlowScreens).
### Related issue

This is mainly to support and finish
- #1928


nicki182:
@kolin-newby take a look and ask all the questions needed and tell me if you understand what is happening. If this helps with what is needed maybe on saturday we can review it.

andrewtavis:
Quick note here from the sync, @kolin-newby: The thinking is that we can bring this PR in with your approval and then you can bring the changes into #1934 so that the changes that you made for the backend can then be leveraged and we can finalize this :)

Let us know if that sounds good on your end!

kolin-newby:
Sounds good to me! I have been up in the mountains the last few days but I will be back home today and able to start reviewing the changes!

frontend/app/services/event/event.ts
```diff
@@ -72,9 +72,11 @@ export async function listEvents(
 
 // MARK: Create
 
-export async function createEvent(data: CreateEventInput): Promise<string> {
+export async function createEvent(
+  data: CreateEventInput
+): Promise<EventResponse> {
   try {
-    const res = await post<string, typeof data>(`/events/events`, data, {
+    const res = await post<EventResponse, typeof data>(`/events/events`, data, {
```
kolin-newby, line 79: Was thinking about this, looks good!

nicki182:
@kolin-newby I will merge it in now and you can work on your end. :wink:

PR #1971, opened 2026-03-08
fix: update E2E docs and script for reliability

### LINKED ISSUE #1970: Check documenation of e2e tests
### Terms

### Issue

We now have e2e tests running on all commits to main and PRs. It would be great if we would do a check of the e2e section of the contributing guide as well as the [run-e2e-tests.sh](https://github.com/activist-org/activist/blob/main/run-e2e-tests.sh) to make sure that these reflect the current setup for testing. In the e2e testing section and also in the suggested extensions for VS Code users, it would also be good if we mention the Playwright extension so that people know that they can use it 😊

aasimsyed:
### Contributor checklist

---

### Description

Updates the E2E test script and documentation to fix reliability issues when running tests locally (addresses #1970).

**`run-e2e-tests.sh`**
- Replace `yarn build` + `npx serve dist/` with `yarn build:local` + `node .output/server/index.mjs` so the Nitro node-server preset is used consistently
- Add `export USE_PREVIEW=true` so the build outputs to `.output/` (node-server) rather than `dist/` (netlify-static)
- Pass `NUXT_SESSION_PASSWORD` and `NUXT_API_SECRET=""` explicitly when starting the node process so env vars are never silently stripped by nohup/yarn
- Add `rm -rf dist` guard before building to prevent nuxi preview from picking up a stale static build
- Add a curl readiness loop with a dot progress indicator so it's clear the script is waiting and not hung

**`CONTRIBUTING.md`**
- Update the manual E2E steps to match the script exactly

**`FRONTEND_TESTING.md`**
- Add a "Checking E2E coverage" subsection documenting the `e2e-coverage.mjs` script and its usage modes

**`backend/Dockerfile`**
- Install `curl` in the Alpine image so Docker healthchecks can execute

**`docker-compose.yml`**
- Fix backend healthcheck endpoint from `/health/` (404) to `/v1/schema/` (200)
- Add `start_period: 60s` to allow time for Django migrations and `populate_db` on first startup

Tested by running `sh run-e2e-tests.sh` end-to-end locally.

### Related issue

- Closes #1970

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -449,14 +449,14 @@ Please see the [frontend testing guide](FRONTEND_TESTING.md) for information on 
 
 #### Run Local E2E Tests
 
-activist uses [Playwright](https://playwright.dev/) for end to end testing. You'll first need to install/update the browsers installed for Playwright as described in their [updating Playwright documentation](https://playwright.dev/docs/intro#updating-playwright). Please run the following command in the frontend:
+activist uses [Playwright](https://playwright.dev/) for end to end testing. You'll first need to install/update the browsers installed for Playwright as described in their [updating Playwright documentation](https://playwright.dev/docs/intro#updating-playwright). Run the following command in the frontend:
 
 ```bash
 # This and all following steps need to be ran each time Playwright is updated.
 yarn playwright install --with-deps
 ```
 
-Please run the following to run the end to end testing suite:
+Run the following to run the end to end testing suite:
 
 ```bash
 # Note: There may be an installation prompts in the build logs. Hit 'n' to say no.
@@ -472,6 +472,9 @@ yarn playwright show-report
 docker compose --env-file .env.dev down
 lsof -ti tcp:3000 | xargs kill -9 2>/dev/null || true
 ```
+
+> [!NOTE]
+> VS Code users can use the [Playwright extension](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) to run the tests. Go to the testing view and then run the test suite. You can also select options including which devices to run and whether to view the browser.
 
 Alternatively, to run the end to end tests using separate shells, please run the following:
 


FRONTEND_TESTING.md
```diff
@@ -81,6 +81,30 @@ The steps to do this are:
...
+node test-e2e/scripts/e2e-coverage.mjs --uncovered
+
+# Machine-readable JSON output:
+node test-e2e/scripts/e2e-coverage.mjs --json
+```
+
+The script distinguishes between:
+
+- **Covered routes** — routes with at least one `goto`, `waitForURL`, or `toHaveURL` assertion in specs, actions, or page objects
+- **Stub routes** — pages marked `:underDevelopment="true"` with no implemented UI (excluded from the testable coverage calculation)
```
andrewtavis, line 105: Let's be sure to keep this in mind and remove it at some point. Could you make an issue to remove this and the functionality to check for `:underDevelopment="true"` that we can get to in the future? Ideally we wouldn't have anything like this in the project.

Let me know if making an issue to note this doesn't make sense! Just seems like something that will naturally be forgotten.

andrewtavis: I can also make the issue, but let me know if you think it's worthwhile :)

CONTRIBUTING.md
```diff
@@ -473,41 +473,46 @@ docker compose --env-file .env.dev down
 yarn kill-port 3000
```
andrewtavis: Do we also want this to be `lsof -ti tcp:3000 | xargs kill -9 2>/dev/null || true`?

andrewtavis: And I'll add a quick note that contributors can also use the extension :)

aasimsyed: @andrewtavis yes we should change that to `lsof -ti tcp:3000 | xargs kill -9 2>/dev/null || true`

andrewtavis (approved):
praise: Thanks for getting to this, @aasimsyed! I'm sure the efforts here will have some great results once new community members start working on the e2e testing issues 😊

andrewtavis:
Most recent commit was just to the contributing. Will merge as tests past before.

PR #1998, opened 2026-03-12
feat(e2e): Event Create Modal E2E tests, create-flow helpers, and app fixes

### LINKED ISSUE #1995: Phase 1: Create dropdown and modal page objects (E2E create flows)
Subtask of #1994 (Phase 1 of E2E Create Flows epic).

## Scope

Add shared infrastructure for testing the entity create flows (state machine modals): Create dropdown component object, modal page objects for Event/Organization/Group create, and notes on auth and test data (captured in this issue).

### 1. Create dropdown component object

- **Add** `frontend/test-e2e/component-objects/CreateDropdown.ts` (or `SidebarCreateDropdown.ts`).
- Scope the Create button to `#sidebar-left` (e.g. `page.locator('#sidebar-left').getByRole('button', { name: ... })`) using `getEnglishText('i18n.components.dropdown_create.create_aria_label')` or equivalent.
- Expose: `root`, `createButton`, `clickNewEvent()`, `clickNewOrganization()`, `clickNewGroup()` using `getEnglishText()` for menu labels (`i18n._global.new_event`, `i18n.components.dropdown_create.new_organization`, `i18n._global.new_group`).
- Follow the same pattern as `EventsFilter`: separate component object, not methods on SidebarLeft. Usage: `sidebar.open(); const createDropdown = newCreateDropdown(page); await createDropdown.clickNewEvent();`

### 2. Modal page objects

- **CreateEventModal** — `frontend/test-e2e/component-objects/CreateEventModal.ts` (or under a subfolder). Root: `page.getByTestId('modal-ModalCreateEvent')`. Locators/methods for: EventDetails (name, tagline, description, orgs, groups), EventType (location type, event type, topics), Location, LinkOnline, Time; Next/Previous/Submit; modal close. Use form id `#event-details` and `data-testid="events-filter-location-type"`, `data-testid="events-filter-event-type"`.
- **CreateOrganizationModal** — Root: `page.getByTestId('modal-ModalCreateOrganization')`. OrganizationDetails (name, tagline, description), Location (country, city); Next/Previous/Submit; modal close.
- **CreateGroupModal** — Root: `page.getByTestId('modal-ModalCreateGroup')`. GroupDetails (name, tagline, description, organization), Location; Next/Previous/Submit; modal close.

Pattern: same style as `EditModal` / `SocialLinksModal` (locators + helpers), scoped to the specific modal testid.

### 3. Notes

- Create flow E2E expects a **signed-in user** (admin or member); default project storage state is admin.
- **Event create** requires at least one **organization** (and optionally groups) in the combobox; backend/fixtures must provide these.
- **Group create** requires one **organization** in the combobox.

### LINKED ISSUE #1996: Phase 2: Event create modal E2E spec
Subtask of #1994 (Phase 2 of E2E Create Flows epic). Depends on Phase 1 (Create dropdown and modal page objects).

## Scope

Add E2E spec for the Event create modal (ModalCreateEvent): open from Create dropdown, assert modal and steps, validation, full flows (physical and online), Previous button, and modal close.

### 1. Spec file

- **Add** `frontend/test-e2e/specs/all/create-flows/event-create-modal.spec.ts`.
- **beforeEach:** Navigate to a page where sidebar is visible (e.g. `/en` or `/home`); use default global storage state (signed-in); open left sidebar (`newSidebarLeft(page)`, `sidebarLeft.open()`); open Create dropdown and click "New event" (`newCreateDropdown(page)`, `createDropdown.clickNewEvent()`). Use `newCreateEventModal(page)` in tests as needed.

### 2. Test cases

- **Modal opens** — After Create → New event, assert `modal-ModalCreateEvent` visible and heading "Create a New Event" (or i18n via `getEnglishText("i18n.components.modal_create_event.create_new_event")`).
- **Step 1 — Event details validation** — Leave name or description empty; click Next step; assert validation messages and modal stays open.
- **Step 1 → 2 — Details filled** — Fill name, description, select one org from combobox; click Next step; assert step 2 (Event type) visible.
- **Full flow (physical event)** — Event details → Event type (physical, type + topics) → Location → Time → Submit. Assert modal closes; optionally assert no error or success toast.
- **Full flow (online event)** — Same but Event type "online"; complete LinkOnline step then Time → Submit. Assert modal closes.
- **Previous button** — From step 2, click Previous; assert step 1 visible with data retained.
- **Modal close** — Open modal, click `data-testid="modal-close-button"`; assert modal hidden.

### 3. Notes

- Event details step requires at least one **organization** in the combobox; rely on backend/seed or fixtures to provide orgs.
- Use fixed locale (e.g. en) and `getEnglishText()` for heading and button assertions.

aasimsyed:
### Contributor checklist

---

## Summary
Adds E2E coverage for the Event Create modal (state-machine flow) and the app changes needed for the tests to pass. Also adds shared create-flow helpers and mobile support so the same tests run on desktop and mobile.

## E2E
- **event-create-modal.spec.ts**: Modal open, step validation, step 1→2, previous (data retained), close, full flow (online), create-another flow, full flow (physical). Tagged `@desktop` and `@mobile`.
- **Mobile**: In `beforeEach`, viewport check; on mobile open hamburger (SidebarRight), use Create from drawer (`#drawer-navigation`); on desktop use sidebar left and Create as before.
- **CreateDropdown**: Optional `root` (e.g. `#drawer-navigation`) so the same helper works for sidebar and hamburger drawer.
- **Component objects**: CreateEventModal, CreateOrganizationModal, CreateGroupModal (locators and step navigation).

## App (required for tests / correctness)
- **ModalBase**: `@click.stop` on panel so combobox clicks don’t close modal; `max-h-[90vh]` so modal scrolls and location-step Next is clickable.
- **useGetOrganizationsByUser**: `MaybeRef` userId, watch to reset page, expose `canFetchMore` so org list loads and pagination is valid.
- **FormSelectorComboboxOrganizations**: Reactive `linked_user_id` and `canFetchMore` so combobox populates and doesn’t request invalid page.
- **organizations/views**: Return 200 + empty results for empty queryset and `NotFound` from pagination (avoid “Invalid page number”).
- **ModalCreateEvent**: `normalizeEventPayload` (orgs/groups→IDs, location lat/lon→strings, times→ISO, strip `createAnother`) before `create()`.
- **MachineStepsCreateEventTime**: Default `endTime = startTime + 1h` so backend `start_time < end_time` validation passes.
- **MachineStepsCreateEventDetails**: `initialDetailsData` + `:initial-values` so form rehydrates when user clicks Previous.
- **create-event store**: Clear Event Details step when looping back for “Create another” so form is fresh for next event.
- **events/serializers**: Remove debug print in `EventPOSTSerializer.validate`.

### Related issues

- closes #1995 
- closes #1996

### DIFF
--- a/frontend/test-e2e/specs/all/create-flows/event-create-modal.spec.ts
+++ b/frontend/test-e2e/specs/all/create-flows/event-create-modal.spec.ts
@@ -216,7 +216,8 @@ test.describe(
       );
     });
 
-    test("Time step retains date and time after Previous then Next", async ({
+    // Time step no longer retains data on Prev→Next (product decision); un-skip if restored.
+    test.skip("Time step retains date and time after Previous then Next", async ({
       page,
     }) => {
       const modal = newCreateEventModal(page);


frontend/app/components/modal/create/ModalCreateEvent.vue
```diff
@@ -15,36 +15,71 @@ const modalName = "ModalCreateEvent";
...
+}
+
 // Runs mid-machine during the "loop" node.
 async function handleLoopSubmit(iterationData: unknown) {
   const dataToSubmit = Object.values(
     iterationData as ContextCreateEventData
   ).reduce((acc, d) => ({ ...acc, ...(d as Record<string, unknown>) }), {});
-  return create(dataToSubmit as unknown as CreateEventInput); // Returns to useFlowScreens
+  const normalized = normalizeEventPayload(
+    dataToSubmit as Record<string, unknown>
```
nicki182: why is it necessary? since its sending it correctly and everything works and has been checked. I would much rather this be normalized on the step rather than here.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
```diff
@@ -225,11 +231,14 @@ const syncTimesArray = (
     if (existing) {
       newTimes.push(existing);
     } else {
-      // Create new entry with null times.
+      // Default: start at midnight, end 1h later so backend receives start_time < end_time.
```
nicki182: there is a validation already and secondly it shouldn't matter if end is later than end since when its all day long, the validation is not activated

frontend/app/composables/queries/useGetOrganizationByUser.ts
```diff
@@ -7,19 +7,28 @@ export const getKeyForGetOrganizationsByUser = (
...
 ) {
   const { showToastError } = useToaster();
   const isLastPageRef = ref(false);
   const organizations = ref<Organization[]>([]);
   const page = ref(1);
-  const userIdRef = computed(() => userId);
+  const userIdRef = computed(() => unref(userId));
   const filtersRef = computed(() => unref(filters));
 
+  watch([userIdRef, filtersRef], () => {
```
nicki182: I made some changes recently for fixes regarding this. Bring it and check on it. I'm not the happienst about it honestly

nicki182: the canFetchMore logic might not be neccesary

frontend/app/stores/machines/create-event.ts
```diff
@@ -92,6 +92,14 @@ export const useCreateEventStore = createFlowStore({
               __lastActionResult: null, // clean up for the next loop iteration
             });
           }
+          // When looping back to Event Details for "Create another", clear that step so the form is fresh.
+          const data = context.allNodeData as unknown as ContextCreateEventData;
```
nicki182, line 96: If we determine with @andrewtavis that when you go previous you should get the data then yes but then I don't think this would work you need to do so through an action to be added like I did with sharedData

frontend/app/components/modal/ModalBase.vue
```diff
@@ -16,9 +16,10 @@
     >
       <DialogPanel
         id="modal"
+        @click.stop
```
nicki182: someone is doing the change to fix this. I will bring it in. You can look at it.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventDetails.vue
```diff
@@ -106,6 +108,14 @@ const eventDetailsSchema = z.object({
   orgs: z.array(z.string()).min(1, "Please select at least one organization"),
   groups: z.array(z.string()).optional(),
 });
+
+const initialDetailsData = computed(() => {
```
nicki182, line 111: this should be done in each step if we do it in one. @andrewtavis

frontend/app/components/modal/create/ModalCreateEvent.vue
aasimsyed: The form steps hand off values that match the UI/Form state (e.g. orgs/groups as full objects from the combobox, `Date` instances for times, `location` with numeric lat/lon). The backend expects IDs, ISO strings, and string coordinates. I added normalization in the modal so we have a single place that maps that merged state into the API contract before `create()`, and so the E2E flow could pass without touching every step at once.

I’m fine moving this to the steps instead. To do that consistently, I’d: (1) in **Event Details**, map orgs/groups to IDs in `handleSubmit` before `flow.next()`; (2) in **Time**, serialize `start_time`/`end_time` to ISO in `handleSubmit`; (3) in **Location**, ensure `lat`/`lon`/`bbox` are strings before `flow.next()`. The modal would then only strip `createAnother` and call `create(merged)`. If you prefer that approach, I can refactor to that and drop `normalizeEventPayload` from the modal.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue
aasimsyed: The backend’s `EventPOSTSerializer` enforces `start_time < end_time` for each time (and doesn’t skip that when we send non–all-day times). With the previous default (`startTime` and `endTime` both set to `currentDate`), we were sending equal times and got a 400. I added the 1h offset only for **new** date entries so that the payload always satisfies that rule when the user doesn’t change the times.

For all-day events the backend overwrites start/end and the check isn’t used for that path. So this change is only to satisfy the non–all-day case. If you’d rather keep the original default and have the E2E (or product) use the all-day path, or handle this in validation/UI differently, I can revert the default and adjust that way.

aasimsyed: I’ve reverted the 1h default in the step and left the existing validation and all-day behaviour as-is. The E2E tests now handle it on the test side:

- **Non–all-day flows:** After selecting the date range, the test sets the first day’s end time to a later time (e.g. 01:00) by focusing the v-calendar end-time control and typing via the keyboard, so the payload satisfies `start_time < end_time` without changing the step’s default.
- **All-day flow:** I added a dedicated test (“full flow (online event, all day)”) that checks the “All day long” checkbox for the first day, so we never send equal times and the backend’s all-day path applies. No start/end manipulation in the test.

So the step stays as you suggested; only the tests were updated.

frontend/app/components/modal/create/ModalCreateEvent.vue
nicki182: @aasimsyed the thing is that if its not breaking I would much rather not change it. I want to keep the code as simple as possible. 

- The backend correctly handles the payload whether the coordinates are sent as numbers or strings, so manually casting them to strings isn't needed.
- the ids map currently maps is already being received correctly by the backend.
-  maybe the time one the conversion to iso string could be valid.

frontend/app/composables/queries/useGetOrganizationByUser.ts
aasimsyed: Merged main and resolved the conflict by keeping **your implementation**: `oldFilters`, `isFinished`, and the existing fetcher/getMore logic.

frontend/app/components/modal/create/ModalCreateEvent.vue
aasimsyed: Makes sense. Reverted to main with no normalization; modal just merges step data and calls `create()`.

frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventDetails.vue
aasimsyed: Done in each step now.

frontend/app/stores/machines/create-event.ts
aasimsyed: Done. Now uses the action + sharedData approach: onExit only sets clearEventDetails: true; Event Details’ onEnter calls clearNodeData(EventDetails) when that flag is set. No direct mutation of node data in onExit.

nicki182:
Is this good for reviewing @aasimsyed?

aasimsyed:
> Is this good for reviewing @aasimsyed?

Yes. Should be I believe. Thanks!

nicki182:
@aasimsyed I made some changes and cleaned some code. You can look at it and tell me about it. I will bring it in if everything finishes

nicki182:
@aasimsyed thanks for fixing the issues and let's bring this in

aasimsyed:
something broke the e2e tests also, looking into that as well

aasimsyed:
@nicki182 The E2E test "Time step retains date and time after Previous then Next" is failing in CI because the Time step no longer retains its data when the user clicks Previous and then Next.

Root cause (from the current code):

1. **frontend/app/components/machine/steps/createEvent/MachineStepsCreateEventTime.vue**
   - **Lines 244–247** (`handlePrev`): Only calls `flow.prev()` with no payload, so the current form values are never saved to nodeData when leaving the step.
   - **Form (lines 6–24)**: No `:initial-values` bound to flow context, so when we navigate back to the Time step the form doesn’t rehydrate from stored data.

2. **frontend/app/stores/machines/flow.ts**
   - **Lines 208–214** (`prev()`): No payload parameter and no merge into `nodeData`, so step data isn’t persisted when going back.

This behavior was removed in a69efd3. Can we either restore Time-step persist/rehydration (e.g. `prev(payload?)` + merge in flow, and in the Time step: read form values in `handlePrev`, pass serialized payload to `flow.prev`, and add `initialTimeData` + `:initial-values` on the Form) or adjust the test/expectation? I’d prefer not to skip the test unless the expectation is wrong.

nicki182:
@aasimsyed I removed it since I didn't see a need for it, why have initial values for the last step?

aasimsyed:
@nicki182 makes sense from a “last step” perspective. The test (and the behavior it was written for) is about the case where the user is on the Time step, has already filled dates/times, then clicks Previous to go back to Link Online (e.g. to check or change the link), then Next again to return to Time. Right now that round-trip clears the time fields, so the test fails when it expects the previously entered "10" / "11" to still be there.

So it’s less “initial values when you first land on the step” and more “retain what the user entered when they navigate away and back.” If we’re okay with the Time step clearing on that round-trip, I can change the test to match (e.g. only assert that we’re back on the Time step / form is visible, and drop the “10”/“11” assertions). If we’d rather keep that UX, we can restore the persist/rehydration for the Time step. Which do you prefer?

nicki182:
@aasimsyed ok, I see what you mean. I prefer to make the sanitazion on the initial values and when you receive, rather as before with changing of handle previous. I want to leave as clean as possible and less manipulation as possible of everything that in nodeData because it should always be the data that is consistently be in a certain way on the forms.

aasimsyed:
Got it, so we keep nodeData updated only when the user leaves a step via Next (sanitize on submit), and when entering a step we sanitize from nodeData into the form’s initial values. No handlePrev writing into nodeData.

With that, the Time step won’t retain the entered times when the user goes Previous and then Next again (we never wrote those values to nodeData). I’ll update the E2E test to match: assert we’re back on the Time step and the form is visible, and remove the assertions on the specific "10"/"11" values so the test passes with the current behavior.

aasimsyed:
Actually, I’ve skipped the E2E test “Time step retains date and time after Previous then Next.” With the current product behavior (no persist on Prev, sanitization only on initial values and on receive), the Time step no longer retains date/time when the user goes Previous and then Next, so the test’s expectations no longer match the app. Rather than rewriting it to only assert that we’re back on the Time step with the form visible (which adds little over existing flow tests), I’ve left the test in place but skipped it, with a comment that it can be un-skipped if we ever restore that retention behavior.

nicki182:
Then I'll bring this in 😉 thank you for making this PR and all the changes

PR #2023, opened 2026-03-18
DRF/Spectacular warnings dumps.

mattburnett-repo:
### Contributor checklist

---

### Description

Stops the repeated drf-spectacular warnings/errors in the backend console when the API schema is generated (e.g. on /v1/schema/ or Swagger UI).

Changes:

- APIView schema: Added @extend_schema with request/responses (and where relevant parameters) for EventCalenderAPIView and SignOutView so spectacular no longer reports “unable to guess serializer”.

- Path parameters: Documented group_id and org_id as UUID path params on GroupImageViewSet and OrganizationImageViewSet via @extend_schema(parameters=[...]).

- Filter “topics”: Kept the existing topics query param behavior; added type-hinted filter_topics methods in OrganizationFilter and EventFilters (and explicit topics params on the list views) so spectacular can infer types and stop warning about “Unable to guess choice types... find 'topics' in model”.

- Enum naming: Set ENUM_NAME_OVERRIDES for CountryCodeEnum in SPECTACULAR_SETTINGS so the shared ISO_CHOICES usage no longer triggers “multiple names for the same choice set”.
Docs: Added full NumPy-style docstrings to the new filter_topics methods for pre-commit numpydoc-validation.

Result: Backend runs without the previous schema-related log spam; OpenAPI/Swagger behavior is unchanged.

### Related issue

- #ISSUE_NUMBER
No known issues currently open for this.

### DIFF
--- a/.github/workflows/ci_e2e_playwright_desktop.yaml
+++ b/.github/workflows/ci_e2e_playwright_desktop.yaml
@@ -9,8 +9,8 @@ on:
       - reopened
       - synchronize
     paths:
-      - "backend/**"
-      - "!backend/**/tests/**/*.py"
+      # - "backend/**"
+      # - "!backend/**/tests/**/*.py"
       - "frontend/**"
       - "!frontend/test/**"
       - "!frontend/i18n/**"
@@ -20,8 +20,8 @@ on:
     branches:
       - main
     paths:
-      - "backend/**"
-      - "!backend/**/tests/**/*.py"
+      # - "backend/**"
+      # - "!backend/**/tests/**/*.py"
       - "frontend/**"
       - "!frontend/test/**"
       - "!frontend/i18n/**"


andrewtavis:
Loading up the dev environment is such a nicer experience with this change :) Thanks again, @mattburnett-repo!

andrewtavis:
@aasimsyed, the e2e tests here have failed twice in a row, so it looks like we have a persistent issue here. Do you have some ideas on what could be causing it?

andrewtavis:
note: The issues here are not related to this change. We also have the same issues with the latest dependabot PR as seen [here](https://github.com/activist-org/activist/actions/runs/23271638590/job/67665512707?pr=2028).

aasimsyed:
@andrewtavis I opened a PR to fix it a little while ago

mattburnett-repo:
> @aasimsyed, the e2e tests here have failed twice in a row, so it looks like we have a persistent issue here. Do you have some ideas on what could be causing it?

@andrewtavis I've noticed this also, and I have some questions, mostly to do with why long-running frontend tests run (and often fail) when the submitted code is backend-only. I looked at the workflow action for the failing check (ci_playwright_e2e_desktop.yaml) and it is indeed set to run whenever a PR with backend changes is submitted. 

Not sure why we run frontend tests for backend code. Without any context, I can't suggest a solution, but it looks like all we would need to do it take out the 'backend' keys for the 'pull_request' and 'main' triggers in the actions file.

mattburnett-repo:
> @andrewtavis I opened a PR to fix it a little while ago

Hi @aasimsyed ! What is the PR number?

aasimsyed:
> > @andrewtavis I opened a PR to fix it a little while ago
> 
> Hi @aasimsyed ! What is the PR number?



> > @andrewtavis I opened a PR to fix it a little while ago
> 
> Hi @aasimsyed ! What is the PR number?

Hi, @mattburnett-repo . This is the PR and it is already merged: https://github.com/activist-org/activist/pull/2027

mattburnett-repo:
> > > @andrewtavis I opened a PR to fix it a little while ago
> > 
> > 
> > Hi @aasimsyed ! What is the PR number?
> 
> Hi, @mattburnett-repo . This is the PR and it is already merged: #2027

Ok good; thanks!

I still don't understand why frontend tests are running when the commit is only backend code. Could you suggest a solution?

mattburnett-repo:
@andrewtavis it looks like we can merge this, in order to clear up the log spam issue this PR addresses.

I've messaged you on element about getting the tests runs sorted out.

andrewtavis:
note, @mattburnett-repo: We made the decision to run only the desktop e2e tests when changes to the backend are made so that we catch any regressions that come from changes there that will affect the frontend as well.

question, @aasimsyed: I saw #2027, but then that PR has already been rebased into this PR? Maybe we can take a look at this further and discussion in the sync :)

I'll bring this in, revert the commenting out of the backend paths for the e2e tests and add a note that let's people know why they're being ran 😊

aasimsyed:
Hi @mattburnett-repo!  E2E runs on backend-only changes to catch regressions that affect the frontend. The UI depends on backend APIs, auth, and data shape; a backend-only change can break the app without touching frontend code.

Examples: changing serializers/views can break response shape or permissions; auth/session changes can break login or role checks; list/ordering changes can break flows the tests rely on. If we only ran E2E when frontend paths changed, backend-only PRs could merge with no check that the full app still works. 

e.g. if `created_by` were removed from the event serializer, the edit icon would never show for the owner (frontend still uses `entity?.createdBy`). An E2E test like “member can edit their own event” would fail even though no frontend code changed. We would only catch that if were manual testing or in production. We run the e2e suite on backend changes as a balance between coverage and CI time.

This is the understanding I have had for the reason to run e2e in this case. Happy to discuss further.

PR #2027, opened 2026-03-18
fix(e2e): deterministic event permission tests

aasimsyed:
### Contributor checklist

---

## Description

In the app, a member is allowed to manage an event when they created it (entity?.createdBy === user.value?.id in useUser.ts). So activist_0 can edit their own events.

When the first event in the list is activist_0’s, the page correctly shows the add/edit buttons. The test expects them hidden, so it fails. When the first event is someone else’s, the buttons are hidden and the test passes.

So the test fails specifically when the first event is one the member created, because in that case the member is allowed to edit it and the UI is correct.

This fixes it by changing only the events list API: when the user `activist_0` exists (e.g. after `populate_db`), order the list so that user’s events come **last**. So the first event is never theirs. No new seed event, no frontend changes just a small conditional in `EventAPIView.get_queryset()` that annotates with `_e2e_last` and orders by it when activist_0 is present. If that user doesn’t exist (e.g. production), behaviour is unchanged.

Member permission tests are now deterministic. Admin tests (and any test that uses the first event) still open the first event; that’s now one of activist_1 … activist_9. The existing seed already gives every event 3 FAQs and 2 resources (`--faq-entries-per-entity 3` and `--resources-per-entity 2` in the populate_db call), so there’s no empty resources/FAQ issue — no extra seed data needed for that.

The activist_0 ordering only runs when `ENVIRONMENT=development` or `CI=true`, so production is unchanged even if someone uses that username.

Tested on my own fork a few times that tests passing now.

### DIFF
--- a/backend/events/views.py
+++ b/backend/events/views.py
@@ -66,9 +66,17 @@ class EventAPIView(GenericAPIView[Event]):
     def get_queryset(self) -> QuerySet[Event]:
         queryset = super().get_queryset().order_by("id")
 
-        # E2E: put activist_0's events last so "first event" is never theirs and
+        # E2E: only in development or CI — put activist_0's events last so
         # member permission tests (open first event, assert no add/edit) are deterministic.
-        e2e_member = UserModel.objects.filter(username="activist_0").first()
+        apply_e2e_ordering = (
+            os.environ.get("ENVIRONMENT") == "development"
+            or os.environ.get("CI") == "true"
+        )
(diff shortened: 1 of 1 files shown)


aasimsyed:
I'll have a follow up PR to add E2E tests for a non-admin able to edit/manage their own event.

nicki182:
@aasimsyed thank you for your PR to fix this issue

PR #2021, opened 2026-03-17
WIP SSR for frontend

nicki182:
### Contributor checklist

- [] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

Description of the proposed change...

### Related issue

- #1602

PR #2036, opened 2026-03-20
Added Clear Filters Button for Events

### LINKED ISSUE #1738: Route queries should be reset when the user navigates to a new route

kolin-newby:
### Contributor checklist

---

### Description

Added a button to the filter section of the events page and implemented function to clear filter queries from the URL.

### Related issue

- closes #1738

### DIFF
--- a/frontend/app/components/sidebar/left/filter/SidebarLeftFilterEvents.vue
+++ b/frontend/app/components/sidebar/left/filter/SidebarLeftFilterEvents.vue
@@ -251,7 +251,7 @@ watch(
     const q = { ...(r.query as Record<string, unknown>) };
     const { view } = q;
 
-    // 🔥 Explicitly set all schema fields to undefined by default.
+    // Explicitly set all schema fields to undefined by default.
     // If they are missing from the URL, this forces VeeValidate to clear them
     // instead of resetting them to their initial page-load values.
     const newFormData = {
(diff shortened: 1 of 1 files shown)


kolin-newby:
@nicki182 I'm guessing I still need to implement tests and possibly localization, but The functionality is ready for review and I will look into the tests.

nicki182:
@kolin-newby from looking at the code it looks good. I'll check on it later and test it but again as I said from looking at the code it should work. Thank you for another great contribution 😉

nicki182:
@kolin-newby wanna check the PR I made about SSR? #2021 its something I don't give much as priority right now and it might be something interesting for you to do on the side. How do you feel about it?

nicki182:
I fixed some issue with toggable and it renders less than before. @kolin-newby 
note:Please its not recommended to make keys that can change in time, they should be constant, for next time. 
suggestion: I think clear all would be better than clear filters or maybe a small button on the left that has the icon of clearing filters but that's something more so I could discuss with @andrewtavis 
praise: works perfectly, we can see maybe we bring it in and finalize it since we want to bring down the amounts of PRs we have

nicki182:
@andrewtavis whenever you can, check the design of the clear button and if its ok. I will add some unit tests and e2e and we can bring it in with this

Issue #2047, opened 2026-03-21
fix: broken or expired link


iampramesh69:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

After clicking at request access their is broken link fix it........

<img width="1919" height="969" alt="Image" src="https://github.com/user-attachments/assets/6cb34f23-76e6-4e7e-9fee-cd796763c0f8" />

Divyansh089:
I am trying to work on this issue

Divyansh089:
This issue was due to an external form link being unavailable. Added a fallback to prevent broken navigation and ensure a smoother user experience.

Sophivorus:
I'm getting the same issue.

rominabrz73:
Hi, I’m seeing this issue as well. I’d like to investigate and work on a proper fix if that’s okay.

Divyansh089:
yes you can proceed .

andrewtavis:
Closing this issue as we're aware of it and have plans to fix the link. Thanks for the issue and discussion :)

Issue #2019, opened 2026-03-17
Unit tests useDebounce


nicki182:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Issue

Write comprehensive unit tests for the useDebounce composable logic and edge cases. Ensure error handling and all main file operation scenarios are tested.
The file is at frontend/composables/generic/useDebounce.ts

arya23923:
May I be assigned this issue?

nicki182:
All your's! @arya23923 if anything don't worry and you can ask about it

arya23923:
hey @nicki182 I have sent a pr for this issue. Can you kindly check review it and let me know if there's any change required :)

nicki182:
@arya23923 I will check on it

PR #2049, opened 2026-03-21
fix bug-broken link

Divyansh089:
### Contributor checklist

---

## Issue
The "Request Access" CTA could lead to a broken experience when the external form endpoint was unavailable.

## Fix
- Added a fallback to the docs "Request Access" page
- Implemented a reachability check before redirecting

## Result
Users are redirected to a stable fallback when the form link is unreachable, avoiding broken navigation.

### Related issue

- #2047

### DIFF
--- a/frontend/app/components/landing/LandingSplash.vue
+++ b/frontend/app/components/landing/LandingSplash.vue
@@ -53,6 +53,7 @@
         fontSize="xl"
         label="i18n.components.landing_splash.request_access"
         :linkTo="REQUEST_ACCESS_URL"
+        @click="handleRequestAccessClick"
       />
     </div>
   </div>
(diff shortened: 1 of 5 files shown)


Divyansh089:
Handled this as an external dependency issue rather than a routing bug. Since the CTA relies on an external form URL that can be unavailable, I added a simple fallback to ensure users are not stuck on a broken link. This keeps the primary flow intact while improving reliability when the form endpoint is unreachable.

andrewtavis:
Thanks for the PR, @Divyansh089 :) We're planning on doing a different solution for this, specifically fixing the link in a way that will be more long lasting. Maybe you'd have interest in another issue like #2019?

PR #2053, opened 2026-03-23
fix: pass label prop to BtnAction in dropdown

### LINKED ISSUE #2052: The dropdown does not show option of file types for downloading barcode.
### Terms

### Behavior

Image

Image

The dropdown does not show option of file types for downloading barcode.
The modal is called ModalQRCode.

saschabuehrle:
Fixes #2052

The dropdown options in BtnActionDropdown were invisible because the option text was passed as slot content (`{{ option }}`) but BtnAction doesn't render a slot — it renders BtnIconsLabel which needs the `label` prop. Switched to passing `:label="option"` directly.

— saschabuehrle

### DIFF
--- a/frontend/app/components/btn/action/BtnActionDropdown.vue
+++ b/frontend/app/components/btn/action/BtnActionDropdown.vue
@@ -47,9 +47,8 @@
             ]"
             :cta="true"
             fontSize="base"
-          >
-            {{ option }}
-          </BtnAction>
+            :label="option"
+          />
         </MenuItem>
       </MenuItems>
     </Menu>


nicki182:
@saschabuehrle I just checked and issue is not fixed. Can you check again and make sure it works before sending a PR?
thank you for taking this on, but please write on the issue you are working on this and please check again manually if it works.

saschabuehrle:
You are right, this one is not sufficient as-is. I will re-test the flow end-to-end and push a corrected update.

Greetings, saschabuehrle

andrewtavis:
Closing this as the current solution is agreed to not be necessary :) Thanks for your interest in the project, @saschabuehrle :)

PR #2094, opened 2026-04-13
test: add backend pytest checks and e2e smoke flow

kowshii27-rbg:
Add starter backend model/API pytest coverage and a frontend Playwright smoke spec for key user flows. Also document the local Docker Postgres pytest setup and use dedicated test settings so contributors can run tests reliably.

Thank you.....


andrewtavis:
@kowshii27-rbg, would you be able to fix the failing tests and provide a bit more context on why these changes are needed?

PR #2109, opened 2026-04-24
fix(frontend): refresh image cache after delete in org/group upload modals

### LINKED ISSUE #1791: Bug: User still needs to refresh the page for the changes in the image carousel to take effect after closing the image upload modal.
### Terms

### Behavior

When a user remove their uploaded image in the upload image modal, the uploaded image still exist in the image carousel after the modal is closed.

Steps to replicate:
1. Upload image to the image carousel.
2. Close the modal, the image you uploaded should be updated on the image carousel.
3. Open the modal and remove the uploaded image.
4. Close the modal.
5. The image you uploaded will still remain on the image carousel and will only be removed after refreshing the page.

### LINKED ISSUE #2108: Bug: Image carousel update after image deletion in image upload modal
### Terms

### Behavior

Similar to #1791, the user needs to refresh the page for the image carousel to update after an image deletion in the image upload modal. A possible solution could be PR #1822.

Refer to #1791 for replication steps, but remove images instead of uploading them.

aasimsyed:
### Contributor checklist

---

### Description

Deleting an image from the upload modal left the carousel stale until a full page reload.
**Cause:** `useFileManager.removeFile()` deletes on the server but can't invalidate the `useAsyncData` cache behind `useGetOrganizationImages` / `useGetGroupImages`. The drop zone already emits `file-deleted` for this case — no parent was listening.
**Fix:** wire `@file-deleted` in both upload modals to the existing `refreshOrganizationImagesData` / `refreshGroupData` helpers (same helpers `uploadImages`/`updateImage` already use). Only refreshes for server-side images (`type === "file"`).
**Files:**
- `ModalUploadImageOrganization.vue`, `ModalUploadImageGroup.vue` — wire the event.
- `MediaImageCarousel.vue` — add `data-testid` (`image-carousel-main` / `-fullscreen`) for reliable E2E selectors.
**Tested:** manual upload → delete round-trip on Desktop and Mobile Chrome, both org and group carousels update without reload. Prettier clean.

### Related issue

- Closes #2108
- Closes #1791

### DIFF
--- a/frontend/app/components/media/image-carousel/MediaImageCarousel.vue
+++ b/frontend/app/components/media/image-carousel/MediaImageCarousel.vue
@@ -3,10 +3,11 @@
   <div class="relative">
     <swiper-container
       ref="swiperRef"
+      class="swiper card-style h-full w-full cursor-pointer overflow-clip"
+      :data-slide-count="imageUrls?.length ?? 0"
       :data-testid="
         props.fullscreen ? 'image-carousel-fullscreen' : 'image-carousel-main'
       "
-      class="swiper card-style h-full w-full cursor-pointer overflow-clip"
       :keyboard="true"
       :loop="true"
       :pagination="{ clickable: true }"
--- a/frontend/app/components/media/image-carousel/MediaImageCarousel.vue
+++ b/frontend/app/components/media/image-carousel/MediaImageCarousel.vue
@@ -3,6 +3,9 @@
   <div class="relative">
     <swiper-container
       ref="swiperRef"
+      :data-testid="
+        props.fullscreen ? 'image-carousel-fullscreen' : 'image-carousel-main'
+      "
       class="swiper card-style h-full w-full cursor-pointer overflow-clip"
       :keyboard="true"
       :loop="true"
--- a/frontend/app/components/modal/upload-image/ModalUploadImageGroup.vue
+++ b/frontend/app/components/modal/upload-image/ModalUploadImageGroup.vue
@@ -12,8 +12,8 @@
       </DialogTitle>
       <div class="mt-4">
         <ImageMultipleFileDropZone
-          @update:modelValue="(newFiles) => (files = newFiles)"
           @file-deleted="handleFileDeleted"
+          @update:modelValue="(newFiles) => (files = newFiles)"
           :model-value="files"
           :uploadLimit="uploadLimit"
         />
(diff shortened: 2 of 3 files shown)


nicki182:
@aasimsyed thank you for bringing this up. Makes total  sense in a delete wouldn't be happening once I saw the code. Can you add the refresh logic in  useFileManager? This file ideally would be refactored by someone in the future, I'm not happy how it is today honestly.

aasimsyed:
@nicki182  thought: tbh I tried putting it there first but got stuck. `useFileManager` doesn't really know whether an image belongs to an org, a group, or whatever other entity we might add later. The caches that'd need invalidating (things like `organizationImages:<id>`, `groupImages:<id>`, etc.) live in the entity-specific composables. So doing the refresh inside `useFileManager` seemed to mean either passing the entity type in and coupling it to every future entity type, or calling every entity's refresh defensively. Neither option felt right, but I'm not super familiar with this file so I might be missing something cleaner.

On the bigger refactor you mentioned, I've been wondering if TanStack Query (or `@pinia/colada` for the Pinia version) might actually fit better than reshaping `useFileManager` by hand. `useAsyncData` plus `refreshNuxtData` kind of feels like a manual subset of what those libraries do natively, and the bug we're fixing here is pretty much the exact shape those libs exist to prevent. Bigger lift for sure, so I wouldn't want to push it without buy-in, just figured it was worth flagging in case it fits where you were already thinking of going.

If it works for you, could we merge this PR as the narrow fix for #2108 / #1791 and I open a separate issue to dig into the TanStack / pinia-colada direction properly? What do you think?

nicki182:
@aasimsyed I just checked the package a bit and looks perfect for my opinion since it even works with what we even have but make it even better and easier. Maybe for later we can add it and make another refactor and perfect the services layer. And yes you have a point with what you just said. I will bring it in as is and yes I do consider we should make a refactor in a sense to still have useFileManager but have in a different compostable everything regarding crud services with images that we have.

nicki182:
@aasimsyed Wanna make an issue about it or I make an issue about it? And then from there how would the refactor would be implemented with a bit more reading of the library to see the steps of doing so. Then I would suggest maybe a divide and conquer between the 2 of us to finish it since I believe it will be a lot of work if one of us does it alone and I want people who work on it to be able to implement such changes properly.

aasimsyed:
@nicki182 awesome, glad it clicks. I can take the issue creation.

nicki182:
@aasimsyed assign it to me and yourself when you do so

Issue #2114, opened 2026-04-25
Add search-dirs to i18n-check nonexistent-keys check


andrewtavis:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

In recent versions of [i18n-check](https://github.com/activist-org/i18n-check) we've added the ability for the `nonexistent-keys` check to search for keys in directories that are not the main `src-dir`. We've just updated i18n-check to the latest version in the project, and using this functionality we're seeing that there are some non-existent keys in the frontend testing directories.

This issue would do the following:
- Uncomment `i18n-check` `nonexistent-keys: search-dirs` arguments found in the configuration file [here](https://github.com/activist-org/activist/blob/41a58ff8ab039b4a62fa553b3dcd3b988e22a0da/.i18n-check.yaml#L16)
- Run `i18n-check -nk` for the `nonexistent-keys` check (we need to turn on the uv venv for this with `source .venv/bin/activate` in the backend directory)
- Fix all of the nonexistent keys

Current output is:

```
> i18n-check -nk           
❌ nonexistent-keys error: There are 13 i18n keys that are not in the en-US.json i18n source file. Please check the validity of the following keys:

i18n._global.participate_in_event_aria_label
i18n.components.form_resource._global.edit_resource
i18n.components.form_resource._global.save
i18n.components.form_resource._global.update
i18n.components.form_social_link._global.edit_links
i18n.components.form_social_link._global.save
i18n.composables.use_menu_entries_state.${subpage}
i18n.test.button
i18n.test.dropdown
i18n.test.external
i18n.test.label
i18n.test.link
i18n.test.main
```

The above keys are found in the `frontend/test` and `frontend/test-e2e` directories that are now being searched through.

andrewtavis:
CC @aasimsyed 👋 Would you have an interest in looking into this? Please let us know what your opinion of the setup for searching the `frontend/test` and `frontend/test-e2e` is!

andrewtavis:
CC @sh-ran for the discussion here in case there's future work that needs to be done :)

andrewtavis:
[...]

PR #2115, opened 2026-04-25
chore(frontend): drop i18n.test.* placeholders from button specs

aasimsyed:
Replace i18n.test.* placeholder strings in the 5 button spec files with plain descriptive strings, and update the comment in test/setup.ts since the example it cited no longer exists.

The runtime missingWarn/fallbackWarn suppression in setup.ts has to stay: button components call t(ariaLabel) at runtime, so vue-i18n warns for any unresolved string regardless of shape. The static-check fix and the runtime suppression are separate concerns.

Prep work for activist-org/activist#2114 so enabling nonexistent-keys: search-dirs in i18n-check no longer flags these 6 keys.

### Contributor checklist

---

## What

Renames the 6 `i18n.test.*` placeholder strings in the 5 button spec files to plain descriptive strings (e.g. `"tag button aria"`), and updates the comment in `test/setup.ts` since the example it cited won't exist anymore.

## Why

Prep work for #2114. Once `nonexistent-keys: search-dirs` gets enabled in `i18n-check`, those 6 keys would get flagged. They're not real translations though — the button templates do `:aria-label="$t(ariaLabel)"`, and the tests just assert prop pass-through (class names, `aria-label` is set, etc.) without caring about the rendered string. Plain strings serve the same purpose and drop off the static checker since `i18n-check` only flags `i18n.*`-shaped strings.

## On the runtime suppression

Tried dropping `missingWarn: false` / `fallbackWarn: false` from `test/setup.ts` first and it doesnt work — `$t("tag button aria")` still misses at runtime and warns, and at least one unrelated test (`DropdownTheme`) relies on the same suppression. Static-check fix and runtime suppression are separate concerns, so this PR only handles the first.

## Validation

- `yarn test` — 1855 passed, no regressions
- `yarn prettier --check` on chang

### Related issue

- #2114

### DIFF
--- a/frontend/test/components/btn/BtnTag.spec.ts
+++ b/frontend/test/components/btn/BtnTag.spec.ts
@@ -9,7 +9,7 @@ describe("BtnTag", () => {
   const defaultProps = {
     cta: true,
     fontSize: "base" as const,
-    ariaLabel: "i18n.test.button",
+    ariaLabel: "tag button aria",
   };
 
   it("renders button with aria-label", async () => {
--- a/frontend/test/components/btn/action/BtnAction.spec.ts
+++ b/frontend/test/components/btn/action/BtnAction.spec.ts
@@ -9,7 +9,7 @@ describe("BtnAction", () => {
   const defaultProps = {
     cta: true,
     fontSize: "base" as const,
-    ariaLabel: "i18n.test.button",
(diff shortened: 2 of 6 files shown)


andrewtavis (approved):
praise: All's looking good here, @aasimsyed! Thanks for the initial work here 😊

PR #2122, opened 2026-04-28
first implementation of @pinia/colada with getEvent and icon images m…

### LINKED ISSUE #2111: Refactor: adopt @pinia/colada for image CRUD and extract from useFileManager
### Terms

### Description

PR #2109 implemented a fix for the bug where deleting an image from the upload modal left the carousel stale until a full reload (#2108, #1791). It works, but the fix sits inside the modal components because `useFileManager` is entity-agnostic and literally can't know which cache to invalidate.

During PR #2109 review, @nicki182 raised the idea of moving the refresh logic into useFileManager itself. After digging into why that doesn't cleanly work (entity-agnostic file manager vs entity-specific caches), shifted toward adopting a proper server-state library.

### The plan

Move image CRUD (fetch / upload / update / delete + cache) onto `@pinia/colada`, keep `useFileManager` for purely local file and drop-zone state, and delete the manual `refreshXData()` wiring we're currently threading through the upload modals.

### Why `@pinia/colada` and not the other options

For this codebase I think colada wins because:

- The team already knows Pinia, so colada is ~zero new mental model.
- Migrating `useAsyncData` to colada is close to a 1:1 swap. Migrating to vue-query is more of a rewrite because the idioms are further away.

### Rough plan

1. **Spike + write-up** Read through colada properly, sketch what a base `useEntityImages` composable looks like, figure out the cache key structure, and confirm Nuxt SSR + hydration isn't broken. Drop the results as a comment on this issue so we have something concrete to review before any real code moves.
2. **Divide the work.** Once the spike is ok'd, split along clean seams (e.g. base composable + org migration in one chunk, group + old composable cleanup in another). Can be one PR or a short series depending on what the spike turns up.
3. **Migrate incrementally.** New composables land alongside the old ones. Modals move one at a time. Old mutation and query composables only get deleted once the last caller is off them.

### In scope

- `frontend/app/composables/useFileManager.ts` (slim down to local file / UI state).
- `frontend/app/composables/mutations/useOrganizationImageMutations.ts` and `useGroupImageMutations.ts` (replaced).
- `frontend/app/composables/queries/useGetOrganizationImages.ts` and `useGetGroupImages.ts` (folded in).
- Modal wiring added in #2109 for cache refresh (goes away once colada is handling it).
- `frontend/app/services/content/image.ts` and the org/group image services stay as the raw HTTP layer the new composable calls into.

### Related

- PR #2109 (the narrow fix that triggered this conversation)
- Issue #2108 (carousel stale after delete)
- Issue #1791 (original reactivity bug)
- Discussion with @nicki182 on PR #2109

cc: @nicki182 

### Contribution

_No response_

### LINKED ISSUE #2245: Bug: Event edit/create modals stay open until data refetch finishes after saving
### Terms

### Behavior

When you edit an item and press submit, the edit modal visibly hangs open for several seconds after the save already succeeded.

**Steps to reproduce (reliable):**

1. Sign in as a user who can edit (e.g. `admin` / `admin` on the dev stack).
2. Open an event and go to its **FAQ** tab; make sure it has at least one entry.
3. Open browser DevTools -> **Network** tab -> set throttling to **Slow 3G** (leave DevTools open).
4. Click the pencil (edit) icon on an FAQ entry, change the text, and press **Submit**.

**Actual:** the save goes through, but the edit modal stays open for several seconds (as long as the background refresh takes), then finally closes.
**Expected:** the modal closes immediately once the save succeeds; refreshing the page data should not hold the modal open.

Without throttling the delay is milliseconds and easy to miss. This is why it shows up as flaky in CI, where the runner's timing is slow enough to trip it, in:

- `frontend/test-e2e/specs/all/events/event-faq/event-faq-page.spec.ts` - "User can manage FAQ entries (CREATE, UPDATE, DELETE)" (the edit modal is still visible after saving)
- `frontend/test-e2e/specs/all/events/event-about/event-about-details-permissions.spec.ts` - "Member can change the organization and see it on the details card"

nicki182:
…utations within event

### Contributor checklist

---

### Description

Based on the disscussion we had @aasimsyed and I in #2111 this is an initial review of the refactor we are planning to make and make sure we are both in the same page as implementation goes for the future refactor.

### Related issue

- Closes #2111
- Closes #2245
- Closes #2247 
- Closes #2251
cc @aasimsyed


nicki182:
@aasimsyed I made changes in useGetEvent and in useEventImageIconMutations can you check if other mutations are ok when making changes like faq entries within event? since now I'm using the new library and havent checked on that and if there is issues then this issue should cover all event mutations too which I don't have issue of doing so

nicki182:
@aasimsyed the e2e tests are breaking and I checked it so yeah... I need to make the changes to all mutations. If you think this looks good I'll move to the rest of it. And iomce it's brought in maybe you do organization and groups and I'll do whichever ones are missing like useLocation and see the implementation of this for pagination

frontend/app/composables/queries/useGetEvent.ts
```diff
@@ -2,70 +2,26 @@
 // Read a single event with useAsyncData. Store-first, then fetch if missing.
```
aasimsyed: This should be changed now to ```// Read a single event through Pinia Colada.``` shouldn't it?

frontend/app/composables/mutations/useEventImageIconMutations.ts
```diff
@@ -1,48 +1,41 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 // Mutation composable for FAQ entries - uses direct service calls, not useAsyncData.
```
aasimsyed: this should be 
```
// Mutation composable for icon image mutations
```

Seems like a copy/paste issue

aasimsyed:
praise: [@nicki182](https://github.com/nicki182) the `EVENT_KEYS.byId` + `useQuery` + `queryCache.invalidateQueries` pattern looks right to me, and agree that the remaining event mutations need the same treatment before the Nuxt and Colada caches stop conflicting. Happy with the work split too. Looks good to me, but I did have a few notes:

---

### Setup

**`modules.ts`**: I might be missing something, but I don't see `@pinia/colada-nuxt` in the modules list. Is Colada registered somewhere else? The [installation docs](https://pinia-colada.esm.dev/nuxt.html#Installation) say to add it via `modules: ['@pinia/colada-nuxt']` in `nuxt.config.ts` alongside `@pinia/nuxt`, otherwise `useQuery`/`useQueryCache` won't be wired up. Just flagging it in case it still needs to be added.


---

### Suggestion

**Shared invalidation helper**: once all the event mutations are migrated, every `onSettled` will call the same `queryCache.invalidateQueries({ key: EVENT_KEYS.byId(unref(eventId)) })`. Would it be worth extracting that into a small helper in `useGetEvent.ts`, something like:
```ts
export function invalidateEventCache(queryCache: ReturnType<typeof useQueryCache>, id: string) {
  return queryCache.invalidateQueries({ key: EVENT_KEYS.byId(id) });
}
```
That way the key logic lives in one place and mutations just call `invalidateEventCache(queryCache, unref(eventId))`. Not a blocker, just a thought for when the full migration lands.

---

### Stale comments

**`useGetEvent.ts` lines 2-3**: the header comment still says "useAsyncData. Store-first, then fetch if missing" but it now uses `useQuery` and no longer writes to the store. Small thing, but could confuse contributors landing on the file cold.

**`useEventImageIconMutations.ts` line 2**: says "Mutation composable for FAQ entries", looks like copy/paste drift from the FAQ mutations file.

---

### Unit tests

**Unit tests for `useEventImageIconMutations`**: the existing tests expect `uploadIconImage()` to return `true`/`false`, expect `refreshNuxtData(getKeyForGetEvent(...))` to be called, and call `refreshEventData` as a public method. All three break after this change. I think these need to be updated, too.

nicki182:
I forgot to do that. I added the library and that's it. It seemed to work properly but I'll add this in since maybe something might not be working properly without it. I can change the comments no worries, it was just to have an initial overview

nicki182:
@aasimsyed take a look with the latest changes for event query and mutations. It looks like it works

nicki182:
@aasimsyed if it's good I'll fix the tests

aasimsyed:
@nicki182 praise: latest change looks right to me: `@pinia/colada-nuxt` in `modules.ts`, `useEventCache` for invalidation, and migrating the rest of the event mutations so FAQ/resources/etc are not left on the old refresh path. That should stop Colada and Nuxt data fighting each other.

Green light to keep going on the broader migration and we can split org/groups vs location/pagination like you suggested.

nicki182:
@aasimsyed so let me work on this and bring it in for tomorrow and you can start with organization and groups

nicki182:
@aasimsyed regarding the issue you find with the mutations in #2251, do you think it could be solved if we just bring this in?

andrewtavis:
If we can finalize this as a way of closing out the e2e issues and then we don't need #2251 or #2247, then this sounds like the most sustainable solution :)

aasimsyed:
@nicki182 @andrewtavis agreed, let's finish this instead of merging #2251/#2247.

Confirming: Colada should fix the same root cause for good, since it drops the old refresh path entirely. Once orgs and groups are migrated too, both PRs become unnecessary.

nicki182:
@aasimsyed wanna take a look on the e2e tests? the rest is done. So we can bring this in and finish up the rest. I can start working and group and organizations if this is ok

aasimsyed:
I checked out this branch and ran the E2E suite. Most of the red looks like the flaky tests we already knew about (social links, pagination, org FAQ), those pass on retry so probably unrelated to this PR.

But `event-create-modal.spec.ts`, `event-faq-server-errors.spec.ts`, and `event-resources-server-errors.spec.ts` fail every single time, on both Desktop and Mobile, with no retry saving them. So I think something real is going on there.

A couple things I noticed that might help narrow it down:

- For the create-event test, the page stays on `/home` instead of moving to the new event's page, so something about getting the created event back seems off.
- For the FAQ/resource error tests, the modal seems to close right away even when the request fails, so it doesn't look like it's waiting on the result before closing.
- Unit tests are all green for these composables, so whatever it is isn't caught there, worth double checking if the test mocks for mutations behave the same way the real library does.

My guess is it's related to how `mutate` vs `mutateAsync` behave differently in `@pinia/colada`, maybe take a look at that in the composables that were touched here.

nicki182:
@aasimsyed I made new changes, you can take look at them.

aasimsyed:
@nicki182 Checked out the latest commit and ran the E2E suite again.

Great!  `event-create-modal.spec.ts` "closes modal on submit", `event-faq-server-errors.spec.ts`, and `event-resources-server-errors.spec.ts` all pass now. Those were failing every time before, so that part looks fixed.

There is one new failure though. `event-create-modal.spec.ts` "shows error toast when create event returns 500" now fails every time, on both Desktop and Mobile. The test error says two error toasts show up instead of one:

```
strict mode violation: locator('[data-sonner-toast][data-type="error"]...') resolved to 2 elements
```

This test was passing before this latest commit, so it looks like something in the latest changes made it start failing.

Everything else still failing is the same flaky stuff we already knew about: `organization-about-social-links.spec.ts` (hard fails every retry), `organization-group-about-social-links.spec.ts` (flaky, passes on retry), and `organization-faq-page.spec.ts` on mobile (also a known flake). Once this PR covers organizations and groups too, it should close out the flakiness of those two.

Not sure yet what's causing the second toast, but one guess: `create` in `useEventMutations` now throws on failure instead of resolving to something like `false`, so maybe the flow that calls it is treating the failure as its own separate error and showing a toast for it too, on top of the one the composable already shows. Could be worth checking if making `create` swallow the error and just return `false` (like the group and org create composables already do) makes the second toast go away.

nicki182:
@aasimsyed I don't mind taking a look later about this and fix it

nicki182:
@aasimsyed I guess once that's fix we can bring this in and continue working with organizations and groups

nicki182:
@aasimsyed fixed :wink: I think its ready to be brought in

andrewtavis:
Thanks, @nicki182! I'll take a look at this later as well :) @aasimsyed, let us know what your thoughts are 😊

aasimsyed:
@nicki182 looks good. let's bring it in.

andrewtavis:
Have updated the PR commit message to close #2111 and #2245, as well as #2247 as that PR won't be needed after this (to my understanding).

andrewtavis:
And #2251 will be closed as well :)

nicki182:
Don't close them yet. This is the initial faze of the refactor. I'm going to make 1 more PR and for pagination and @aasimsyed should work on the rest of the services. This is to see how to approach the refactor since it's refactoring the whole services layer

andrewtavis (approved):
praise: All's looking and functioning really well, @nicki182! Thanks for taking this on to improve the infrastructure so much and close the bug with it!

PR #2124, opened 2026-04-28
ci(e2e): start filescan service in E2E workflows and wait for it to be ready

### LINKED ISSUE #2123: ci(e2e): filescan service not started in E2E workflows, causing all image uploads to fail
### Terms

### Behavior

Both `ci_e2e_playwright_desktop.yaml` and `ci_e2e_playwright_mobile.yaml` start the Docker stack with:
```yaml
docker compose --env-file .env.dev up --build -d backend db
```
The `filescan` service is never started. Django's image upload view (`content/views.py`) calls `scan_uploads_and_rewind` before saving any file, which POSTs to `http://filescan:9101/scan`. Since that hostname is unreachable in CI, `httpx` throws a `RequestError`, which Django wraps as a 400 response:
```json
{"nonFieldErrors": ["The file could not be scanned. Please try again later."]}
```
Any E2E test that uploads an image will fail in CI, regardless of whether the app and test code are correct. The failure is not obvious because `uploadImages()` on the frontend catches errors internally and closes the modal anyway, so tests fail at a downstream assertion rather than at the upload step itself.

**Steps to reproduce**
Any E2E test that uploads an image via the modal (e.g. `organization-about-image-upload.spec.ts`).

**Expected behavior**
The `filescan` service is running and reachable at `http://filescan:9101` so Django can scan uploaded files and return 200.

**Proposed fix**

In both `ci_e2e_playwright_desktop.yaml` and `ci_e2e_playwright_mobile.yaml`:

1. Add `filescan` to the compose startup command:```yaml
run: docker compose --env-file .env.dev up --build -d backend db filescan
```

2. Add a wait step after "Wait for Backend":
```yaml
- name: Wait for Filescan
  run: |
    timeout 180 bash -c 'until curl -s -o /dev/null http://localhost:9101/health; do sleep 3; done'
```

The wait uses `curl` without `-f` intentionally. The `/health` endpoint returns 503 even when the service can scan real files. Using `-f` would block indefinitely. Checking for any HTTP response is sufficient because the entrypoint only starts uvicorn after confirming the ClamAV socket exists, so if the server responds at all, real file scans will work.

**Affected files**
- `.github/workflows/ci_e2e_playwright_desktop.yaml`
- `.github/workflows/ci_e2e_playwright_mobile.yaml`

aasimsyed:
The filescan service was not included in the docker compose startup for either E2E workflow. Django's image upload view calls scan_uploads_and_rewind before saving any file, which POSTs to http://filescan:9101/scan. With the service absent, httpx throws a RequestError that Django wraps as HTTP 400 ("The file could not be scanned."), causing every image upload test to fail in CI regardless of whether the app and test code are correct.

Add `filescan` to the compose command and poll http://localhost:9101/health until uvicorn responds before running Playwright tests. The wait uses curl without -f because the /health endpoint permanently returns 503 (the ClamAV instream protocol rejects the empty-byte probe used by the health check) even when real file scans work correctly. A plain HTTP response is sufficient since the entrypoint only starts uvicorn after confirming the ClamAV socket exists, so any response means actual uploads will succeed.

### Contributor checklist

---

## Summary

Both `ci_e2e_playwright_desktop.yaml` and `ci_e2e_playwright_mobile.yaml` only started `backend` and `db` in their Docker Compose step. The `filescan` service was never started.

Django's image upload view calls `scan_uploads_and_rewind` before saving any uploaded file, which POSTs to `http://filescan:9101/scan`. With the service absent, `httpx` throws a `RequestError` that Django wraps as HTTP 400:

```json
{"nonFieldErrors": ["The file could not be scanned. Please try again later."]}
```

This caused every E2E test that uploads an image to fail in CI regardless of whether the app and test code were correct. The failure was not obvious because the frontend's `uploadImages()` catches errors internally and closes the modal anyway, so tests failed at a downstream assertion (`data-slide-count` staying at the placeholder value) rather than at the upload step itself.

## Changes

- Add `filescan` to the `docker compose up` command in both workflows
- Add a "Wait for Filescan" step that polls `http://localhost:9101/health` until uvicorn responds before Playwright runs

The wait uses `curl` without `-f` intentionally. The `/health` endpoint returns 503 even when the service can scan real files. Using `-f` would block indefinitely. A plain HTTP response (any status) is sufficient: the service entrypoint only starts uvicorn after confirming the ClamAV socket exists, so if the server responds at all, real file scans will succeed.

Tested on my fork and it passes CI

### Related issue

- Discovered while debugging #1665.
- Closes #2123

### DIFF
--- a/.github/workflows/ci_e2e_playwright_desktop.yaml
+++ b/.github/workflows/ci_e2e_playwright_desktop.yaml
@@ -54,7 +54,7 @@ jobs:
           ! nc -z localhost 5432
 
       - name: Run Docker Compose
-        run: docker compose --env-file .env.dev up --build -d backend db
+        run: docker compose --env-file .env.dev up --build -d backend db filescan
 
       - name: Wait for Database
         run: |
@@ -63,6 +63,10 @@ jobs:
       - name: Wait for Backend
         run: |
           timeout 180 bash -c 'until curl -s http://localhost:8000/v1/auth/sign_in -o /dev/null -w "%{http_code}" | grep -qE "^(200|405)"; do sleep 3; done'
+
+      - name: Wait for Filescan
+        run: |
+          timeout 180 bash -c 'until curl -s -o /dev/null http://localhost:9101/health; do sleep 3; done'
 
       - name: Setup Node Environment
         uses: actions/setup-node@v5
--- a/.github/workflows/ci_e2e_playwright_mobile.yaml
+++ b/.github/workflows/ci_e2e_playwright_mobile.yaml
@@ -54,7 +54,7 @@ jobs:
           ! nc -z localhost 5432
 
       - name: Run Docker Compose
-        run: docker compose --env-file .env.dev up --build -d backend db
+        run: docker compose --env-file .env.dev up --build -d backend db filescan
 
       - name: Wait for Database
         run: |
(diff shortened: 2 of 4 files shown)


aasimsyed:
added a small change to address flaky test

andrewtavis (approved):
praise: All's looking great, @aasimsyed! Thanks for getting this in to unblock @UnknownSean8 :)

PR #2128, opened 2026-05-01
fix: correct bugs, typos, and misconfigurations across backend and docs

techvoyager-varun:
I went through the codebase and noticed a few bugs, typos, and misconfigurations that needed fixing. Here's what I changed:

CONTRIBUTING.md

I renamed the duplicate Deployment heading to  Testing since it lists pytest, Vitest, and Playwright — clearly testing tools, not deployment.

backend/core/settings.py

I moved CorsMiddleware higher up in the MIDDLEWARE list, right after SecurityMiddleware. It was sitting below CsrfViewMiddleware which goes against the django-cors-headers setup guide.
I removed a duplicate CommonMiddleware entry that was listed twice.
I changed the LOGGING dictionary's "version" from 1.0 (float) to 1 (int) to match what Python's dictConfig actually expects.

backend/authentication/views.py

I fixed a grammatically broken response message — "Email confirmed is verified." is now "Email is verified.".
I fixed another broken message — "Password reset has been successfully." is now "Password has been reset successfully.".

backend/events/views.py

I changed the error message in EventFaqViewSet.destroy() from "Org is None." to "Event is None." since we're dealing with events here, not organizations.
I renamed a misleading variable faq to resource inside EventResourceViewSet.update() — it was holding an EventResource object, not a FAQ. Looks like it was left over from a copy-paste.
I fixed a spelling mistake in the class name — EventCalenderAPIView is now EventCalendarAPIView.

backend/events/urls.py

I updated the import to replace EventCalenderAPIView with EventCalendarAPIView.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,9 +74,9 @@ The following are the current and planned technologies for [activist.org](https:
 
 - [Docker](https://www.docker.com) • [Netlify](https://www.netlify.com)
 
-### Deployment
+### Testing
 
-[pytest](https://docs.pytest.org/en/stable/) (backend) • [Vitest](https://vitest.dev/) (frontend) • [Playwright](https://playwright.dev/) (end to end)
+- [pytest](https://docs.pytest.org/en/stable/) (backend) • [Vitest](https://vitest.dev/) (frontend) • [Playwright](https://playwright.dev/) (end to end)
 
 ### Localization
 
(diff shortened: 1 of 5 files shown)


andrewtavis (approved):
praise: Thanks a lot for the fixes here, @techvoyager-varun! Really is a nice PR. Appreciate the detailed description as well!

PR #2129, opened 2026-05-01
add authentication enforcement and proper authorization to EventDetailAPIView

techvoyager-varun:
I noticed that EventDetailAPIView was missing permission_classes entirely, which means the PUT and DELETE endpoints had no explicit authentication enforcement at the class level. The view was relying solely on a manual request.user != event.created_by comparison, but without DRF-level permission enforcement, this is fragile.

What I changed

backend/events/views.py

Added a get_permissions() method that returns AllowAny for GET requests and IsAuthenticated for PUT/DELETE — this ensures DRF enforces authentication before the view logic even runs.
Changed the authorization check status code from 401 Unauthorized to 403 Forbidden. HTTP 401 means "not authenticated" while 403 means "authenticated but not allowed" — the existing code was using the wrong one since the user is authenticated but just isn't the owner.
Added not request.user.is_staff to the authorization check so staff/admin users can also update and delete events. Every other ViewSet in the project (EventFaqViewSet, EventResourceViewSet, OrganizationFaqViewSet, etc.) already allows staff access, so this makes it consistent.
Updated the OpenAPI schema annotations to reflect the corrected status codes.

backend/events/tests/test_event_update.py

Expanded from 1 test case to 3: unauthenticated user gets 401, non-owner gets 403, and the event owner can successfully update (200).

backend/events/tests/test_event_delete.py

Expanded from 1 test case to 3: unauthenticated user gets 401, non-owner gets 403, and the event owner can successfully delete (204).


andrewtavis (approved):
praise: A second really nice PR, @techvoyager-varun :) Would you be able to do similar checks of the other DetailAPIViews that we have? Really thanks so much for the fixes and detailed PRs!

techvoyager-varun:
Hi @andrewtavis yes surely i will do that.

techvoyager-varun:
Hi @andrewtavis i have implemented the requested changes.

andrewtavis:
Thanks so much, @techvoyager-varun!

PR #2133, opened 2026-05-06
fix(backend): add authentication enforcement and proper authorization to OrganizationDetailAPIView

techvoyager-varun:
Follow-up to #2129 as requested by @andrewtavis — applying the same authentication enforcement pattern to `OrganizationDetailAPIView`.

I found that `OrganizationDetailAPIView` had the same issues that `EventDetailAPIView` had before #2129:

### What I changed

**`backend/communities/organizations/views.py`**
- Added a `get_permissions()` method that returns `AllowAny` for `GET` requests and `IsAuthenticated` for `PUT`/`DELETE` — this ensures DRF enforces authentication before the view logic runs.
- Changed the authorization status code from `401 Unauthorized` to `403 Forbidden`. HTTP 401 means "not authenticated" while 403 means "authenticated but not allowed" — the existing code was using the wrong one since the user *is* authenticated but just isn't the owner.
- Updated the OpenAPI schema annotations to reflect the corrected status codes (including `200` → `204` for the delete endpoint).

**`backend/communities/organizations/tests/test_org_update.py`**
- Refactored the single monolithic test into 4 focused test functions: unauthenticated user gets 401, non-owner gets 403, non-existent org gets 404, and org owner can successfully update (200).

**`backend/communities/organizations/tests/test_org_delete.py`**
- Refactored the single monolithic test into 3 focused test functions: unauthenticated user gets 401, non-owner gets 403, and non-existent org gets 404. The owner-delete success test is intentionally omitted because soft-delete requires StatusType fixtures that aren't set up in the test environment (same reason the original test had it commented out).

### DIFF
--- a/backend/communities/organizations/tests/test_org_delete.py
+++ b/backend/communities/organizations/tests/test_org_delete.py
@@ -8,7 +8,6 @@ from uuid import uuid4
 import pytest
 from django.test import Client
 
-from authentication.factories import UserFactory
 from communities.organizations.factories import OrganizationFactory
 
 pytestmark = pytest.mark.django_db
--- a/backend/communities/organizations/tests/test_org_delete.py
+++ b/backend/communities/organizations/tests/test_org_delete.py
@@ -1,4 +1,8 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+Test cases for deleting organizations.
+"""
+
 from uuid import uuid4
 
 import pytest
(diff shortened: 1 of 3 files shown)


andrewtavis (approved):
Thanks so much for the further updates here, @techvoyager-varun! Really great to have them :) Do you want to take a look at the groups code as well?

techvoyager-varun:
Hi @andrewtavis yes surely i will do that.

techvoyager-varun:
Hi @andrewtavis sir,
Could I get an internship opportunity at activist-org?I would be highly grateful for the chance to work with you and contribute to the organization.

andrewtavis:
Hi @techvoyager-varun 👋 Thanks for your interest in contributing in a more formal manner. We're not offering internships now, but hope to in the future. We can certainly have the conversation then :)

andrewtavis:
And no need to call me sir :)

Issue #1983, opened 2026-03-10
[Epic] Add E2E tests for error and empty states across entity pages


aasimsyed:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

#### Background

As identified in the E2E coverage audit tracked by #1959, error and empty states are **entirely untested** across all entity pages. These are Tier 3 quality gaps — the most common real-user failure modes — and their absence means regressions in empty state rendering, form validation errors, and server-side error handling would go undetected in CI.

The current test suite exercises only the happy path: authenticated admin, seed data present, network healthy. No spec asserts what a user sees when:
- An entity has no FAQs, resources, or social links yet (empty state)
- A form is submitted with invalid input (validation error state)
- A network or server error occurs during a mutation (error state)

This Epic tracks the full set of subtasks needed to close this gap across events, organizations, and organization groups.

---

##### 1. Empty states

Empty states are rendered by the `<EmptyState>` component and appear when a list has zero items. The `permission` prop changes the message shown to admins vs non-editors — this branching is also untested.

**Pages to cover:**

| Page | Empty condition |
|---|---|
| `/events/[id]/faq` | Event has no FAQ entries |
| `/events/[id]/resources` | Event has no resources |
| `/organizations/[id]/faq` | Organization has no FAQ entries |
| `/organizations/[id]/resources` | Organization has no resources |
| `/organizations/[id]/groups/[id]/faq` | Group has no FAQ entries |
| `/organizations/[id]/groups/[id]/resources` | Group has no resources |
| `/events` | No events match the active filter |
| `/organizations` | No organizations match the active filter |

**Test cases per page:**
- Empty state component is visible when the list is empty
- Empty state message is appropriate for the page type
- Empty state shows the correct variant for admins vs non-editors
- Empty state is NOT shown when at least one item exists

---

##### 2. Form validation error states

Currently only successful submissions are tested. Validation errors must be explicitly exercised.

**Forms to cover:**

| Form | Invalid input scenarios |
|---|---|
| FAQ create/edit modal | Empty question field, empty answer field |
| Resource create/edit modal | Missing name, missing URL, malformed URL |
| Social links modal | Malformed URL, missing label |
| Event create modal (`ModalCreateEvent`) | Missing title, missing date |
[...]

PR #2146, opened 2026-05-14
test(e2e): empty list when events or org filters match nothing

### LINKED ISSUE #2145: Add E2E empty state tests for events and organizations listing pages (no filter results)
### Terms

- [x] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

#### Background

Subtask of #1983. The epic requires E2E coverage when **no rows match the active filters** on the **events** and **organizations** catalog pages: users should see the shared empty state (and no false “error” state) when the list API returns success with `results: []` for the **current query string**, not only when the global catalog is empty or when the list endpoint fails.

Today, `frontend/test-e2e/specs/all/events-list-api-errors.spec.ts` covers list **500** and a **fully empty** unfiltered list mock; `events-filter.spec.ts` exercises URL filter parameters but does not assert empty-state UX when filters exclude all events. Organizations home specs assert cards vs empty state against **live** data rather than a routed “filtered empty” response.

This subtask adds deterministic Playwright specs using `page.route()` so that for a **specific filter query** (or equivalent client request shape) the list API returns **HTTP 200** with `count: 0` and `results: []`, then asserts `data-testid="empty-state"` (or the catalog empty component used on that page), appropriate copy, and **no** error toast. Optionally assert that clearing filters or changing parameters restores non-empty results if the spec seeds a second mock pass-through.

---

#### Scope

**Surfaces:**

| Page | Condition to simulate |
|---|---|
| `/events` (e.g. list view) | Active filters yield **zero** matching events (`results: []` for the request the UI makes with current query params) |
| `/organizations` | Active filters yield **zero** matching organizations |

**Endpoints to mock (align URL patterns with the app):**

- Events list — public list endpoint used by `listEvents` (see `EVENTS_LIST_ROUTE` / `**/api/public/events/events**` in existing list specs); match **GET** and optionally the query string so only the “filtered empty” scenario is fulfilled while other navigations continue.
- Organizations list — public organizations list endpoint used by the organizations home/catalog (discover the exact path from `listOrganizations` or equivalent service and mirror existing `organizations-home-page` navigation).

---

#### Test cases

**Events**

1. Apply one or more filters (location, type, days ahead, etc.) via the existing filter page object or URL so the UI issues a list request that includes those params
2. Mock that list request to return `200` + `{ results: [], count: 0, next: null, previous: null }`
3. Assert empty state is visible and **no** error toast is shown
4. (Recommended) Assert that with a different mock or `unroute`, non-empty results hide the empty state

**Organizations**

1. Apply a filter that maps to query params on the organizations catalog (or set URL directly if that is how the app loads)
2. Mock the organizations list GET for that query to return an empty `results` array with `200`
3. Assert empty state is visible and **no** error toast
4. (Recommended) Same “clear filter / non-empty” check as for events where the product supports it

---

#### Spec file locations

```
frontend/test-e2e/specs/all/events-list-empty-filtered.spec.ts
frontend/test-e2e/specs/all/organizations-list-empty-filtered.spec.ts
```

---

#### Implementation notes

**Route matching** — Use precise `page.route` URL predicates or glob + `URLSearchParams` checks so unrelated list requests in the same test are not accidentally fulfilled. Prefer `afterEach` → `page.unrouteAll()`.

**Stability** — Follow patterns in `events-list-api-errors.spec.ts` (timeouts, heading wait) and organizations home specs for post-navigation assertions.

**Tags** — `@desktop` and `@mobile` where the catalog is supported on both; otherwise match project conventions for list pages.

### Contribution

Part of #1983 — E2E error and empty state coverage. Complements closed #2073 / #2143 (entity subpage empty states) and existing list API error coverage. Parent audit context: #1959.

aasimsyed:
Add Playwright specs that mock list GETs only for distinct filter query values and assert empty state without error toasts.

### Contributor checklist

---

## Summary

Two small Playwright specs cover the “filters are on but nothing matched” case, not just an empty site overall.

**Events:** opens the list with a dummy `location` query param, intercepts the public events list call so only that filtered GET gets a 200 with an empty `results` array, then checks the empty state shows and no error toast fires.

**Organizations:** same pattern with a dummy `city` param and the public org list endpoint.
All deterministic, no seed data needed for these flows.

### Related issue

- Closes: https://github.com/activist-org/activist/issues/2145  
- Parent epic: https://github.com/activist-org/activist/issues/1983


nicki182:
@aasimsyed all good :wink:

PR #2185, opened 2026-05-31
chore(backend): isolate complexipy in its own dependency group

### LINKED ISSUE #2184: Backend dev container crash-loops building complexipy on Alpine/arm64 (no musllinux wheel)
### Terms

### Behavior
After `complexipy` was added to the backend (commit `6129e00a`), the backend Docker container fails to start and keeps restarting in a loop on Apple Silicon / ARM machines. This blocks local backend and end-to-end testing.

```
× Failed to build `complexipy==5.5.0` ... Rust not found ...
Error loading shared library libgcc_s.so.1 ... cargo: _Unwind_Resume: symbol not found
hint: `complexipy` was included because `backend:dev` depends on `complexipy`
```

The reason: `complexipy` does not ship a ready-to-use package for the kind of Linux image our backend container uses, so the container tries to build it from scratch on startup, fails, and restarts over and over.

**How to reproduce:** on an Apple Silicon / ARM machine, check out `main` at or after commit `6129e00a` and start the backend with Docker. The backend logs show a repeated "failed to build complexipy" error and the container restarts continuously.

**Expected:** the backend container starts normally, sample data still loads, and `complexipy` still runs where it's actually used (CI).

**Proposed fix:** stop the backend container from installing `complexipy` (which it never needs to run), while keeping it available for the CI complexity check and for developers who want to run it directly. The sample-data tools the container does need are unaffected.

aasimsyed:
### Contributor checklist

---

### Description

After `complexipy` was added, the backend Docker container would not start on Apple Silicon / ARM machines and kept restarting. It was trying to install `complexipy` on startup, which doesn't work on the type of image the container uses.

This change stops the container from installing `complexipy` (it never needs to run it), while keeping it available for the CI complexity check and for anyone who wants to run it directly. The tools the container actually needs to load sample data are untouched.

**Changes**
- `backend/pyproject.toml`: move `complexipy` into its own `complexity` group instead of the default `dev` group.
- `.github/workflows/ci_backend_static_analysis.yaml`: run the check with `uv run --group complexity complexipy .`.
- `backend/uv.lock`: updated to match.

**Testing**
- Rebuilt and started the backend container: it now starts normally, runs migrations, and loads the sample data (users, organizations, events, etc.).
- Confirmed the complexity check still works when run with the new group.

**Risk:** Low. Nothing about how the app runs changes.

### Related issue

- Closes #2184

### DIFF
--- a/.github/workflows/ci_backend_static_analysis.yaml
+++ b/.github/workflows/ci_backend_static_analysis.yaml
@@ -60,4 +60,4 @@ jobs:
 
       - name: Run complexipy - Cognitive Complexity Check
         if: always()
-        run: uv run complexipy .
+        run: uv run --group complexity complexipy .
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
 
 [dependency-groups]
 dev = [
-    "complexipy>=4.2.0",
     "factory-boy>=3.3.3",
     "i18n-check>=1.19.5",
     "mypy>=2.1.0",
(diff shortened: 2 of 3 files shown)


andrewtavis (approved):
praise: Thanks for the quick fix here, @aasimsyed! Makes total sense :)

PR #2221, opened 2026-06-18
Fix group selector pagination in create event

Shinu-Cherian:
### Contributor checklist

* [x] This pull request is on a separate branch and not the main branch
* [ ] I have run the relevant tests locally

---

### Description

This PR fixes an issue in the Create Event flow where the group selector only displayed the first page of groups for organizations that had more than 10 groups.

#### Changes made

* Enabled infinite scrolling in `FormSelectorComboboxGroups.vue`
* Fixed pagination accumulation logic in `useGetGroups.ts`
* Reset pagination state when filters change
* Added an integration test covering multi-page group loading

#### Result

Users can now load and select groups beyond the first 10 entries when creating an event.

---

### Related issue

Fixes #2200

PR #2226, opened 2026-06-25
Fix group selector pagination in create event

### LINKED ISSUE #2200: Group dropdown in "Create event" only ever shows an organization's first 10 groups
### Terms

### Behavior

When picking groups while creating an event, the group dropdown shows only the first 10 groups of the selected organization. Scrolling to the bottom of the list does not load any more, the remaining groups can never be seen or selected.

**Steps to reproduce:**

1. Sign in and start creating a new event.
2. In the details step, select an organization that has more than 10 groups.
3. Open the **groups** selector dropdown and count the entries.
4. Scroll to the bottom of the dropdown list and wait.

**Actual:** exactly 10 groups are listed and nothing more ever loads; the organization's other groups cannot be selected, and there is no hint that they exist.
**Expected:** scrolling to the bottom loads the next batch of groups until all of them are available.

Note: this needs an organization with more than 10 groups. The default dev seed creates only 2 groups per org, so reseed with e.g. `--groups-per-org 20` to observe it.

Shinu-Cherian:
## Summary

This PR fixes the pagination behavior in the Group selector used during event creation.

## Changes

- Enable infinite scrolling for the Group selector.
- Append newly fetched groups instead of replacing the existing list.
- Prevent duplicate groups while loading additional pages.
- Follow the existing pagination pattern used by the Organizations selector.

## Files changed

- frontend/app/components/form/selector/FormSelectorComboboxGroups.vue
- frontend/app/composables/queries/useGetGroups.ts

Fixes #2200


aasimsyed:
@Shinu-Cherian thanks for the contribution! This is a duplicate of your earlier
#2221, which has the same change and the full review history, so I think should close
this one in favor of that. Let's keep the discussion on #2221. 🙂

PR #2231, opened 2026-06-28
Update all frontend dependencies and fix static analysis issues

andrewtavis:
### Contributor checklist

---

### Description

Updates the frontend dependencies. WIP, with so for the static analysis checks being ran locally and fixes having been mode. We now need to see about vitest issues and making sure that the e2e tests work 😊

### Related issue

- Closing dependabot PRs


andrewtavis:
Quick note @aasimsyed: @nicki182 and I needed to remove the `sudo` from the `yarn install` instances for the e2e tests as `sudo` was defaulting to the system Node 22 rather than using 24 explicitly. This was discovered when working on another issue, but we felt like adding it in made sense so we we're using a consistent Node :)

andrewtavis:
Thanks for sending along the changes, @nicki182 and @aasimsyed! Let's look to merge this tonight :)

andrewtavis:
Looking at this right now with @nicki182, @aasimsyed :) We were a bit confused by some of the changes to the mutation files, so the choice was made to revert those for now. We still have some flaky e2e tests, but we'll bring this in as we think that it's important to get the updates in and go from there. Feel free to open a PR with some of the changes to the mutation files and the corresponding tests with a bit of a more detailed explanation of why the changes are needed 😊

Generally we're seeing the functionality on the platform working well, so if the modal isn't closing quick enough in certain cases, then maybe we can have the test wait a bit for it to happen rather than editing the mutations?

We're happy to go through this with you at some point soon! Looking forward to that :) :)

aasimsyed:
No worries at all on reverting them, totally fair to keep this moving and get the dependency updates in. Here's a short explanation to make it clearer, since I think a couple of these are actually user-facing and not just test timing.

I figured the modal-close one was the main thing to explain. The way I saw it, it isn't the modal closing a little slowly with the test giving up too early. What I think is happening is that right after a save, we tell the page to re-fetch its data, and we wait for that fetch to finish before anything else happens. The modal and that data share the same loading state, so my guess is that while the re-fetch is in flight the modal stays open instead of closing.

So my thinking with the change was just to reorder things: let the modal close first, then kick off the re-fetch a moment later. That's also why I didn't think a longer wait in the test would reliably fix it, since the modal is stuck waiting on a network request to come back, not on a timer. I believe that's the same thing that showed up as the About-section modal staying open in CI.

The second change felt worth a note too, since it's more subtle. My understanding is that after a save or delete we ask the page to re-fetch its data, but if that re-fetch lands at the same time as another one that just finished, the app can reuse the older result and hand back the data from before the change, so a deleted item can stick around in the list until you manually reload.

My idea with that fix was to clear the cached data for that page before re-fetching, so the app is forced to actually go get fresh data instead of reusing the stale copy. This one seems easy to miss when clicking around manually, because you'd have to delete something and look at the list almost instantly to catch it, which probably lines up with everything looking fine on the platform.

Where I think it failed in the suite: the FAQ and resource delete tests. After deleting an entry, the test checks that the list count drops (e.g. expects 0 items), and my suspicion is the deleted entry was still showing from the reused stale data, so the count came back as 1 and the test failed. That would also explain why it was flaky rather than a hard fail, since it should only happen when the two re-fetches collide, which is timing dependent, so it passed on some runs and failed on others.

Happy to open issues/PRs for these.

andrewtavis:
Thanks so much for the detailed explanation, @aasimsyed! Sounds like we have two issues here:

1. Closing the modal before fetch
2. Clear deleted date before refetching

Let's make issues for each of these, and please tag both @nicki182 and I in them for the discussion 😊

PR #2246, opened 2026-07-05
fix: close FAQ modal immediately on save success (#2245)

krishnasingh281:
### Contributor checklist

---

### Description

The FAQ mutation composables (`useEventFAQEntryMutations`,
`useOrganizationFAQEntryMutations`) were awaiting the background
cache-refresh call before returning success. Since the edit/create
modal waits for that return value before closing, this made the
modal visibly hang until the refetch finished — invisible on fast
connections but reliably reproducible under network throttling,
which is why it showed up as a flaky CI test.

Fix: the refresh is now fired in the background (not awaited) after
the actual save succeeds, with errors caught so a failed refresh
doesn't surface as an unhandled rejection. The modal now closes as
soon as the save itself completes.

### Related issue

- #2245

### DIFF
--- a/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
+++ b/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
@@ -18,7 +18,9 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
       // Service function handles the HTTP call and throws normalized errors.
       await createEventFaq(currentEventId.value, faqData as FaqEntry);
 
-      await invalidateCacheRefreshEventData();
+      // Don't block the caller (e.g. a modal closing) on the background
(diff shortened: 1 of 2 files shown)


andrewtavis:
Thanks for your interest in the project and the PR, @krishnasingh281! We're discussing in #2245 where we want to go with this, but we'll look into this PR after that :)

nicki182:
@krishnasingh281 Hi, I just reviewed this and I'm closing this because its redundant. The handleError is already handled in the try catch. As example:

```ts
    try {
      await deleteEventFaq(faqId);

      await invalidateCacheRefreshEventData();

      return true;
    } catch (err) {
      handleError(err);
      return false;
    } finally {
      loading.value = false;
    }
  }
```
thank you for your interest in the project :wink:

PR #2247, opened 2026-07-06
fix: close event modals before refresh completes

### LINKED ISSUE #2245: Bug: Event edit/create modals stay open until data refetch finishes after saving

ahfoysal:
## Summary
Event mutation helpers were waiting for the event data refresh before returning success, so modals that awaited those helpers stayed open until the refresh finished.

This keeps the save request awaited, but starts the event data refresh without blocking the modal close. It also matches the existing group and organization FAQ mutation behavior.

Fixes #2245

## Testing
- yarn typecheck
- yarn lint
- yarn prettier app/composables/mutations/useEventFAQEntryMutations.ts app/composables/mutations/useEventTextsMutations.ts app/composables/mutations/useEventResourcesMutations.ts app/composables/mutations/useEventSocialLinksMutations.ts app/composables/mutations/useEventImageIconMutations.ts --check --config ../.prettierrc --ignore-path ../.prettierignore

### DIFF
--- a/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
+++ b/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
@@ -18,7 +18,7 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
       // Service function handles the HTTP call and throws normalized errors.
       await createEventFaq(currentEventId.value, faqData as FaqEntry);
 
-      await invalidateCacheRefreshEventData();
+      void invalidateCacheRefreshEventData();
 
       return true;
     } catch (err) {
@@ -38,7 +38,7 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
       // Direct service call - no useAsyncData needed for mutations.
       await updateEventFaq(currentEventId.value, faq);
 
-      await invalidateCacheRefreshEventData();
+      void invalidateCacheRefreshEventData();
 
       return true;
     } catch (err) {
@@ -57,7 +57,7 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
     try {
       await reorderEventFaqs(currentEventId.value, faqs);
 
-      await invalidateCacheRefreshEventData();
+      void invalidateCacheRefreshEventData();
 
       return true;
     } catch (err) {
@@ -76,7 +76,7 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
     try {
       await deleteEventFaq(faqId);
 
-      await invalidateCacheRefreshEventData();
+      void invalidateCacheRefreshEventData();
 
       return true;
     } catch (err) {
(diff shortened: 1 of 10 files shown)


andrewtavis:
@aasimsyed, would you be able to do an initial check of this? :)

aasimsyed (changes requested):
Thanks for taking this on, @ahfoysal! Small, focused change that does the right thing for #2245: the save still finishes, but the refresh afterward no longer holds the modal open, so it closes right after submit. 

**On the red E2E checks, two different things are going on:**

1. The failing `organization-about-social-links` test is organization code this PR doesn't touch, so it's not from your change, just a separate flaky test. I'll re-run CI on my end; it shouldn't block this PR.

2. The `event-faq-page` "manage FAQ entries" test is tied to your change. It passes, but only on a retry. The wobbly step is the delete: right after deleting an entry, the list sometimes still shows it for a moment.

That second one is worth fixing here, and it's a one-liner per helper. Running the refresh in the background is what fixes the hanging modal, but our data layer (`dedupe: "defer"`) will occasionally skip a background refresh when edits happen back-to-back (create, update, delete, like this test does), so the list can look stale. Clearing the cached entry right before the refetch stops it from being skipped:

```ts
async function invalidateCacheRefreshEventData() {
  if (!currentEventId.value) return;

  const key = getKeyForGetEvent(currentEventId.value);
  clearNuxtData(key);
  await refreshNuxtData(key);
}
```

Same change in all five helpers. The modal still closes instantly (the caller keeps `void`); only the background refetch becomes reliable.

**nitpick (optional):** since the refresh runs in the background and isn't awaited, a failure would be silent. A small `.catch()` to log it would help future debugging. The org helpers have the same gap today, so no worries either way.

Nice work, happy to re-review once that's in.

ahfoysal:
Updated the event refresh helpers to clear the event cache key before refetching, while keeping the refresh in the background. Added the matching mutation spec coverage too.

andrewtavis:
Closed as #2122 superseded this :) Thanks for opening the PR, @ahfoysal!

Issue #2204, opened 2026-06-10
List shows duplicate or extra cards after changing filters on a paginated events/organizations list


aasimsyed:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Behavior

On the **Events** or **Organizations** list, if you scroll down and load more than one page of results, then change a filter (e.g. location or city), the list does not always reset cleanly. Sometimes you see too many cards at once, or the same items appear more than once, even though the filter should have replaced the list with a fresh first page of matching results.
### Steps to reproduce
1. Open **Events** (list view) or **Organizations**.
2. Scroll down until more than 10 items have loaded (page 2+).
3. Open the sidebar filters and apply a filter (e.g. set **Location** to `Berlin` on Events, or **City** to `Berlin` on Organizations).
4. Look at the list that remains on screen.
### Expected behavior
- The list resets to filtered results only.
- Each item appears once.
- You see at most one "page" of new results until you scroll again.
### Actual behavior
- The list can show more items than a single page should contain (e.g. 13 cards when the page size is 10).
- Duplicate entries can appear after the filter is applied.
- Stale, unfiltered items can linger briefly alongside filtered ones.
### When it shows up
Most noticeable after pagination: load page 2 first, then change filters. It is intermittent and depends on timing (e.g. how quickly the filter applies relative to in-flight list requests).

atharv01h:
Hi @aasimsyed , I've submitted a fix for this in ##2207 .

**Root cause:** `useGetEvents` and `useGetOrganizations` had a race condition .when a user scrolled to page 2+ and then changed a filter, the old page-N request was still in-flight. If it resolved *after* the new filtered fetch started, it passed the stale-data append condition and wrote duplicates into the store. A secondary issue was that `page.value` was reset to `1` only *after* the API call, so the stale request had already sent `page: N` to the backend.

**Fix:** Added a `fetchGeneration` counter to both composables:
- `filtersChanged` is computed **before** the API call, and `page.value` is reset to `1` immediately
- Each handler invocation stamps a `currentGeneration`
- After `await`, if a newer fetch has started (`currentGeneration !== fetchGeneration.value`), the stale response is discarded
- Also removed a redundant `store.setPage()` call in `useGetOrganizations`

[...]

PR #2248, opened 2026-07-09
test(e2e): add empty state tests for organization faq and resources (closes #2143)

### LINKED ISSUE #2143: Add E2E empty state tests for organization FAQ and resource pages
### Terms

- [x] I agree to follow activist's [Code of Conduct](https://github.com/activist-org/activist/blob/main/.github/CODE_OF_CONDUCT.md)

#### Background

Subtask of #1983. Organization FAQ (`/organizations/[id]/faq`) and organization resources (`/organizations/[id]/resources`) pages render `<EmptyState>` when embedded lists (`faqEntries`, `resources` on the organization payload from `getOrganization`) are empty, with different copy for admins (can add) vs non-editors. Event FAQ and resources already have dedicated empty-state E2E specs (#2073); the organization entity does not yet mirror that coverage and still relies on seed-dependent display branches in generic page specs.

This subtask adds Playwright specs that use `page.route()` to fulfill **HTTP 200** on the **public organization detail** GET with an organization JSON body whose `faqEntries` and `resources` arrays are empty (or overridden to include items for the “not empty” case), then assert empty-state visibility, messaging/variant, and that the empty state is **not** shown when at least one item is present.

---

#### Scope

**Pages:**

| Page | Empty condition |
|---|---|
| `/organizations/[id]/faq` | Organization has no FAQ entries |
| `/organizations/[id]/resources` | Organization has no resources |

**API to mock (align with the app):**

- `GET` organization by id via the public API base used by `getOrganization` (`withoutAuth: true` → `/api/public/communities/organizations/:id` or equivalent glob used in the Nuxt proxy). FAQ and resource lists for these subpages are read from the **same** organization detail payload (not separate list URLs), analogous to events and `mock-public-event-detail.ts`.

**Permission branching:**

- **Admin / can-edit:** empty state visible with the variant that includes create/add affordances (per `EmptyState` `permission` prop)
- **Non-editor (viewer):** empty state visible with the read-only variant (no misleading edit CTAs)

---

#### Test cases

For **each** page (FAQ and resources):

1. **Empty list** — After mocking empty `faqEntries` or `resources` on the organization detail response, `data-testid="empty-state"` (or page-object `emptyState`) is visible and the message matches expectations for that page type
2. **Admin vs viewer** — With the same empty API response, assert the correct empty-state variant for an admin vs a member/viewer (see epic #1983 and `EmptyState` props)
3. **Not empty** — When the mocked payload includes at least one FAQ entry or resource (or default backend data is allowed), empty state is **not** shown and list/cards are visible

---

#### Spec file locations

```
frontend/test-e2e/specs/all/organizations/organization-faq/organization-faq-empty-states.spec.ts
frontend/test-e2e/specs/all/organizations/organization-resources/organization-resources-empty-states.spec.ts
```

---

#### Implementation notes

**Detail mock setup** — Add a small helper module (for example `frontend/test-e2e/utils/mock-public-organization-detail.ts`) patterned after `mock-public-event-detail.ts`: stable UUID constant for cold `page.goto`, `routeMockPublicOrganizationDetail(page, orgId, body)`, and `mockOrganizationDetailPayload(orgId, overrides)` returning minimal JSON valid for `mapOrganization` (including required fields such as `texts`, `location`, `socialLinks`, `status`, and empty `faqEntries` / `resources` by default). Fulfill `200` + `application/json` before `page.goto` so client-only fetches see the mock.

**Page objects** — Prefer `OrganizationFAQPage` (`~/test-e2e/page-objects/organization/faq/OrganizationFAQPage.ts`) and organization resources locators from `newOrganizationPage` (`~/test-e2e/page-objects/organization/OrganizationPage.ts`), reusing `emptyState` / `emptyStateMessage` patterns from event specs.

**Auth / roles** — Use `test.use({ storageState: ... })` or existing constants such as `MEMBER_AUTH_STATE_PATH` vs default admin storage, consistent with `event-faq-empty-states.spec.ts` and other entity permission specs.

**Cleanup** — `await page.unrouteAll()` in `afterEach` (or scoped route teardown) to avoid leaking mocks.

**Tags** — Use `@desktop` and `@mobile` to match sibling entity specs.

### Contribution

Part of #1983 — E2E error and empty state coverage. Should follow the implementation and assertions established for events in #2073. Can be worked in parallel with organization group empty-state and form-validation subtasks. Parent audit context: #1959.

Aemili09:
### Contributor checklist

---

### Description

This PR adds end-to-end tests for the Organization FAQ and Resource pages to cover empty state scenarios. 

The goal is to ensure that when an organization has no FAQ entries or resources, the UI behaves as expected specifically, that the "Add" buttons appear for Admins/Editors and are hidden for standard Viewers to prevent misleading interactions.

**Key changes:**
- **`mock-public-organization-detail.ts`**: Created a utility to mock organization detail payloads. This allows testing these components without needing to hit the database for every test run.
- **`organization-faq-empty-states.spec.ts`**: New spec to verify FAQ empty state visibility and permission gating.
- **`organization-resources-empty-states.spec.ts`**: New spec to verify Resource empty state visibility and permission gating.

**Testing:**
I have verified these changes by running the tests locally using the Playwright UI. I also ran `yarn lint` and `yarn typecheck` to ensure the new files conform to the project's standards.

### Related issue

- closes #2143


aasimsyed:
Thanks for taking this on, @Aemili09! The file layout and naming line up with #2143, the admin vs viewer split is the right idea, and the mock-helper approach is exactly what the issue asked for. A few things need addressing before this can go in, and CI is currently red on four checks (E2E Desktop, E2E Mobile, Frontend Static Check, and the license header check). I've grouped feedback:

For reference, the closest existing templates to mirror are:

- `frontend/test-e2e/utils/mock-public-group-detail.ts`
- `frontend/test-e2e/specs/all/events/event-faq/event-faq-empty-states.spec.ts`

## issue (blocking): the mock URL never matches the app's request

`mock-public-organization-detail.ts` intercepts:

```ts
await page.route(`**/api/v1/organizations/${orgId}/`, ...)
```

but the app fetches the org through `getOrganization`, which calls `/communities/organizations/:id` with `withoutAuth: true` (`app/services/communities/organization/organization.ts` L30-34). With `withoutAuth`, the browser request goes through the public proxy as:

```
/api/public/communities/organizations/<id>
```

Since the intercept address does not match (wrong base, and missing the `communities/` segment), the mock never fulfills. The app then makes the real request, and because `MOCK_ORG_ID` is a synthetic UUID with no backend row, that request 404s and the empty state never renders. That is what fails both E2E jobs. Please mirror the group helper (GET-guarded, id-validated):

```ts
await page.route("**/api/public/communities/organizations/**", async (route) => {
  if (route.request().method() !== "GET") return route.continue();
  if (!isPublicOrganizationDetailGet(route.request().url())) return route.continue();
  const id = orgIdFromDetailUrl(route.request().url());
  if (!id || id !== want) return route.continue();
  await route.fulfill({ status: 200, contentType: "application/json", body: JSON.stringify(body) });
});
```

## issue (blocking): missing SPDX license header

All three files need this as the first line (this is the `check_license_header` failure):

```ts
// SPDX-License-Identifier: AGPL-3.0-or-later
```

## issue (blocking): the "not empty" case from #2143 is missing

The issue lists three cases per page: empty (admin), empty (viewer), and not empty (list shown, empty state hidden). Only the two empty variants are here. Please add the negative case by passing a sample entry, the way the event spec does with `sampleFaqEntryForMock` / `sampleResourceForMock`, and assert the list is visible and `emptyState` is not.

## issue (blocking): mock payload shape does not match `mapOrganization`

`mapOrganization` reads `texts: res.texts ?? []`, so `texts` should be an array (as in the event/group mocks: `texts: []`). The current object form:

```ts
texts: { en: { name: ..., description: ..., language: "en" } }
```

does not fit, and the payload also omits fields the mapper and org pages read (`orgName`, `name`, `tagline`, `iconUrl`, `creationDate`, `images`, `groups`, `events`). Please align with the group helper's minimal-but-complete payload so nothing throws while rendering.

## issue (blocking): import from the shared fixtures, not raw Playwright

The suite imports from the project fixtures rather than `@playwright/test`:

```ts
import { expect, test } from "~/test-e2e/global-fixtures";
```

This keeps traceability and shared setup consistent across specs.

## issue (blocking): use `~` path aliases

Instead of `../../../../page-objects/...`, please use the aliases the rest of the suite uses:

```ts
import { newOrganizationFAQPage } from "~/test-e2e/page-objects/organization/faq/OrganizationFAQPage";
import { MOCK_ORG_ID, ... } from "~/test-e2e/utils/mock-public-organization-detail";
```

## issue (blocking): tags via the `tag` option, and deeper assertions

- Move tags out of the describe title into the options: `test.describe("...", { tag: ["@desktop", "@mobile"] }, () => {...})`, and consider `@member` on the viewer describe (matches the event spec).
- The current checks only assert a generic `getByRole("button")` visible/hidden. Since the whole point is the admin vs viewer variant, please assert the actual copy and control, e.g. `getEnglishText("i18n.components.empty_state.message_with_permission")` vs `..._no_permission`, the real add button (`faqPage.newFAQButton`), and that the list is not visible. A bare button role can match unrelated chrome and does not prove the correct variant.

## nitpick

- Add a trailing newline at end of file, and remove trailing whitespace on the blank lines inside tests (part of the static-check failure).
- Consider `logTestPath(testInfo)` for traceability, as the reference specs do.
- In the helper, exporting `isPublicOrganizationDetailGet` / `organizationIdFromDetailUrl` (like the group/event helpers) makes matching precise and reusable, and exporting `sampleFaqEntryForMock` / `sampleResourceForMock` lets the "not empty" case reuse them.
- Constant naming: the sibling helpers use the full entity word (`MOCK_EVENT_EMPTY_STATE_ID`, `MOCK_GROUP_EMPTY_STATE_ID`) and every function here already uses "Organization" in full (`isPublicOrganizationDetailGet`, `organizationIdFromDetailUrl`, `mockOrganizationDetailPayload`). Only the constant was abbreviated, so please rename `MOCK_ORG_ID` -> `MOCK_ORGANIZATION_EMPTY_STATE_ID` for consistency. (The one `ORG` abbreviation in the group helper, `MOCK_GROUP_EMPTY_STATE_ORG_ID`, is a different case: a group's parent-org id.)
- `status: 2` in the payload is a magic number. There's no `OrganizationStatus` enum in the frontend (`status` is typed `number`) and it matches the real API shape, so it's fine to leave, but a short comment noting what `2` means would help the next reader.

## Summary

Great start, and the direction is correct. The main thing is the mock URL: once it points at `/api/public/communities/organizations/...`, uses an array-shaped payload, includes the "not empty" case, and adds the license header, this should go green. `mock-public-group-detail.ts` and `event-faq-empty-states.spec.ts` are the best patterns to follow. Happy to re-review once updated.

aasimsyed (changes requested):
See above comment

Aemili09:
Thanks for the detailed review! I've pushed an update to address all your feedback:

* **Mock URL**: Fixed the intercept to match the public proxy (`/api/public/communities/organizations/`).
* **Payload Shape**: Updated the mock payload to match `mapOrganization` (using arrays like `texts: []`).
* **Test Cases**: Added the "not empty" test variants to both spec files.
* **Assertions & Setup**: Switched to `getEnglishText` for stricter assertions, moved tags to the test options, and fixed path aliases (`~/`).
* **Compliance**: Added the missing SPDX license headers and resolved all static analysis/Prettier formatting errors.

Let me know if anything else is needed for this to be merged!

aasimsyed (changes requested):
Thanks for the quick turnaround, @Aemili09! Big progress here: the mock URL now points at `/api/public/communities/organizations/**` with a GET guard and id validation, the payload is array-shaped (`texts: []`, `faqEntries`, `resources`), the "not empty" case is in, tags moved to the `tag` option, and the constant is renamed to `MOCK_ORGANIZATION_EMPTY_STATE_ID`. That's the hard part done.

CI is still red on the same four checks (E2E Desktop, E2E Mobile, Frontend Static Check, license header), but they all come down to a handful of concrete things in `5cc4798`. Grouped in Conventional Comments style below.

## issue (blocking): the two `getEnglishText` / `logTestPath` imports point at modules that don't exist

This is what actually fails both E2E jobs, and it fails them at import time (the whole spec file can't load), before any test runs:

```
Error: Cannot find module '~/test-e2e/utils/i18n'
```

Both specs have:

```ts
import { getEnglishText } from "~/test-e2e/utils/i18n";          // no such file
import { logTestPath } from "~/test-e2e/utils/log-test-path";    // no such file
```

The real locations (see `event-faq-empty-states.spec.ts`) are:

```ts
import { getEnglishText } from "#shared/utils/i18n";
import { logTestPath } from "~/test-e2e/utils/test-traceability";
```

Fixing these two lines in both spec files should turn E2E Desktop and E2E Mobile green.

## issue (blocking): the page object is called with an argument it doesn't take, and `.goto()` isn't a method on it

`newOrganizationFAQPage` / `newOrganizationResourcesPage` take a single `page` argument and return an object of locators only, no `goto`. The current usage:

```ts
const faqPage = newOrganizationFAQPage(page, MOCK_ORGANIZATION_EMPTY_STATE_ID); // 2 args
await faqPage.goto();                                                            // no such method
```

produces the typecheck failures:

```
error TS2554: Expected 1 arguments, but got 2.
error TS2339: Property 'goto' does not exist on type '{ newFAQButton: Locator; ... }'.
```

Navigate with `page.goto(...)` directly, the way the event spec does:

```ts
const faqPage = newOrganizationFAQPage(page);
await page.goto(`/organizations/${MOCK_ORGANIZATION_EMPTY_STATE_ID}/faq`);
```

(and `.../resources` for the resources spec). Keep routing the mock before the `goto`, as you already do.

## issue (blocking): SPDX license header is still missing

The `check_license_header` job still fails (`Files with license 'AGPL-3.0-or-later': 0 / 1`). None of the three files have the header yet, so this needs to be the very first line of each:

```ts
// SPDX-License-Identifier: AGPL-3.0-or-later
```

## issue (blocking): `organizationIdFromDetailUrl` return type doesn't hold under strict indexing

The mock helper trips the typecheck too:

```
mock-public-organization-detail.ts(14,18): error TS2322: Type 'string | undefined' is not assignable to type 'string | null'.
```

`match[1]` is `string | undefined` under `noUncheckedIndexedAccess`, so:

```ts
export function organizationIdFromDetailUrl(url: string): string | null {
  const match = url.match(/\/organizations\/([^/?]+)/);
  return match?.[1] ?? null;
}
```

## nitpick (non-blocking)

- `MOCK_ORGANIZATION_EMPTY_STATE_ID` is `00000000-0000-0000-0000-000000000001`, which isn't a valid v4 UUID (the version/variant nibbles are `0`). The sibling helpers use valid v4 ids (e.g. `00000000-0000-4000-8000-...`); worth matching so nothing downstream that validates UUID shape trips.
- In the "not empty" test you assert both empty-state messages are hidden, which is good; consider also asserting the list is visible (`page.getByRole("list")` / the page object's `faqList`) so the test proves the list rendered, not just that the empty copy is gone.

## Summary

Really close now. The E2E red is just the two bad import paths, and the Static Check red is the page-object call + `.goto()` + the helper's return type; the license header is a one-line-per-file add. `event-faq-empty-states.spec.ts` has the exact import paths and navigation pattern to copy from. Happy to re-review as soon as it's pushed.

Aemili09:
Thanks for catching those! I've pushed the final fixes:

* **Imports**: Fixed the missing module paths (`#shared/utils/i18n` and `test-traceability`).
* **Navigation**: Switched to direct `page.goto(...)` for routing.
* **TypeScript & UUID**: Updated the UUID to a valid v4 format and added the optional chaining (`match?.[1] ?? null`) to pass strict indexing.
* **Headers**: Verified the SPDX header is strictly on line 1 for all files.

Local type checks and linters are completely green. Ready for another look!

aasimsyed (changes requested):
Thanks for the update, @Aemili09! The import paths and page-object usage are sorted now, nice. One thing on the latest commit though.

**issue (blocking): let's not disable an ESLint rule to work around this, let's fix the cause.**

The `beforeEach` only needs `testInfo`, so it forces an empty `{}` for the fixtures arg, which trips `no-empty-pattern`, and the commit silences it:

```ts
// eslint-disable-next-line no-empty-pattern
test.beforeEach(async ({}, testInfo) => {
  logTestPath(testInfo);
});
```

`no-empty-pattern` is firing for a good reason (an empty destructure is almost always a mistake), and disabling it hides the smell rather than removing it. As a rule of thumb, we should avoid `eslint-disable` and fix the underlying issue instead; a suppressed rule tends to get copied around and quietly erodes the linting.

The clean fix here is also what our other empty-state specs already do: drop the `beforeEach` and call `logTestPath(testInfo)` at the top of each test, where the callback already takes the fixtures it needs plus `testInfo`. No empty pattern, nothing to disable. See `event-faq-empty-states.spec.ts`:

```ts
test("Admin sees the empty state ...", async ({ page }, testInfo) => {
  logTestPath(testInfo);
  const faqPage = newOrganizationFAQPage(page);
  // ...
});
```

For the viewer test it's `async ({ browser }, testInfo)`, same idea. This also keeps these specs consistent with the event ones.

**nitpick:** the multi-line `expect(...)` calls got collapsed into some very long single lines. If those exceed the Prettier print width they'll get re-wrapped, so please make sure `prettier --check` (part of the Frontend Static Check) is actually green, not just ESLint.

Happy to re-review once the disable is gone. Thanks again!

aasimsyed (changes requested):
Thanks for pushing the ESLint fix, @Aemili09! Static Check, Frontend Tests, license header, and i18n are all green now.

E2E is still red, but not for the reasons we've already covered. Three pre-existing flakes on `main` are unrelated to this PR (`organization-about-social-links.spec.ts`, `organization-group-about-social-links.spec.ts`, and the `events-pagination.spec.ts` filter-change test, all show up as "flaky" in the run and are being tracked separately). The real issue is your new specs themselves: all 6 tests in `organization-faq-empty-states.spec.ts` and all 6 in `organization-resources-empty-states.spec.ts` fail on every run and every retry, both desktop and mobile.

## issue (blocking): `page.getByRole("list")` is not unique, so the visibility assertions never resolve

```
Error: expect(locator).toBeHidden() failed
Locator: getByRole('list')
Error: strict mode violation: getByRole('list') resolved to 4 elements:
    1) <ol> ... aka locator('ol')                          // toast notifications
    2) <ul id="submenu"> ...                                // sidebar submenu
    3) <ul> ... aka getByText('< Organizations | ...')      // breadcrumb
    4) <ul class="list-disc..."> ... getByTestId('empty-state').getByRole('list')  // the actual empty-state bullet list
```

`getByRole("list")` matches every `<ul>`/`<ol>` on the page, toasts, the sidebar submenu, the breadcrumb, and the actual list, so Playwright can't resolve it uniquely and the assertion fails regardless of what you're trying to check.

Both page objects already export a scoped locator for exactly this:

```ts
// OrganizationFAQPage.ts
faqList: page.getByTestId("organization-faq-list"),

// OrganizationResourcesPage.ts
resourcesList: page.getByTestId("organization-resources-list"),
```

Swap the three `page.getByRole("list")` calls in each spec for `faqPage.faqList` / `resourcesPage.resourcesList`:

```ts
await expect(faqPage.faqList).toBeHidden();   // empty-state tests
await expect(faqPage.faqList).toBeVisible();  // "not empty" test
```

and the same for `resourcesPage.resourcesList` in the resources spec. That should get both E2E jobs to just the pre-existing flakes.

aasimsyed:
The scoped-locator fix worked, `organization-faq-empty-states.spec.ts` and `organization-resources-empty-states.spec.ts` are both fully green now on Desktop and Mobile.

E2E is still showing red, but everything left is pre-existing and unrelated to this PR's files:

- `organization-about-social-links.spec.ts` and `organization-group-about-social-links.spec.ts` - known stale-data-after-save bug, tracked in [#2250](https://github.com/activist-org/activist/issues/2250).
- `organization-faq-page.spec.ts` - same root cause as above, just on the FAQ page.
- `events-pagination.spec.ts` (filter-change test) - tracked in [#2204](https://github.com/activist-org/activist/issues/2204).
- `organization-group-resources-page.spec.ts` (mobile drag-and-drop reorder) - new to me, but this PR doesn't touch group resources or drag-and-drop at all, so it doesn't look related either.

None of these touch the mock helper or the two new specs this PR added. 

cc: @andrewtavis @nicki182 I think this should be mergeable now and the failing tests shouldn't block.

Aemili09:
Thanks for confirming, @aasimsyed!
@andrewtavis @nicki182 , Just giving this a friendly bump! It looks like there is still a stale 'Changes requested' block on the PR from an earlier review. Could one of you take a look and clear it when you have a moment?
I'm ready for my next issue once this merges. I enjoyed the E2E testing, but I'm equally happy to tackle the Django backend next. Let me know where I can be most helpful!

andrewtavis:
Thanks for the bump here, @Aemili09! I've been quite busy over the last few days, but @nicki182 and I should be able to look into this on Monday :)

Aemili09:
Hi @andrewtavis @nicki182! Just checking in since it's been about a month.

I noticed I got unassigned from the issue (totally understand if it was automated or just routine repo cleanup!). I just wanted to flag that all the requested fixes are in, and the E2E tests are green on my end (aside from the known pre-existing flakes).

Let me know if there's anything else you need from me to clear that stale block and get this merged. Still very eager to grab my next Django or E2E ticket once this is wrapped up!

andrewtavis:
Thank you so much for the ping here, @Aemili09! Is quite helpful :) Would be good if we could rebase this and get it to a point where e2e tests are passing as the issues we were having with them are fixed now 😊

Aemili09:
Awesome, that makes perfect sense! I've just rebased the branch against the latest main to pull in your test fixes and force-pushed the update.
Let me know if everything looks good on your end.

andrewtavis:
Looks like we made progress on this issue. @aasimsyed, you'd requested changes. Could you let us know how things are looking here now?

aasimsyed (approved):
Confirmed, everything's addressed. Went through all four rounds against the current diff (`24372bb`), not just the CI badges:

- mock URL, payload shape, and the "not empty" case from round 1
- import paths and page-object usage from round 2
- the `no-empty-pattern` fix from round 3
- the scoped `faqPage.faqList` / `resourcesPage.resourcesList` locators from round 4, which were causing all 12 new tests to fail

CI is fully green now, including both E2E jobs. 

Approving. Thanks for sticking with this through the back and forth, @Aemili09, nice work.

andrewtavis:
Quick note that b5b72b8 fixed the formatting of the links at the bottom of the page. Letting tests run and then we can bring this in :)

andrewtavis (approved):
note: Failing e2e test is flaky, so all good to bring this in :)

praise: Thanks for the PR and the reaction to the review, @Aemili09! Thanks also for the review, @aasimsyed 😊

PR #2251, opened 2026-07-12
Clear cached entity data before refresh in mutation composables (#2250)

### LINKED ISSUE #2250: Bug: Saved changes sometimes don't appear on the page until a reload
### Terms

### Behavior

After saving a change on an organization, group, or event page (for example creating, editing, or deleting an FAQ entry, resource, or social link, or changing an event's details), the page sometimes keeps showing the old content. The change did save on the server, because reloading the page shows it correctly, but the page in front of you never updates on its own.

I think this is happening because of how the app refreshes data after a save. After every save the app asks the server for a fresh copy of the page's data. But if the app is already in the middle of fetching that same data at that moment, the new request is silently skipped and the page keeps its old copy. This is most likely when two saves happen close together, since the first save's refresh is often still running when the second save asks for one.

**Steps to reproduce:**

1. Sign in as `admin` / `admin` on the dev stack.
2. Open any organization and go to its **FAQ** tab.
3. Create two FAQ entries, one after the other.
4. Delete the first entry, and as soon as the confirmation modal closes, immediately delete the second entry.
5. Watch the list.

**Actual:** the second entry you deleted often stays visible in the list. It only disappears after you reload the page.
**Expected:** both entries disappear from the list right away.

Note: this is a timing issue, so on a fast machine it may take a few tries. To make it reproduce every time, open DevTools, go to the Network tab, and add a custom throttling profile with around 2000 ms of latency. With that enabled, doing any two saves in quick succession shows the problem reliably.

**Flaky/failing e2e tests this causes:**

These tests do exactly this kind of quick create/update/delete sequence, so they hit the stale data and fail or pass only on retry:

- `event-faq-page.spec.ts` - "User can manage FAQ entries (CREATE, UPDATE, DELETE)"
- `organization-faq-page.spec.ts` - "User can manage FAQ entries (CREATE, UPDATE, DELETE)"
- `organization-about-social-links.spec.ts` - "User can manage social links (CREATE, UPDATE, DELETE)"
- `organization-group-about-social-links.spec.ts` - "User can manage social links (CREATE, UPDATE, DELETE)"
- `event-about-details-permissions.spec.ts` - "Member can change the organization and see it on the details card"

aasimsyed:
### Contributor checklist

---

### Description

Fixes #2250: after a save, the page could keep showing old content until a manual reload.

The mutation composables refresh entity data with `await refreshNuxtData(key)`, but the matching query composables (`useGetEvent`, `useGetOrganization`, `useGetGroup`) use `dedupe: "defer"`. If a fetch for the same key is already in flight when the refresh fires, the refresh is silently dropped and the pre-mutation data stays cached. This PR calls `clearNuxtData(key)` before `refreshNuxtData(key)` in each refresh helper so a real refetch always happens.

Same two-line change in all 14 entity-level mutation composables (event, organization, and group variants for FAQ entries, resources, social links, texts, and images).

**Testing performed**

- All 230 mutation composable unit tests pass
- The previously flaky e2e specs listed in #2250 (event/org FAQ, org and group social links) pass repeatedly with `--repeat-each` against a local preview build
- eslint, prettier, and typecheck are green

### Related issue

- Closes #2250

### DIFF
--- a/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
+++ b/frontend/app/composables/mutations/useEventFAQEntryMutations.ts
@@ -91,8 +91,10 @@ export function useEventFAQEntryMutations(eventId: MaybeRef<string>) {
   async function invalidateCacheRefreshEventData() {
     if (!currentEventId.value) return;
 
-    // Invalidate the useAsyncData cache so next read will refetch.
-    await refreshNuxtData(getKeyForGetEvent(currentEventId.value));
+    const key = getKeyForGetEvent(currentEventId.value);
+
+    clearNuxtData(key);
+    await refreshNuxtData(key);
   }
 
   return {
--- a/frontend/app/composables/mutations/useEventImageIconMutations.ts
+++ b/frontend/app/composables/mutations/useEventImageIconMutations.ts
@@ -32,7 +32,10 @@ export function useEventImageIconMutations(eventId: MaybeRef<string>) {
   async function invalidateCacheRefreshEventData() {
     if (!currentEventId.value) return;
 
-    await refreshNuxtData(getKeyForGetEvent(currentEventId.value));
+    const key = getKeyForGetEvent(currentEventId.value);
+
+    clearNuxtData(key);
+    await refreshNuxtData(key);
     // Clear cached events to force refetch with new data.
     store.setItems([]);
   }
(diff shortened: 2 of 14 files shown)


nicki182:
@aasimsyed there is an issue that if you use clearNuxtData for a split second you delete the cache data which in the background of the modal of FAQ for example when you edit it will show that there isnt any faqs. Maybe with the refactor with pinia colada could this issue be solved?

Issue #2272, opened 2026-07-24
Refactor: extend @pinia/colada migration to organization and group mutations


aasimsyed:
### Description

PR #2122 migrated event data (create, FAQ, resources, social links, texts, images) to `@pinia/colada`, fixing the stale-data bugs tracked in #2250. This issue tracks doing the same for organizations and groups, following that same pattern. It also covers the original scope of #2111 (org/group image CRUD), which #2122 didn't get to since it used events as the proof of concept first.

### In scope

- `useOrganizationImageMutations.ts`, `useGroupImageMutations.ts`, `useGetOrganizationImages.ts`, `useGetGroupImages.ts`, replaced to match `useEventImageIconMutations.ts`
- `useFileManager.ts`, slimmed down to local file/UI state only
- Org and group equivalents of FAQ, resources, social links, and texts mutations
- Keep `create()` resolving to `false` on failure (not throwing) when migrating `useOrganizationMutations.ts` and `useGroupMutations.ts`

### Related

- #2111 
- #2250 
- #2122

andrewtavis:
Is there anything else left in this issue, @aasimsyed? If all that's left is covered in https://github.com/activist-org/activist/issues/2250, then we should be good to close this :)

aasimsyed:
Good to close.

andrewtavis:
Thanks so much, @aasimsyed!

PR #1730, opened 2025-11-18
Add pagination to organizations/ page

kirthiry:
### Contributor checklist

[X] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
[] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the [testing section of the contributing guide](CONTRIBUTING.md#testing)

---

### Description

* Collaborators: @kirthiry and @jtauro-umich 
* This PR attempts to add pagination to the organizations/ page in response to #1677.
* We are requesting feedback/code review since we have been unable to successfully implement pagination. Our current implementation changes do not break anything, but they also do not make pagination work.
* We haven't resolved merge conflicts yet since we have not finished this implementation!

### Future tasks

* Add unit tests and confirm that all unit tests pass.
* Re-submit or submit a new PR that is functional.

### Related issue

#1677

PR #2281, opened 2026-07-26
Refactor: extend @pinia/colada migration to organization and group mutations

aasimsyed:
### Description

Extends the @pinia/colada migration from #2122 (events) to organizations and
groups, following the same pattern: a cache composable per entity for
invalidation, `useGetX` migrated to `useQuery`, and mutation composables
migrated to `useMutation`.

**This PR will not pass CI until the query
migration commit lands on top of this one** (the cache composables reference
key factories defined there); that's expected while this stays in draft.

### Related issue

- Part of #2272

### Contributor checklist

- [x] This pull request is on a separate branch and not the main branch
- [ ] I have run the tests for the backend and frontend depending on what's needed for my changes as described in the testing section of the contributing guide


nicki182:
@aasimsyed I have created a composable in cache folder within composables that has everything related to cache management, useEventCache is named, maybe we can expand this and no longer use @pinia as a cache in between pages and remove the data folder within the store. Just a thought I had. Secondly I put there the keys there in #1730 PR.

aasimsyed:
@nicki182 good thought, and I agree the `stores/data/` folder is redundant with what Colada's cache already does.

Not something to fold into this PR though: `PageBreadcrumbs.vue` reads `useOrganizationStore`/`useGroupStore`/`useEventStore` directly for the entity name, and `useGroupMutations.ts` still reads `store.getOrganization()`. Removing the stores means moving those reads over too, across all three entities, which is bigger than #2272's scope and would touch events again after #2122 already merged.

Woudl be great as its own issue once orgs/groups here are done. Happy to file it if you don't get to it first.

andrewtavis:
praise: Thanks for the persistence here, @aasimsyed! So amazing that the e2e tests are being fixed as well :)
question: Are we good to go here now? Feel free to mark this as ready for review if it is :)

CC @nicki182 👋

aasimsyed:
Thanks @andrewtavis!

Yes, this is ready for review. The org/group Colada migration that mirrors #2122 is in, and the e2e failures that motivated #2272 (FAQ / social links / about content / QR share) are fixed. Static checks and unit tests are green; desktop e2e has been green too. Mobile still occasionally trips `failOnFlakyTests` on unrelated pre-existing flakes (navigation timing / `ECONNRESET`), not on the specs this PR targeted.

CC @nicki182

aasimsyed:
@andrewtavis not quite done with the full #2272 scope yet, though. This PR covers the e2e-critical slice (org/group `useQuery`/`useMutation` for FAQ and social links, Colada invalidation elsewhere, plus the `entityId`/modal races). Still left for follow-ups:

- finish the half-migrated mutations (group FAQ, resources, texts) as full `useMutation`
- migrate org/group image queries and mutations (the #2111 piece)
- slim `useFileManager` down to local file/UI state
- list queries / entity `create()` still on `useAsyncData`/`refreshNuxtData`

Happy to keep working on those before merge.

andrewtavis:
question, @aasimsyed: Do you have anything else that you'd like to add to this? If not, @nicki182 and I would look into it :)

aasimsyed:
@andrewtavis: this PR is green on the last two runs, both e2e suites included.

What's left is the organization and group list reads, which are still on `useAsyncData` with `refreshNuxtData` on create. Those paginate through infinite scroll, so they want `useInfiniteQuery` rather than a straight `useQuery` swap, and there's no existing example in the codebase to mirror yet. I'd rather take those as a follow-up than grow this PR further.

andrewtavis:
Sounds good, @aasimsyed! Do you want to make an issue or two for the rest of the work and then we can bring this in when we can? You can block the issues for now maybe until this PR is merged?

nicki182:
@andrewtavis I have done a PR for the paginated queries but I will change to use useInfiniteQuery from @pinia/colada

andrewtavis:
We're halfway through here now, @aasimsyed, and will pick up the review later on in the week :)

aasimsyed:
I think 781fe60 could be causing the new mobile E2E failure here.

Looking at the CI history on this branch, `E2E Mobile Chrome` was green on 217ac05 and started failing on 781fe60, with nothing else changed in between. The failure is on `organization-about-social-links.spec.ts`, consistently on all 3 attempts, timing out waiting for the "Add" button in the social links modal to appear.

That's the same flow 60d8212 addressed: `entity` comes from an async Colada query, so `entity?.id ?? ''` can still be `''` when the edit icon is clicked, and that empty id gets passed into `ModalSocialLinksOrganization`. Mobile runners seem to hit that window more often than desktop, which might explain why only mobile is red.

What do you think?

nicki182:
@aasimsyed I think so too. I can make some changes to make it work and maintain the props in logical way

andrewtavis:
Thanks for pointing it out @aasimsyed! Will be great when we're back to all green 😊

nicki182:
@aasimsyed can you fix the test? I prefer to not touch the code as it is. Entity gives enough information as a prop and doing anything else like adding a new prop feels unnecessary

aasimsyed:
Pushed 2b44f591, waits for the organization heading to have real text before the test clicks the connect card's edit icon, tests green (ran it 3x locally).

One caveat: this only fixes the test's timing, not the underlying window. A real user could still click that icon in the same narrow gap, on a slow connection, and the app would send an empty id into the social links modal, a silent failure rather than a flake. So CI will be green, but the click-before-load gap itself is still there.

What if we also gated the edit icon on `entity` being loaded, instead of just being signed in? Something like:

```vue
<IconEdit
  v-if="userIsSignedIn && entity"
  @click="() => entity?.id && openModalSocialLinks({ entityId: entity.id })"
  ...
/>
```

Still just `entity`, no new prop, it just waits for that data before acting on it rather than assuming it's there. I tried this too and it also passes locally. Only tradeoff is the icon shows up a beat later on page load instead of instantly, since it's now waiting on the fetch.

Not pushing that one since you'd mentioned preferring not to touch the code, just flagging it here in case it changes anything for you.

nicki182:
> Pushed [2b44f59](https://github.com/activist-org/activist/commit/2b44f591ef8eb5c3c89a806e9d093586d428045e), waits for the organization heading to have real text before the test clicks the connect card's edit icon, tests green (ran it 3x locally).
> 
> One caveat: this only fixes the test's timing, not the underlying window. A real user could still click that icon in the same narrow gap, on a slow connection, and the app would send an empty id into the social links modal, a silent failure rather than a flake. So CI will be green, but the click-before-load gap itself is still there.
> 
> What if we also gated the edit icon on `entity` being loaded, instead of just being signed in? Something like:
> 
> ```
> <IconEdit
>   v-if="userIsSignedIn && entity"
>   @click="() => entity?.id && openModalSocialLinks({ entityId: entity.id })"
>   ...
> />
> ```
> 
> Still just `entity`, no new prop, it just waits for that data before acting on it rather than assuming it's there. I tried this too and it also passes locally. Only tradeoff is the icon shows up a beat later on page load instead of instantly, since it's now waiting on the fetch.
> 
> Not pushing that one since you'd mentioned preferring not to touch the code, just flagging it here in case it changes anything for you.

Within IconEdit we can do this and it might be the safest way so that in the future we don't have the same issue and its cleaner
 ```
<!-- SPDX-License-Identifier: AGPL-3.0-or-later -->
<template>
  <button
    v-if="entity && canEditEntity"
    aria-labelledby="icon-edit-label"
    class="cursor-pointer text-primary-text focus-brand hover:text-distinct-text"
    data-testid="icon-edit"
  >
    <Icon :name="IconMap.EDIT" size="1.2em" />
    <span id="icon-edit-label" hidden>{{
      $t("i18n.components.icon_edit.aria_label_alt_text")
    }}</span>
  </button>
</template>

<script setup lang="ts">
const props = defineProps<{
  entity?: Entity | null;
}>();

const { canEdit } = useUser();
const canEditEntity = computed(() => canEdit(props?.entity));
</script>
 ```
what do you think @aasimsyed ? I added a v-if within the button

aasimsyed:
Agreed, this is cleaner than my CardConnect version, fixes it once for every IconEdit consumer instead of just the connect card.

One thing: `IconDraggableEdit.vue` and `EntityIconMobile.vue` use the same `canEditEntity = computed(() => canEdit(props.entity))` pattern. Should update those, too, I think.

nicki182:
@aasimsyed sounds perfect to me to do that

andrewtavis (approved):
note: Checked this with @nicki182 earlier and a usability check just now show's that all's working :) I didn't check everything everything, but if there's an issue then we can move the code over from a spot where it's working.

praise: Thanks so much for getting this through, @aasimsyed! Also thanks for the review support and discussion, @nicki182 😊 So great that this is in and we have passing e2e tests again! ✅

PR #2302, opened 2026-07-31
fix: implement calendar .ics download for single events

lb1192176991-lab:
## What

This PR wires up the `downloadCalendarEntry` function in both the event tooltip menu (`TooltipMenuSearchResultEvent.vue`) and the event about page (`about.vue`) to call the backend `/v1/events/event_calendar` endpoint and trigger a browser file download.

Previously both functions were empty stubs, so clicking the "Download the calendar entry" button had no effect.

## Why

The backend already exposes `GET /v1/events/event_calendar?event_id=<UUID>` which returns an iCalendar file with a suggested filename via `Content-Disposition`. The frontend buttons were wired to an empty function, so users could not actually download event calendar entries.

## Testing

- Verified the download function fetches the .ics file from the API
- Extracts the filename from Content-Disposition header (falls back to `event-<id>.ics`)
- Creates a blob URL and triggers browser download via a temporary anchor element
- Cleans up the object URL after download
- Errors are caught and logged to console without disrupting the UI

### DIFF
--- a/frontend/app/components/tooltip/menu-search-result/TooltipMenuSearchResultEvent.vue
+++ b/frontend/app/components/tooltip/menu-search-result/TooltipMenuSearchResultEvent.vue
@@ -46,14 +46,36 @@
 </template>
 
 <script setup lang="ts">
-defineProps<{
+const props = defineProps<{
   event: CommunityEvent;
 }>();
 
 const emit = defineEmits(["tab"]);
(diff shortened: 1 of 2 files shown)


andrewtavis:
Closing this in favor of #2317 :) Thanks for your interest in the project, @lb1192176991-lab :)

PR #2312, opened 2026-08-02
fix: remove invalid --silent=passed-only flag from Vitest CI command

### LINKED ISSUE #2309: Dependency update check isn't working for Vitest
### Terms

### Behavior

In #2316 and similar issues where we've been testing [.github/workflows/dependency_update_check.yaml](https://github.com/activist-org/activist/blob/main/.github/workflows/dependency_update_check.yaml) we can see that the Vitest test suite isn't being ran properly. We can run this suite within `frontend/` via `yarn test`, which is the basis of the command that we're running in the snippet for Vitest in the update check:

```yaml
echo -e "\n### Vitest Results" >> ../reports.md
echo "\`\`\`" >> ../reports.md
yarn test --reporter=dot --silent='passed-only' --no-color 2>&1 | tee -a ../reports.md || true
echo "\`\`\`" >> ../reports.md
```

These tests had worked in the past, and now what we're getting is that all tests are failed or all tests are skipped. All the arguments in `yarn test --reporter=dot --silent='passed-only' --no-color` are valid now. The tests just aren't being ran properly.

We'd be happy to get some support on this if people have ideas :)

waterWang:
## Summary

Fixes #2309

The Vitest CI command in dependency_update_check.yaml uses `--silent=passed-only` which is **not a valid Vitest CLI option**. In Vitest, `--silent` is a boolean flag that doesn't accept a value. Passing `--silent=passed-only` causes the test runner to either fail or silently skip all tests.

## Change

Removed the invalid `--silent=passed-only` flag from the Vitest test command. The `--reporter=dot` flag is kept to maintain concise output (dots for passed tests, failure details at the end).

## Before

```yaml
yarn test --reporter=dot --silent=passed-only 2>&1 | tee -a ../reports.md || true
```

## After

```yaml
yarn test --reporter=dot 2>&1 | tee -a ../reports.md || true
```

## Testing

The fix is in the CI workflow file only — no functional code changes to the frontend application.

### DIFF
--- a/.github/workflows/dependency_update_check.yaml
+++ b/.github/workflows/dependency_update_check.yaml
@@ -179,7 +179,7 @@ jobs:
 
           echo -e "\n### Vitest Results" >> ../reports.md
           echo "\`\`\`" >> ../reports.md
-          yarn test --reporter=dot 2>&1 | tee -a ../reports.md || true
+          yarn test --reporter=dot --silent='passed-only' 2>&1 | tee -a ../reports.md || true
           echo "\`\`\`" >> ../reports.md
 
       # MARK: Create Issue
--- a/.github/workflows/dependency_update_check.yaml
+++ b/.github/workflows/dependency_update_check.yaml
@@ -179,7 +179,7 @@ jobs:
 
           echo -e "\n### Vitest Results" >> ../reports.md
           echo "\`\`\`" >> ../reports.md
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Thanks for sending along this change, @waterWang :)

note: Do read the docs about the arguments in question. Your solution and explanation led to a potential solution, but it's incorrect that silent is only a boolean. See [the docs here](https://vitest.dev/config/silent.html).

Let's see if switching it to a string will work 😊

PR #2317, opened 2026-08-03
fix(frontend): download individual event calendars

### LINKED ISSUE #2301: Single-event calendar actions do not download the available .ics file
### Terms

### Behavior

The frontend currently shows a calendar action for individual events in two places:

- The event About page.
- The event card tooltip/menu.

Both actions currently call an empty `downloadCalendarEntry` function, so selecting them has no effect.

The existing accessible label, “Download the calendar entry for this event”, indicates that these actions are intended to download the individual event rather than create a persistent calendar subscription.

The backend already exposes a public endpoint for exporting one event:

```http
GET /v1/events/event_calendar?event_id=<UUID>
```

The endpoint returns an iCalendar file and provides a suggested filename through the `Content-Disposition` response header.

Selecting either action should request this endpoint using the selected event ID and download the returned `.ics` file. The filename should come from `Content-Disposition` when available, with a safe `.ics` fallback otherwise.

Both entry points should use a shared frontend implementation and follow the project’s existing public API proxy, error-handling, accessibility and localization patterns. Any temporary browser resources created for the download should also be cleaned up.

Since this is a static export for one event, “Add to calendar” may be more accurate than “Subscribe to event”. The final wording should be confirmed by the maintainers.

### Scope

This issue only covers the two existing calendar actions for individual events.

Organization and group calendar actions should remain unchanged. Their export and subscription behavior was discussed separately in #1448 and PR #1483.

No backend changes are expected.

### Acceptance criteria

- [ ] Both single-event actions download the `.ics` file for the correct event.
- [ ] Both entry points use the same shared implementation.
- [ ] The filename is read from `Content-Disposition` when available, with a safe fallback otherwise.
- [ ] Failed requests show accessible feedback using the project’s i18n system, without unhandled exceptions.
- [ ] Keyboard activation triggers only one download.
- [ ] Temporary browser resources are cleaned up.
- [ ] Relevant frontend tests are added.

### Related work

- #716: original calendar subscription/export feature
- #1276: added the existing frontend calendar actions
- #1307: defined the `.ics` export work
- #1313: implemented the single-event backend endpoint
- #1448 / PR #1483: discussed organization and group calendar export separately

### Contribution

I would be happy to implement this after confirmation from the maintainers.

Ulbrec87:
## Related issue

Closes #2301.

## Summary

- Connect the two existing single-event calendar actions to the public event calendar export endpoint.
- Add a shared frontend implementation used by both entry points.
- Use the filename provided through `Content-Disposition`, with a safe `.ics` fallback.
- Route download failures through the existing `useAppError` handling.
- Clean up temporary object URLs and download elements.
- Preserve the native button keyboard behavior and avoid duplicate downloads.

## Testing

- Added HTTP service tests for raw responses, including response headers.
- Added event service tests for the calendar export endpoint.
- Added composable tests for `Content-Disposition` filename handling, downloads and failures.
- Added utility tests for download behavior and resource cleanup.
- Added component tests for both entry points, verifying that keyboard activation triggers a single download.
- Verified that all 26 targeted tests, typecheck and ESLint pass.


andrewtavis:
Thanks for the PR, @Ulbrec87! We'll check this in the coming days :) 

note: We still don't have the issue for the backend test standardization. Let me know if I should make that. Happy to, if need be!

Ulbrec87:
> Thanks for the PR, @Ulbrec87! We'll check this in the coming days :)
> 
> note: We still don't have the issue for the backend test standardization. Let me know if I should make that. Happy to, if need be!

Thanks! I’d be happy to take care of opening the issue in the next few days, if that works for you.

I’d like to review the existing HTTP 401 and 403 tests first, so I can outline a clear proposal for the conventions and scope, initially focusing on the groups and organizations areas. I’ll share the issue once it’s ready.

andrewtavis:
quick check in here, @Ulbrec87 👋 Sorry for the wait here. We were fighting through the most recent Nuxt/Vite update and put other reviews on hold. This would be close to our top priority now. Would you be able to fix the merge conflicts. e2e tests should also be able to pass now, so once we bring in main and have passing tests we should be good to merge 😊

Let us know if you need support!

andrewtavis (approved):
praise: As always a really nice contribution, @Ulbrec87 😊 Really appreciate the care that you're putting into these PRs.

note: @nicki182 and I just checked this and did a few minor changes to make sure that the response can include events with multiple times as well as integrate some new changes in main.

PR #2319, opened 2026-08-03
chore(frontend): upgrade Playwright to 1.62.1

aasimsyed:
### Contributor checklist

---

### Description

Upgrades Playwright from 1.61 to 1.62.1, which brings Chromium 151. Only `frontend/package.json` and `frontend/yarn.lock` change; CI installs browsers off the pinned version, so there's no image pin or cache key to update.

It also fixes a skew we already had: `resolutions` forced `playwright-core` to `1.61.0` while `@playwright/test` was `1.61.1`, so the installed driver did not match the test package. All three pins now sit at 1.62.1.

**Why this is worth doing, mostly for flakiness.**

- `Reporter.preprocess()` is a new hook that can mark individual tests as skipped or excluded before the run starts. With `failOnFlakyTests` on, a single recurring flake currently reds the job on PRs that never touched it. This lets us keep a quarantine list in one place instead of scattering `test.fixme` through specs, and makes it obvious what's quarantined and why.
- `locator.waitForFunction()` waits until a function called with the matching element returns truthy. It's a first-class replacement for the `expect.poll` around `locator.count()` pattern we fall back on whenever a spec has to wait for DOM state to settle, which is where several of our races have come from.
- A new `scroll: "none"` option on actions opts out of automatic scroll-into-view, which is worth trying against the mobile drag-and-drop reorder flake.

**Risk.** A browser bump can shift timing on existing flakes either way. This is deliberately separate from the open @pinia/colada work so its own CI run shows what moves under the new browser with no application changes in the diff.

### Related issue

- No related issue; this is a routine dependency upgrade.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -38,7 +38,7 @@
     "@nuxtjs/i18n": "10.4.0",
     "@pinia/colada-nuxt": "1.0.0",
     "@pinia/nuxt": "0.11.3",
-    "@playwright/test": "1.61.1",
+    "@playwright/test": "1.62.1",
     "@testing-library/vue": "8.1.0",
     "@types/node": "26.1.1",
     "@types/zxcvbn": "4.4.5",
@@ -60,7 +60,7 @@
     "nuxt": "4.4.8",
     "nuxt-auth-utils": "0.5.29",
     "nuxt-security": "2.6.0",
-    "playwright-core": "1.61.1",
+    "playwright-core": "1.62.1",
     "prettier": "3.9.5",
     "prettier-plugin-tailwindcss": "0.8.0",
     "rollup": "4.62.2",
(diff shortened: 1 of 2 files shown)


aasimsyed:
Heads up on the two red E2E jobs: they look pre-existing on `main` rather than caused by the version bump.

This branch only touches `package.json` and `yarn.lock`. Comparing against #2317, an unrelated main-based PR still on Playwright 1.61, the same specs fail there:

- `event-create-modal.spec.ts:751`
- `organization-about-social-links.spec.ts:33`
- `organization-faq-page.spec.ts:50`

Each run also picks up one extra failure the other doesn't (`organization-group-about-social-links:19` here, `organization-faq-form-validation:72` on #2317), both in families that are already flaky, so that part looks like normal variance.

Those specs are the same ones #2281 fixes (stale modal assertion on failed create, social link form clobbering, and the FAQ `entityId` race), and #2281 is currently green on both E2E suites. So I'd expect this to go green after #2281 lands and I rebase. Happy to hold this until then.

andrewtavis (approved):
Thanks for the upgrade, @aasimsyed!

PR #2323, opened 2026-08-03
fix(frontend): un-skip mobile event reorder e2e tests

aasimsyed:
### Contributor checklist

---

### Description

The mobile branch of navigateToEventSubpage was never exercised in CI, since the specs that use it are tagged @desktop, and it carried two races:

- navigateToFirstEvent clicked the list view radio unconditionally. Both guards swallowed their errors, so when the switcher was absent the click burned a 10s timeout. The switcher lives in the left sidebar filter, which is not rendered on mobile, and the events page already defaults to list view.
- The submenu re-renders while event data resolves, which could swallow the click that opens it or close it again before the option was selected. Opening and selecting are now retried as a single unit.

The event FAQ page passed entityId: event?.id, but event is a ref, so the value was always undefined and the create modal could fail to close.

Verified on Playwright 1.62.1 against a local preview build: mobile resources 10 of 10, mobile FAQ 12 of 12, desktop event suite 108 of 108.

### Related issue

- N/A


aasimsyed:
most of the failures are pre-existing on `main` and are already fixed in #2281.

- `event-create-modal.spec.ts:751` — assertion expects the modal to close after a 500; a failed create correctly keeps it open. #2281 updates that assertion.
- `organization-about-social-links` / `organization-group-about-social-links` — social link form clobber / `entityId` race. Same family as the fixes in #2281.

The one failure that belonged to this PR was the newly un-skipped mobile event resources reorder. `verifyReorder` was treating a brief optimistic swap as success, then failing when a stale refetch briefly reverted the list. Pushed a fix that polls with `expect().toPass` until the order settles (same pattern the desktop resources spec already uses).

the remaining red here should clear after #2281 merges.

aasimsyed:
this also stabilizes the reorder drags themselves and picks up two related fixes:

- Mobile e2e drags now use CDP `Input.dispatchTouchEvent` instead of `page.mouse`. Sortable's touch fallback only reacts to real `TouchEvent`s, synthetic pointer events could reorder the DOM without firing `@end`. Needed `force-fallback` and `touch-action: none` on the draggable lists to make that path deterministic, this is a real mobile UX fix, not just test scaffolding.
- `scrollHandlesClearOfChrome` now measures the real header/submenu/bottom-nav bounding boxes instead of hardcoded pixel offsets.
- `useEventResourcesMutations`/`useEventFAQEntryMutations` write the new order into the Pinia Colada cache in `onMutate` and roll back in `onError`, so a stale refetch can't clobber an in-flight drag and a failed reorder doesn't stick. Also fixed the shared `@pinia/colada` test mock, which never called `onMutate`.
- Unrelated: fixed search-result card titles crowding the text below them (`-mb-2`).

Verified: mobile/desktop reorder specs green across repeats, full unit suite (1899 tests) passing, no regressions in create/update/delete flows.

andrewtavis:
@aasimsyed, do you think we're ready for a review here now?

aasimsyed:
@andrewtavis yes, I think it's ready

nicki182:
@aasimsyed can you look at the conflicts and resolve them?

nicki182:
@aasimsyed I noticed you ara manually changing the cache. As a design choice should be done everywhere or nowhere in my opionion since this would be a change of how we are doing are cache strategy. I prefer how we have it for now and maybe afterwards we can do the other one of we feel like it

aasimsyed:
@nicki182 I dropped the manual cache writes so reorder uses the same invalidate-on-success path as the other mutations.

The snap-back was the page watch overwriting the local drag list with a stale refetch. The event FAQ/resource pages now keep local order only when the incoming query has the same IDs in a different order. Create, update, and delete still sync.

nicki182:
@aasimsyed sorry for doing this but can you make all faqs and resources reorder mutations with optimistic update?. I didnt consider in the moment that for this case we have to use this cache strategy since we are ordering assuming that it goes well based on the UX. But only for this case and maybe later on we can create a generic useAppMutation where we have an option called cache or cacheStrategy and it can be default or optimistic or maybe even more

nicki182:
@aasimsyed I looked at the latest changes and looks perfect. I will do a check later and if everything works well as I imagine it does, I will merge this in

nicki182:
@aasimsyed just checked and it works perfectly. I will approve the changes after it passes the tests

nicki182:
@aasimsyed can you check on desktop flaky test regarding paginations for events?

nicki182:
Thank you for everything this is great :smile:

PR #2325, opened 2026-08-04
test(backend): standardize HTTP 401 and 403 tests

### LINKED ISSUE #2318: test(backend): standardize HTTP 401 and 403 tests for groups and organizations
### Terms

### Issue

Following the discussion in #2299, the backend HTTP 401 and 403 tests should be made consistent within an initial, limited scope:

- `backend/communities/groups`
- `backend/communities/organizations`

### Current inconsistencies

The tests currently use several different approaches:

- fresh unauthenticated `APIClient` instances;
- Django's `Client` fixture;
- the shared `authenticated_client` fixture;
- real sign-in requests followed by token handling;
- status-only assertions;
- status and exact `detail` assertions;
- denied writes with and without database state assertions;
- multiple HTTP outcomes inside a single test function.

Specific naming and structure issues include:

- `test_org_by_user_forbidden_403` asserts HTTP 401.
- `test_org_faq_create_unathorized` contains a typo and describes a 403 scenario as unauthorized.
- `test_org_faq_update_unauthorized_forbidden_403` mixes the 401 and 403 terminology.
- `test_org_faq_delete` covers 401, 403, 204 and 404 responses in one function.
- `test_org_by_user_unauthorized_401` performs a successful sign-in but does not authenticate the GET request.
- Some group tests perform real sign-in requests or contain unrelated successful GET requests before testing the denied operation.
- Some comments and docstrings use “unauthorized” for authenticated users who should receive 403.

### Proposed conventions

- Use a fresh, unauthenticated `APIClient` for HTTP 401 tests.
- Use `authenticated_client` for HTTP 403 tests.
- For ownership-based 403 scenarios, explicitly create a separate resource owner and ensure the authenticated actor is neither the owner nor staff.
- Avoid real sign-in requests when the test is concerned with endpoint permissions rather than the sign-in flow.
- Use one denied request and one authentication or authorization scenario per test function.
- Always assert the expected HTTP status.
- Consistently assert the response `detail` for 401 and 403 responses:
  - the standard authentication error for 401;
  - the endpoint-specific permission error for 403.
- Use consistent names:
  - `test_<resource>_<action>_unauthorized_401`
  - `test_<resource>_<action>_forbidden_403`
- For rejected writes, assert that no database mutation occurred:
  - POST: the record count remains unchanged;
  - DELETE: the target record still exists;
  - PUT/PATCH: refresh the record and assert that its original values remain unchanged.
- Keep the implementation test-only and avoid changing application behavior.

The tests introduced in #2299 can be used as one initial reference, particularly for verifying that rejected write requests do not modify the database.

### Indicative files

Relevant tests are mainly under:

- `backend/communities/groups/tests/faq`
- `backend/communities/groups/tests/flag`
- `backend/communities/groups/tests/resource`
- `backend/communities/groups/tests/social_link`
- `backend/communities/groups/tests/text`
- `backend/communities/groups/tests/test_group_delete.py`
- `backend/communities/groups/tests/test_group_update.py`
- `backend/communities/organizations/tests/faq`
- `backend/communities/organizations/tests/flag`
- `backend/communities/organizations/tests/resource`
- `backend/communities/organizations/tests/social_link`
- `backend/communities/organizations/tests/text`
- `backend/communities/organizations/tests/test_org_by_user.py`
- `backend/communities/organizations/tests/test_org_delete.py`
- `backend/communities/organizations/tests/test_org_update.py`

A later issue can extend the same conventions to the rest of the backend once this initial scope has been completed.

### Contribution

I would be happy to work on this task once the proposed scope and conventions have been confirmed.

Ulbrec87:
### Contributor checklist

- [x] This pull request is on a separate branch and not the main branch
- [x] I have run the relevant backend tests as described in the contributing guide

---

### Description

Standardizes the existing HTTP 401 and 403 tests under the groups and organizations areas.

- Uses unauthenticated clients for 401 tests and authenticated non-owner users for 403 tests.
- Adds consistent status, `detail` and database state assertions.
- Corrects misleading test names and removes unnecessary sign-in requests.
- Splits the organization FAQ delete scenarios into separate tests.
- Keeps the changes test-only and does not modify application behavior.

Testing:

- Groups and organizations test suites: `152 passed`

### Related issue

- Closes #2318


andrewtavis:
Thanks so much for the PR, @Ulbrec87! Could you check out the backend static analysis tests and send along a fix for them? :)

andrewtavis:
Note: It looks like these errors weren't introduced here, so no stress on it, @Ulbrec87 :) Looks to have come from https://github.com/activist-org/activist/commit/6ef264d8f047806668f36a3cb6d4019d0220bc85 where we updated `ruff` and `djangorestframework-stubs`.

andrewtavis:
CC @sh-ran for checking out and appreciating some very nice backend testing code coming in via this PR ✨😊

backend/communities/groups/tests/social_link/test_group_social_link_delete.py
```diff
@@ -36,16 +38,20 @@ def test_group_social_link_delete_not_found_404(authenticated_client):
 
 def test_group_social_link_delete_forbidden_403(authenticated_client):
     client, user = authenticated_client
+    user.is_staff = False
```
andrewtavis: The one thing I'm wondering about here so far, @Ulbrec87, is these sets of lines:

```py
user.is_staff = False
user.save(update_fields=["is_staff"])
```

We're not using `user` in later parts of the test. Maybe we want to do `client, _ = authenticated_client` to be clear that we're not using the user.

Thoughts on this, @sh-ran? :)

backend/communities/organizations/tests/faq/test_org_faq_delete.py
```diff
@@ -13,86 +13,71 @@
...
-    # Test 5: Deleting a non-existent FAQ returns 404.
+@pytest.mark.django_db
+def test_org_faq_delete_not_found_404(authenticated_client) -> None:
+    client, _ = authenticated_client
     fake_uuid = "00000000-0000-0000-0000-000000000000"
-    response = creator_client.delete(
-        f"{'/v1/communities/organization_faqs'}/{fake_uuid}"
-    )
+
+    response = client.delete(f"/v1/communities/organization_faqs/{fake_uuid}")
```
andrewtavis, line 77: Thanks for this code quality catch, @Ulbrec87!

andrewtavis: Seems like there were a few instances. Good to get these fixed up 😊

andrewtavis:
Thanks for the PR, @Ulbrec87! The comment on the `user` usages is the only question I have on this. Would be great if we could do similar checks for other parts of the backend testing :)

backend/communities/groups/tests/social_link/test_group_social_link_delete.py
Ulbrec87: I took another look at the code, and I think we should keep binding the user here because it is used to set up the permission scenario: the authenticated user must explicitly be non-staff before making the delete request.

Using `_` would make it look like the value is ignored, while it actually matters for the test setup. That said, I’d be happy to change it if you think that’s the right solution!

andrewtavis: I could be a bit confused, but then `test_group_social_link_delete_forbidden_403` doesn't use the `user` variable anywhere else in the code, and if we remove the two lines discussed the test still passes. Sometimes we do use the `user` from `authenticated_client` like in `test_group_social_link_delete_no_content_204`, but in some cases we don't.

Let me know if there's something specific that I'm missing here!

Ulbrec87: You're right, thanks for pointing this out! I checked the fixture again, and the authenticated user is already non-staff by default, so those two lines are redundant. I'll remove them and use client, _ = authenticated_client here.

andrewtavis (approved):
praise: Thanks so much for this, @Ulbrec87! Really a nice contribution to the backend tests 😊 I went ahead and did some variable renames so that they're a bit more clear and corrected some where we had a plural instead of a singular (i.e. `_social_links` instead of `_social_link`).

PR #2334, opened 2026-08-08
fix: surface DRF field-keyed array error messages to the user

### LINKED ISSUE #2332: We need to communicate the maximum filesize to the user on an image that's too large being uploaded
### Terms

- [x] I have searched [open and closed design issues](https://github.com/activist-org/activist/issues?q=is%3Aissue+label%3Adesign)

### Description

We have a default message to the user to let them know that their image is too large, but this isn't communicating what the maximum size is. It would be great if this could be updated in the backend (at time of writing we're assuming it's 5MB).

How to upload an image:

- Start your development environment
- Sign in at http://localhost:3000/auth/sign-in using `admin` `admin` for the username and password
- Go to any organization and click on the pencil icon to edit the images
- In this case upload a large image file

### Contribution

Happy to support on this issue and review a PR! 😊

hamodywe:
extractMessage's fallback now flattens array-valued fields (not just string-valued ones) from the error payload, so a DRF validate()-level ValidationError -- which serializes as {"non_field_errors": ["message"]} -- surfaces its actual message instead of falling through to a generic error string. No other call site changed.

### Related issue

Closes #2332


andrewtavis:
Thanks for the PR, @hamodywe :) We'll try to get to the review soon 😊

andrewtavis:
@hamodywe, I'm still getting the generic "Payload too large" toast from trying to upload something above 5MB. Can you check to see if it's working and send along a screenshot or screencast of what the user sees when trying to upload an image that's above the limit?

hamodywe:
> @hamodywe, I'm still getting the generic "Payload too large" toast from trying to upload something above 5MB. Can you check to see if it's working and send along a screenshot or screencast of what the user sees when trying to upload an image that's above the limit?

Tracked it down: the request wasn't reaching the serializer at all.

`nuxt-security`'s `requestSizeLimiter` runs on the Nuxt server before the proxy to the backend, and `nuxt.config.ts` had `maxUploadFileRequestInBytes: 5000000` while the backend's `IMAGE_UPLOAD_MAX_FILE_SIZE` is `5 * 1024 * 1024` (5242880). The edge limit was the lower of the two, so anything the backend was meant to reject got a 413 "Payload Too Large" from Nitro first — that's the toast, and it's why my earlier change never surfaced.

The same setting has a second effect: it measures the whole multipart request rather than each file. Two 3MB images are already past it, and the modals allow ten per upload.

The fix derives the ceiling from what the app itself permits (uploadLimit × the backend's per-file limit, plus the multipart envelope) and puts both constants in `shared/utils/uploadLimits.ts`, imported by `nuxt.config.ts`, so they can't drift apart again.

Worth flagging: this does raise the edge cap a fair amount, and that's your call rather than mine. If you'd prefer to stay conservative, a value just above 5 MiB fixes the single-file case and leaves batches as they are — one constant either way.

andrewtavis:
praise: Really appreciate the care you've put into this investigation, @hamodywe! I'll have another look as soon as I can 😊

andrewtavis (approved):
praise: This was a really nice review process and the result turned out great, @hamodywe 😊 Thanks so much for your contribution! Please let us know if there's another issue you'd have interest in :)

hamodywe:
> praise: This was a really nice review process and the result turned out great, @hamodywe 😊 Thanks so much for your contribution! Please let us know if there's another issue


 you'd have interest in :) you're very welcome, I really enjoyed

PR #2336, opened 2026-08-09
fix(backend): apply EXIF orientation before stripping metadata on upload

### LINKED ISSUE #2333: Images being flipped on upload
### Terms

### Behavior

An issue that we realized in a recent check of the application is that vertical images are flipped 90 degrees to be horizontal on upload to the platform. There are a few potential reasons for this:

- The most likely cause is the filescan service is returning the image flipped based on some logic defined in that part of the services
- There could also be something defined in the backend that's causing this, but this is less likely

How to upload an image:

- Start your development environment
- Sign in at http://localhost:3000/auth/sign-in using `admin` `admin` for the username and password
- Go to any organization and click on the pencil icon to edit the images
- In this case upload an image that has a greater height than width

sisodiajatin:
### Contributor checklist

---

### Description

Fixes #2333

Images uploaded with a non-default EXIF orientation tag (e.g. photos taken in portrait mode on a phone) were displaying rotated/flipped after upload.

**Root cause:** `scrub_exif()` in `backend/content/serializers.py` opens uploaded images with PIL and strips their EXIF metadata for privacy, but never read the `Orientation` tag before doing so. Since browsers/OSes rely on that tag (not the raw pixel data) to display images correctly, stripping it without applying it first left the raw sideways/upside-down pixel data as the final saved image, permanently baked in wrong.

**Fix:** call `PIL.ImageOps.exif_transpose()` immediately after opening the image, which physically rotates the pixel data to match the EXIF orientation *before* the EXIF data is stripped. A `or img` fallback handles the (rare) case where `exif_transpose()` returns `None`.

**Testing:** verified locally using the standard [EXIF orientation test images](https://github.com/recurser/exif-orientation-examples) (`Portrait_3.jpg`, `Portrait_6.jpg`) — both displayed correctly rotated after the fix, versus sideways/upside-down before.

### DIFF
--- a/backend/content/serializers.py
+++ b/backend/content/serializers.py
@@ -10,6 +10,7 @@ from typing import Any
 from django.conf import settings
 from django.core.files.uploadedfile import InMemoryUploadedFile, UploadedFile
 from PIL import Image as PILImage
+from PIL import ImageOps  # New Import
 from rest_framework import serializers
 
 from communities.groups.models import GroupImage
@@ -96,6 +97,9 @@ def scrub_exif(image_file: InMemoryUploadedFile) -> InMemoryUploadedFile:
     """
     try:
         img: PILImage.Image = PILImage.open(image_file)
+        img = (
+            ImageOps.exif_transpose(img) or img
+        )  # physically apply the orientation before stripping EXIF
         output_format = img.format
 
         if output_format == "JPEG":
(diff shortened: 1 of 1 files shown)


sisodiajatin:
Before/after using the standard EXIF orientation test images (Portrait_6.jpg):
****Before:** displayed sideways, ignoring the EXIF orientation tag.**
before_rotated

****After:** displays correctly upright once `exif_transpose()` is applied.**
after_fixed

andrewtavis (approved):
praise: Just checked and all's working here, @sisodiajatin :) Thanks for your contribution!

note: In the future, let's try to make sure that the comments that we're added are appropriate (we didn't need the new import comment).

PR #2340, opened 2026-08-09
feat(backend): communicate maximum filesize in MB for oversized image uploads (#2332)

UwaisBaig:
### Description

Updating the image upload validation error messages in the backend serializers to explicitly communicate the human-readable maximum file size in megabytes (`5MB`) alongside the byte count.

### Changes Made

- **`backend/content/serializers.py`**: Updated `ImageSerializer` and `ImageIconSerializer` error messages when `file_object.size > settings.IMAGE_UPLOAD_MAX_FILE_SIZE` to include `({max_mb}MB)`: `The file size (... bytes) is too large. The maximum file size is ... bytes (5MB).`
- **`backend/content/tests/image/test_content_image_upload.py`**: Updated upload test assertions to match the updated error message.

### Testing

- Verified image upload validation error message formatting.

Related #2332

### DIFF
--- a/backend/content/serializers.py
+++ b/backend/content/serializers.py
@@ -178,14 +178,13 @@ class ImageSerializer(serializers.ModelSerializer[Image]):
             )
 
         # DATA_UPLOAD_MAX_MEMORY_SIZE and IMAGE_UPLOAD_MAX_FILE_SIZE are set in core/settings.py.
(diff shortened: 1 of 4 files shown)


andrewtavis (approved):
I switched the PR message to not close the issue as there's more work to be done, but I agree that this is a nicer message for the user, @UwaisBaig :)

Thanks for sending this along!

Issue #2355, opened 2026-08-18
Password reset email page displays i18n key instead of invalid email message


Ayananth:
### Terms

- [x] I have searched all [open bug reports](https://github.com/activist-org/activist/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

## Summary

On `/auth/pwreset/email`, submitting an invalid email address displays
the raw translation key `i18n.pages.auth._global.invalid_email` instead
of the localized error message (for example, **"Invalid email
address"**).

## Steps to reproduce

1.  Open `/auth/pwreset/email` (or
    `http://localhost:3000/auth/pwreset/email` locally).
2.  Enter an invalid email address, for example `not-an-email`.
3.  Submit the form.
4.  Observe the validation error displayed below the input field.

## Actual behavior

The raw i18n translation key is displayed:

`i18n.pages.auth._global.invalid_email`

## Expected behavior

A localized, user-friendly validation message should be displayed, for
example:

**"Invalid email address"**

## Screenshot

<img width="1918" height="1005" alt="Image" src="https://github.com/user-attachments/assets/0bf5f070-e3b5-4d3a-a94d-a16ed6727ac5" />

The screenshot above shows the raw translation key being displayed in
the validation error.

### Notes

- The Zod schema in `frontend/app/pages/auth/pwreset/email.vue` passes the i18n key directly instead of using `t()`.
- Other auth pages use translated messages in their Zod schemas.
- `FormErrorMessage` renders the message as-is, and the current unit/E2E tests allow or expect the raw key.

### Possible fix

Call `t(...)` in the Zod schema, matching the other auth pages, and update the corresponding unit test.

## Contribution

If this issue is confirmed, it looks like a potentially good **first-time contributor / beginner-friendly issue**, so anyone interested is very welcome to pick it up.

If there’s no one working on it after 2 days, I’d be happy to take it up. 😊

andrewtavis:
Thank you so much for the issue, @Ayananth! Yes, we definitely need the string localized. Everything else will work from there as i18n-check is showing that that's a valid key.

minutechreview:
@Ayananth I'd be happy to take this, but you mentioned you'd pick it up yourself if nobody had after a couple of days — since you found it, say the word and I'll happily leave it to you. No hard feelings either way.

If it's free, here's what I've confirmed:

[...]

PR #2356, opened 2026-08-18
fix: localize password reset email validation messages

minutechreview:
### Contributor checklist

---

### Description

`frontend/app/pages/auth/pwreset/email.vue` passed its i18n keys to the Zod schema as bare strings, so the raw key was rendered as the validation message:

```ts
.email("i18n.pages.auth._global.invalid_email")
.min(1, "i18n._global.required"),
```

`t` was already in scope in that file (line 36), and the other auth pages already wrap their schema messages — `sign-in.vue:97` and `sign-up.vue:202` both use `t("i18n._global.required")` — so this page was the outlier rather than the pattern needing to change. That matches @andrewtavis's note that the key itself is valid.

**Files changed**

- `app/pages/auth/pwreset/email.vue` — wrap both schema messages in `t()`
- `test/pages/auth/pwreset/email.spec.ts` — the existing test asserted the raw key as the expected output (with a comment stating the raw key is rendered), so the bug was codified in the suite. It now asserts the localized string via `getEnglishText()`, matching how the e2e specs assert translated text.

**Testing**

- `yarn test test/pages/auth/pwreset/email.spec.ts` — 5 passed, 2 todo
- `yarn test` (full unit suite) — 1888 passed, 177 files
- `yarn prettier --check` and `yarn eslint` on both changed files — clean

I also confirmed the updated test actually discriminates: reverting the `.email()` fix makes it fail, so it isn't passing vacuously.

**One thing worth flagging**

I fixed `.min(1, ...)` alongside `.email(...)` for consistency, but I could not write a meaningful test for it. `z.string().email()` rejects an empty string before `.min(1)` is reached, so that message appears unreachable through the form — reverting only the `.min()` fix leaves the whole suite green. I removed a test I had written for it rather than ship something that looks like coverage but proves nothing. Happy to drop the `.min()` change if you'd rather keep this strictly to the reported line.

**Risks**

Low. No behaviour changes beyond the rendered message text; both keys already exist in `en-US.json` (`"Invalid email address"` and `"Required"`).

**AI disclosure**

Per the template: this contribution was written with AI assistance. The diagnosis, the code change, the test update, and this description were produced with an AI coding assistant, and every claim above was verified by running the commands locally rather than asserted.

### Related issue

- #2355

### DIFF
--- a/frontend/app/pages/auth/pwreset/email.vue
+++ b/frontend/app/pages/auth/pwreset/email.vue
@@ -38,8 +38,8 @@ const t = useI18n().t;
 const emailResetPasswordSchema = z.object({
   email: z
     .string()
-    .email("i18n.pages.auth._global.invalid_email")
-    .min(1, "i18n._global.required"),
+    .email(t("i18n.pages.auth._global.invalid_email"))
+    .min(1, t("i18n._global.required")),
 });
 
 const submit = async (values: unknown) => {
--- a/frontend/test/pages/auth/pwreset/email.spec.ts
+++ b/frontend/test/pages/auth/pwreset/email.spec.ts
@@ -74,11 +74,10 @@ describe("pwreset/email", () => {
     });
     await fireEvent.click(submitButton);
 
-    // The FormErrorMessage renders the raw i18n key as its text content.
     await waitFor(() => {
       expect(screen.getByRole("alert")).toBeTruthy();
       expect(screen.getByTestId("form-item-email-error").textContent).toContain(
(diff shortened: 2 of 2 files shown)


andrewtavis:
Thanks for the PR, @minutechreview! We'll try to bring this in soon :)

andrewtavis (approved):
praise: Thanks so much for the contribution here, @minutechreview! Please let us know if there's another issue that's of interest 😊

PR #2396, opened 2026-09-10
fix(frontend): render theme icon using picture element in spa loading template (#2391)

### LINKED ISSUE #2391: activist a is shown in light and dark mode on page load
### Terms

### Behavior

We have a weird introduced bug that came in where the activist a is shown in both light and dark mode on page load. This is not only in development, but also in production via Netlify. Would be great if we could fix this asap. Note that what's happening is `spa-loading-template.html` is being triggered.

UwaisBaig:
### Summary of Changes

Fixes #2391 where the activist "a" icon was displayed in both light and dark mode simultaneously on page load when `spa-loading-template.html` was rendered.

#### Problem:
`frontend/app/spa-loading-template.html` rendered two distinct `` tags (`.img-light-mode` and `.img-dark-mode`) and relied on CSS classes in a trailing `<style>` block to hide `.img-dark-mode` or `.img-light-mode`. Prior to stylesheet parsing or when styles were delayed/blocked, both images were rendered side-by-side or stacked on top of each other.

#### Solution:
- Replaced the two separate `` elements with a single semantic `<picture>` element using a `<source>` with `media="(prefers-color-scheme: dark)"` and fallback ``.
- The browser natively selects and renders exactly one image based on user color scheme preference without relying on class toggles or risking multiple image elements displaying at once.
- Moved the `<style>` block above the DOM elements to prevent flash of unstyled content.

### Related Issue

- Fixes #2391

### DIFF
--- a/frontend/app/spa-loading-template.html
+++ b/frontend/app/spa-loading-template.html
@@ -1,21 +1,4 @@
 <!-- SPDX-License-Identifier: AGPL-3.0-or-later -->
-<div>
-  <img
-    class="img-light-mode"
-    src="/icons/activist/activist_icon_light.png"
-    height="160"
-    width="160"
-    alt="activist - loading the platform"
-  />
-  <img
-    class="img-dark-mode"
-    src="/icons/activist/activist_icon_dark.png"
-    height="160"
-    width="160"
-    alt="activist - loading the platform"
-  />
-</div>
-
 <style>
   html,
   body {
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
A couple of things were needed to finalize this and clean it up a bit - specifically `spaLoadingTemplate: true` in `nuxt.config.ts`. Thanks for sending along the fix, @UwaisBaig :)

andrewtavis:
Bringing this in as e2e tests are needed here :)

Issue #1376, opened 2025-07-23
Correct code smells found in file activist > backend > events > serializers.py


saranettle:
### Terms

- [x] I have searched [open and closed issues](https://github.com/activist-org/activist/issues?q=is%3Aissue)
- [x] I agree to follow activist's [Code of Conduct](.github/CODE_OF_CONDUCT.md)

### Description:
I was reviewing some backend files, and I located a few code smells that could be corrected to improve future readability.

- commented-out code in lines 67-68: this can be removed
- super long doc strings lines 127-142, 196-206, and 243-258: The doc strings are just repeating the same information that appears in the function declaration, and they could be shortened to just include the description.
- the validate(self, data) function starting at line 125 is very long: could be broken down into a few helper functions.

Regarding that last item: refactoring a function is a little more complex than correcting doc strings and removing a comment. Please let me know if it makes sense to put this into a separate issue.

### Contribution:
I can work on this issue ASAP.

andrewtavis:
Assigned, @saranettle! Thanks for finding all of this and opening the issue! Please also let us know if the PR needs to be merged by your deadline. We can very very likely accommodate. Mostly just good to know :)

sailanka02:
@saranettle and @andrewtavis, if this still open is there any chance I can help contribute to this?

andrewtavis:
@saranettle will be working on this one, @sailanka02, but you're assigned to #1175 😊

saranettle:
Update on this: I will keep the docstrings as they are. I am realizing that there is a required format for docstrings.

I have created a helper function for validating start dates are before end dates - I will create the PR soon.

andrewtavis:
Ah yes the docstrings are set by numpydocs, which is what we agreed on using. Looking forward to the PR! 😊

andrewtavis:
Closed by #1415 🚀 Please let us know if you find anything else like this in the codebase, @saranettle! Appreciate your support :)

PR #2399, opened 2026-09-11
Fix duplicate code in event serializers

HarxhitSingh:
This fixes issue #1376. I removed repeated code by creating one shared function instead of copy-pasting it three times.

### DIFF
--- a/backend/events/serializers.py
+++ b/backend/events/serializers.py
@@ -35,6 +35,7 @@ from events.models import (
 )
 from utils.utils import (
     validate_creation_and_deprecation_dates,
+    validate_object_exists,
 )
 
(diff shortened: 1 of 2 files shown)


andrewtavis:
This seems like something we could bring in, @HarxhitSingh, but could you fix the failing [static analysis tests for the backend](https://github.com/activist-org/activist/actions/runs/34626105454/job/103365889574?pr=2399)?

andrewtavis (approved):
Failing e2e test is flaky, so we can bring this in :)

praise: Thanks for the code improvements and the reaction to the initial review, @HarxhitSingh! I modified this so we're using the same code for all entities (events, orgs and groups).

